PR #61, opened 2016-02-05
Create 2018_File_Spec.csv

lesgou:
Updates the document file with the 2018 File Spec.


jmarin:
@lesgou Can we have two documents, one for each table? I think it would read better

Issue #62, opened 2016-02-09
Parser error handling


jmarin:
Provide a sensible mechanism to report parsing errors

schbetsy:
Combining with #64, since they're basically duplicates. Text: 

"As mentioned on [#63 (comment)](https://github.com/cfpb/hmda-platform/pull/63#discussion_r53020374), we need to make sure individual invalid data fields cannot blow up the whole parser. The exact design of this is TBD."

Text of linked comment from Hans:

(on `LarCsvParser`) "The rest of this section is duplicate from the first LarCsvParser. I think this should be split up, with the first step simply being to translate a row into a common data structure (perhaps just an array of strings?), and then parse each element into the domain objects."

PR #63, opened 2016-02-11
Parser

jmarin:
Some additional information to assist in reviewing this PR

There is a big refactor associated with this, in order to provide support for Scala.js [1] in the following projects: 

- `model`: this project contains the domain model for `Transmittal Sheet` and `Loan Application Register`, both of which are grouped into `FIData`. All code lives in the `shared` subproject, there are no platform specific implementations (they are the same for both the JVM and JS). **Classes here are only for 2017 data model**
- `parser`: contains manual parser that splits `HMDA` files (both .DAT and .CSV formats) into the model above. The actual parser is cross platform, there are a couple of tests that take a file as input and those live in the respective platform implementation. The `js` project includes a small wrapper around `Node.js` File IO operations (minimal for now, just the ones used in the test). **Parser only supports 2017 data model for now**
- `platform-test`: this project is meant to serve as a resource for testing implemented modules. In this PR it just includes HTML and JavaScript that exercises the parser client side, with two implementations in the `js` subproject, one done in JavaScript and the other one entirely in Scala.js as a proof of concept of interoperability. 

As of this commit, the only project that remains Scala-only is `api`, which is where the Akka Http endpoints live.

[1] [This](http://www.scala-js.org/doc/project/cross-build.html) might be useful when reviewing some of the changes in this PR


model/shared/src/main/scala/hmda/model/fi/FIData.scala
```diff
@@ -0,0 +1,20 @@
+package hmda.model.fi
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.model.fi.ts.TransmittalSheet
+import scala.scalajs.js.annotation.JSExport
+import scala.scalajs.js.JSConverters._
+
+case class FIData(ts: TransmittalSheet, lars: Iterator[LoanApplicationRegister]) {
+  def toCSV: String = {
```
hkeeler: This is going to be a _very_ large `String` for a big dataset.  Seems like this should be some sort of output stream.

model/shared/src/main/scala/hmda/model/fi/FIData.scala
```diff
@@ -0,0 +1,20 @@
...
+    s"${ts.toCSV}\n" +
+      s"${lars.map(lar => lar.toCSV + "\n")}"
```
hkeeler: extra indent?

parser/shared/src/main/scala/hmda/parser/fi/FIDataParser.scala
```diff
@@ -0,0 +1,10 @@
+package hmda.parser.fi
+
+import hmda.model.fi.FIData
+
+trait FIDataParser[A] {
```
hkeeler: What other types than `String` are you imagining for `A`?

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
```diff
@@ -0,0 +1,98 @@
+package hmda.parser.fi.lar
+
+import hmda.model.fi.lar._
+
+object LarCsvParser {
+  def apply(s: String): LoanApplicationRegister = {
+    val values = s.split('|').map(_.trim)
```
hkeeler: In the long run, I think we'll want a real CSV (or PSV in our case) parsing library.  I'm guessing there _will_ be legit reasons for `|` characters, which would cause the string to be split incorrectly.

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
```diff
@@ -0,0 +1,98 @@
...
+    val id = values(0).toInt
+    val respId = values(1)
+    val code = values(2).toInt
```
hkeeler: Is this `agencyCode`?

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
```diff
@@ -0,0 +1,98 @@
...
+    val loanId = values(3)
+    val loanDate = values(4)
+    val loanType = values(5).toInt
```
hkeeler: Isn't this going to blow up the whole parser with the first non-Int string it comes across?

parser/shared/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
```diff
@@ -0,0 +1,98 @@
...
+    val appIncome = s.substring(95, 99)
+    val purchaserType = s.substring(99, 100).toInt
+    val denial1 = s.substring(100, 101)
+    val denial2 = s.substring(101, 102)
+    val denial3 = s.substring(102, 103)
+    val rateSpread = s.substring(103, 108)
+    val hoepaStatus = s.substring(108, 109).toInt
+    val lienStatus = s.substring(109, 110).toInt
+
+    val loan =
```
hkeeler: The rest of this section is duplicate from the first `LarCsvParser`.  I think this should be split up, with the first step simply being to translate a row into a common data structure (perhaps just an array of strings?), and then parse each element into the domain objects.

platform-test/js/src/main/resources/index.html
```diff
@@ -0,0 +1,49 @@
...
+    <script type="text/javascript" src="../platformtestjs-fastopt.js"></script>
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css">
+</head>
+<body style="margin: 0px">
+
+  <div class="container">
+     <h1>HMDA Parser Example (JavaScript)</h1>
+     <input type="file" id="files" name="files[]">
+     <h2>Transmittal Sheet</h2>
+     <table if="tsTable" class="table">
```
hkeeler: I think `if` is supposed to be an `id`?

README.md
```diff
@@ -70,16 +71,55 @@ $ sbt
...
 
+To build the project artifacts, there are two options depending on which runtime is being targeted. 
+
+* To build JVM artifacts (the default, includes all projects), from the sbt prompt:
+
+```shell
+> assembly
+```
+
+This task will create a `fat jar`, which can be executed directly on any JDK8 compliant JVM:
```
hkeeler: Is this supposed to be "fat jar"?  It seems like there's a `fat jar` command as it stands.

hkeeler:
All reviewed up.  Most of the comments are more for discussion as we evolve the parser over the coming sprints.  The only bug I see at this point is https://github.com/cfpb/hmda-platform/pull/63#discussion_r53016601.

model/shared/src/main/scala/hmda/model/fi/FIData.scala
jmarin: Correct. This may not be needed. I think it's more important to have the individual CSV representation of TransmittalSheet and LAR

model/shared/src/main/scala/hmda/model/fi/FIData.scala
jmarin: For some reason the IDE _forces_ me to do this

parser/shared/src/main/scala/hmda/parser/fi/FIDataParser.scala
jmarin: Maybe binary. Maybe nothing

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
jmarin: Agreed on the parsing library. There are no legit reasons for `|` characters that I know of, certainly not for the LAR part. If there are, we need another separator.

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
jmarin: It will. Intend to fix when addressing this issue https://github.com/cfpb/hmda-platform/issues/62

parser/shared/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
jmarin: Can you create a ticket with a description? Might be a good refactoring to tackle next sprint

README.md
jmarin: Most documentation on `sbt assembly` refers to _fat jars_, basically the whole project gets packaged into one big jar with all necessary dependencies. See https://github.com/sbt/sbt-assembly

hkeeler:
Looks good.  Created https://github.com/cfpb/hmda-platform/issues/64 to remedy https://github.com/cfpb/hmda-platform/pull/63#discussion_r53020617.  Merging...

PR #78, opened 2016-02-22
Refactor LAR collection from Iterator to Iterable

jmarin:
`Iterator` inherits from `TraversableOnce`, which means that once consumed it cannot be reused. Changing to `Iterable`, which extends `Traversable` and allows consuming the collection several times.


schbetsy:
It makes sense to me that `Iterable` would be sufficient, so we wouldn't need `Iterator` here. Looks good to me.

PR #79, opened 2016-02-23
Travis setup

jmarin:


marcesher:
you can use `oraclejdk8` for testing jdk8

.travis.yml
```diff
@@ -12,10 +12,12 @@ scala: 2.11.7
...
 
 script: 
-  - sbt ++$TRAVIS_SCALA_VERSION clean assembly 
+  - sbt ++$TRAVIS_SCALA_VERSION clean coverage test assembly
   
   # Trick to avoid unnecessary cache updates
   - find $HOME/.sbt -name "*.lock" | xargs rm
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
hkeeler: How does [codecov-bash](https://github.com/codecov/codecov-bash) differ `codecov` task described in their [Integration Guide](https://codecov.io/guide)?  Just to get around the `pip install codecov`, or different functionality?

hkeeler:
This is currently failing in Travis with a:

```
[error] Not a valid command: coverage
[error] Not a valid project ID: coverage
[error] Expected ':' (if selecting a configuration)
[error] Not a valid key: coverage (similar: homepage, package, compilerCache)
[error] coverage
[error]         ^
```

I think we need to add something similar to [grasshopper's `plugins.sbt`](https://github.com/cfpb/grasshopper/blob/master/project/plugins.sbt#L16-L19) to get the `coverage` command.

vickumar1981:
I believe the latest version of that plugin is 1.3.5.

https://github.com/scoverage/sbt-scoverage

Also, if you ever run into dependency problems, I've found this plugin to be useful as well, sbt-dependency-graph.

https://github.com/jrudolph/sbt-dependency-graph

jmarin:
This is ready for review. I am not sure what to make of the Codecov results (0.00%), the coverage reports are being written and right now the project is at ~66%. I believe it's trying to do a diff with reports on master and that is why it's 0, but not sure.

ericspry:
Makes sense @jmarin  - the codecov.io site just says, no reports for master branch. So once this is merged, should be good. You can view the detailed report on the PR here: https://codecov.io/github/cfpb/hmda-platform?pr=79

jmarin:
Interesting. That 0 jumped up to 62%, which seems about correct. Maybe it was just waiting to get updated

hkeeler:
All reviewed up.  Looks good.  Can't really confirm the badges will work until it's merged, but seems good.

hkeeler:
Confirmed both TravisCI and CodeCov badges are working.

Issue #69, opened 2016-02-17
Implement toCSV method (Loan Application Register)


jmarin:
https://github.com/cfpb/hmda-platform/blob/master/model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala#L22

See example for Transmittal Sheet [here](https://github.com/cfpb/hmda-platform/blob/master/model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala#L22)

[File Spec](http://www.ffiec.gov/hmda/pdf/spec2016.pdf) (pages 10 and 11)

schbetsy:
OK to close this issue?

PR #81, opened 2016-02-29
Implement toCSV method on LoanApplicationRegister

schbetsy:
#69 

I wanted to make a PR of my work from last week, in case it helps with sprint planning.

### DIFF
--- a/model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
+++ b/model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
@@ -20,8 +20,16 @@ case class LoanApplicationRegister(
     lienStatus: Int
 ) {
   def toCSV: String = {
-    //TODO: implement CSV output
-    ""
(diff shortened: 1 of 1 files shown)


schbetsy:
I didn't write specs for this method individually, since it seems like it will be tested by [this spec](https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/lar/LarCsvParserSpec.scala).

model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
```diff
@@ -20,8 +20,16 @@ case class LoanApplicationRegister(
     lienStatus: Int
 ) {
   def toCSV: String = {
-    //TODO: implement CSV output
-    ""
+    s"$id|$respondentId|$agencyCode|${loan.id}|${loan.applicationDate}" +
```
vickumar1981: is there a reason not to use multiline strings here?

Issue #71, opened 2016-02-17
Implement LAR test data generators


jmarin:
https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala

For an example, look at the Transmittal Sheet [generators](https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/ts/TsGenerators.scala)

hkeeler:
This _seems_ closable now with #82 and #84 merged.  We may get more strict with dates and geo-related fields at some point (see https://github.com/cfpb/hmda-platform/pull/84#issuecomment-191516995), but this seems good for this sprint.  Closing.

PR #82, opened 2016-03-01
initial shot at LAR generator. still missing several cases.

lgdean:
#71 
here's a first version. it is both too restrictive (missing some cases, particularly "NA" ones) and too permissive, as the many TODOs suggest. and the code could be cleaned up in places (once the functionality is clear/complete/correct).
it looks like other work is about to be blocked by not having this code, so I'm guessing an incomplete version is still helpful.

### DIFF
--- a/parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
+++ b/parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
 
 trait LarGenerators extends FIGenerators {
 
-  //TODO: implement LarGenerators
+  implicit def larGen: Gen[LoanApplicationRegister] = {
+    for {
+      respondent <- respIdGen
+      agencyCode <- agencyCodeGen
+      loan <- loanGen
+      preapprovals <- preapprovalGen
+      actionTaken <- actionTypeGen
+      actionTakenDate <- actionDateGen
+      geography <- geographyGen
+      applicant <- applicantGen
+      purchaserType <- purchaserTypeGen
+      denial <- denialGen
+      rateSpread <- rateSpreadGen
+      hoepaStatus <- hoepaStatusGen
+      lienStatus <- lienStatusGen
+    } yield LoanApplicationRegister(
+      2, // TODO we may not want to hard-code this everywhere (though it may be appropriate here)
+      respondent,
+      agencyCode,
+      loan,
+      preapprovals,
+      actionTaken,
+      actionTakenDate,
+      geography,
+      applicant,
+      purchaserType,
+      denial,
+      rateSpread,
+      hoepaStatus,
+      lienStatus
+    )
+  }
 
-  implicit def larGen: Gen[LoanApplicationRegister] = ???
+  implicit def loanGen: Gen[Loan] = {
+    for {
+      id <- Gen.listOf(Gen.alphaNumChar).map(_.mkString)
+      //applicationDate <- ??? // TODO Format is ccyymmdd or NA
+      loanType <- Gen.oneOf(1, 2, 3, 4)
+      propertyType <- Gen.oneOf(1, 2, 3)
+      purpose <- Gen.oneOf(1, 2, 3)
+      occupancy <- Gen.oneOf(1, 2, 3)
+      amount <- Gen.posNum[Int] // TODO is that right? can it be zero? (what would a loan for zero thousand dollars mean?)
+    } yield Loan(
+      id,
+      "NA",
+      loanType,
+      propertyType,
+      purpose,
+      occupancy,
+      amount
+    )
+  }
 
-  implicit def loanGen: Gen[Loan] = ???
+  implicit def preapprovalGen: Gen[Int] = Gen.oneOf(1, 2, 3)
 
-  implicit def preapprovalGen: Gen[Int] = ???
+  implicit def actionTypeGen: Gen[Int] = Gen.choose(1, 8)
 
-  implicit def actionTypeGen: Gen[Int] = ???
+  implicit def actionDateGen: Gen[Int] = Gen.choose(20170101, 20201231) // TODO this allows non-date numbers!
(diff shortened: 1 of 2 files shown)


parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
...
+
+  implicit def loanGen: Gen[Loan] = {
+    for {
+      id <- Gen.listOf(Gen.alphaNumChar).map(_.mkString)
+      //applicationDate <- ??? // TODO Format is ccyymmdd or NA
+      loanType <- Gen.oneOf(1, 2, 3, 4)
+      propertyType <- Gen.oneOf(1, 2, 3)
+      purpose <- Gen.oneOf(1, 2, 3)
+      occupancy <- Gen.oneOf(1, 2, 3)
+      amount <- Gen.posNum[Int] // TODO is that right? can it be zero? (what would a loan for zero thousand dollars mean?)
```
jmarin: I think the edits say it has to be > 0. At this stage we're not looking at particular edits, this generators are for the parser. So I think we're good

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
...
+
+  implicit def actionTypeGen: Gen[Int] = Gen.choose(1, 8)
+
+  implicit def actionDateGen: Gen[Int] = Gen.choose(20170101, 20201231) // TODO this allows non-date numbers!
+  // so play with things like LocalDate.of(2017, 01, 01) and figure out how to convert
+
+  implicit def geographyGen: Gen[Geography] = { // TODO any of these can be NA
+    for {
+      msa <- stringOfN(5, Gen.numChar) // actually it's more specific; do we care, for this purpose?
+      state <- stringOfN(2, Gen.numChar) // TODO limit to real ones? http://www2.census.gov/geo/docs/reference/state.txt
```
jmarin: We should probably generate valid state abbreviations as per that list

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
...
+      tract
+    )
+  }
+
+  implicit def censusTractGen = {
+    for {
+      tract <- stringOfN(4, Gen.numChar).map(Some(_)) // TODO are leading zeroes required? (they are allowed.)
+      suffix <- Gen.option(stringOfN(2, Gen.numChar))
+    } yield List(tract, Some("."), suffix).flatten.mkString
+  }
```
jmarin: I believe leading zeroes are required. Also, If I'm reading this one correctly it can generate empty suffix portion, I think that's not valid

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
...
+
+  implicit def applicantGen: Gen[Applicant] = {
+    for {
+      ethnicity <- Gen.choose(1, 4)
+      coEthnicity <- Gen.choose(1, 5)
+      race1 <- Gen.choose(1, 7)
+      //coRace <-  TODO this should be 8 if and only if coEthnicity is 5 (i.e. no co-applicant). do we care?
```
jmarin: Not at this point. Just that this field contains potential valid values. We will override some of these generators when we take a look at edits, both to produce valid and invalid combinations and test that the rules are correct.

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -6,30 +6,125 @@ import org.scalacheck.Gen
...
+  }
+
+  implicit def hoepaStatusGen: Gen[Int] = Gen.oneOf(1, 2)
+
+  implicit def lienStatusGen: Gen[Int] = Gen.oneOf(1, 2, 3, 4)
+
+  // TODO this is quite general and probably belongs elsewhere (if it turns out to be useful in clean/idiomatic code)
+  def stringOfN(n: Int, genOne: Gen[Char]): Gen[String] = {
+    Gen.listOfN(n, genOne).map(_.mkString)
+  }
```
jmarin: I suggest you move this one [here](https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/FIGenerators.scala) for now

jmarin:
@lgdean Good work on this, I think this is a good start. I put a couple of small comments in there that you might want to take a look before we merge this PR.

hkeeler:
:+1: Agreed.  Looks good.  As @jmarin said, we can merge now, or wait if you want to get a few of the mentioned tweaks in.

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
lgdean: makes sense. I hesitate to ask, but... do we need to do something similar for counties?

parser/shared/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
lgdean: You may be right in the context of HMDA data. I was going by what I found from the census bureau online. When we find things like this (where the document doesn't specify leading and trailing zeroes), should we be reporting them so the document can be updated?

lgdean: Actually, there's a broader question: should this generator cover 2016 data formats, or 2017, or both? (And are those supposed to be identical, in terms of allowed values?) I was looking at the 2017 document, which says nothing about leading zeroes for this field (though the example does have one, which is how I knew they were allowed).

hkeeler:
Looks good.  Merging...

PR #83, opened 2016-03-01
Upgrade to Akka 2.4.2

jmarin:


schbetsy:
So, all of the akka-related dependencies need to be included separately, but they're all on the same versioning scheme. Is this common in the Scala world?

schbetsy:
Also, LGTM.

PR #84, opened 2016-03-02
handled most of the cases

lgdean:
#71
All that remains is to use the correct list of FIPS state codes, and (if needed) more-specific values for MSA and/or county codes.
Seems worth going ahead and getting any comments on this part.

### DIFF
--- a/parser/shared/src/test/scala/hmda/parser/fi/FIGenerators.scala
+++ b/parser/shared/src/test/scala/hmda/parser/fi/FIGenerators.scala
@@ -1,8 +1,5 @@
 package hmda.parser.fi
 
-import java.time.LocalDate
-import java.time.format.DateTimeFormatter
-
 import org.scalacheck.Gen
 
 trait FIGenerators {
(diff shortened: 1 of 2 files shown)


schbetsy:
:+1:

parser/shared/src/test/scala/hmda/parser/fi/FIGenerators.scala
```diff
@@ -12,9 +15,22 @@ trait FIGenerators {
...
+  def optional[T](g: Gen[T], emptyVal: String = ""): Gen[String] = {
+    Gen.oneOf(g.map(_.toString), Gen.const(emptyVal))
+  }
+
+  def dateGen: Gen[String] = {
+    for {
+      year <- Gen.choose(2017, 2020) // or could parameterize if needed
+      month <- Gen.choose(1, 12)
+      day <- Gen.choose(1, 28) // NOTE: this misses some dates! if that's a problem, we can do something more complex.
+    } yield LocalDate.of(year, month, day).format(DateTimeFormatter.ofPattern("yyyyMMdd"))
```
jmarin: I suspect this is going to fail when trying to run it in Scala.js since it uses Java APIs that are not supported [1]. Have you tried this in a test? If not, could you write a small test that exercises this generator with dates to prove that it works in both environments? (something in the shared parser project should tell us what we need)

[1] http://www.lihaoyi.com/hands-on-scala-js/#JavaAPIs

jmarin: Just confirmed this fails on Scala.js due to using `java.time.LocalDate`, which is not implemented

jmarin: This library could be useful for this use case https://github.com/scala-js/scala-js-java-time

lgdean:
It is now the case that:
- state codes come from the list,
- county codes are (still) any combination of 3 digits,
- MSAs are (still) any combination of digits,
- dates will include a few impossible values such as 20180231, and
- the code doesn't use any date library (scala-js or otherwise), because it doesn't need to check date validity.

jmarin:
:+1:

PR #85, opened 2016-03-02
Upload api

jmarin:
Preliminary API upload endpoint
Asynchronously processes incoming CSV file (prints to screen for now)


api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -2,18 +2,34 @@ package hmda.api.http
...
+import scala.concurrent.Future
+import scala.util.{ Failure, Success }
+
 trait HttpApi extends HmdaApiProtocol {
 
+  implicit val system: ActorSystem
+  implicit val materializer: ActorMaterializer
   val log: LoggingAdapter
 
+  val splitLines = Framing.delimiter(ByteString("\n"), 2048, allowTruncation = true)
```
hkeeler: What's `allowTruncation = true` all about?  I'm having a hard time finding a good definition of that argument in their docs.  I think this snippet from [Streaming TCP](http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.3/scala/stream-io.html#Streaming_TCP) doc describes it, but it's a little unclear:

>The last boolean argument indicates that we require an explicit line ending even for the last message before the connection is closed.

Is that what's going on here?

api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -25,6 +41,37 @@ trait HttpApi extends HmdaApiProtocol {
           ToResponseMarshallable(status)
         }
       }
-    }
+    } ~
+      path("upload") {
+        post {
```
hkeeler: Why not the [`fileUpload`](http://doc.akka.io/docs/akka/current/scala/http/routing-dsl/directives/file-upload-directives/fileUpload.html) directive?

api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -25,6 +41,37 @@ trait HttpApi extends HmdaApiProtocol {
...
+          val processingActor = system.actorOf(ProcessingActor.props)
+          entity(as[Multipart.FormData]) { formData =>
+            val uploaded: Future[Done] = formData.parts.mapAsync(1) {
```
hkeeler: Why `mapAsync(1)`?  Is that because there should only be 1 form part?

api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -25,6 +41,37 @@ trait HttpApi extends HmdaApiProtocol {
...
+              //TODO: check Content-Type type as well?
+              case b: BodyPart if b.filename.exists(_.endsWith(".csv")) =>
```
hkeeler: In the [2017 HMDA File Specification](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf), we say:

>The file will be a text file with a `.txt` file format extension.

hkeeler:
All reviewed up.  Overall, looks good.  Just a few questions, and one nitpick about the file extension.

api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -25,6 +41,37 @@ trait HttpApi extends HmdaApiProtocol {
...
+                b.entity.dataBytes
+                  .via(splitLines)
+                  .map(_.utf8String)
+                  .runForeach(line => processingActor ! line)
+
+              case _ => Future.failed(throw new Exception("File is not CSV"))
```
hkeeler: Are there other failure scenarios we need to consider besides just the wrong file extension?

api/src/main/scala/hmda/api/http/HttpApi.scala
jmarin: That is correct

api/src/main/scala/hmda/api/http/HttpApi.scala
jmarin: No particular reason, it's just a higher level API. As it's implemented it would be equivalent, this way it's more explicit and gives more low level control

api/src/main/scala/hmda/api/http/HttpApi.scala
jmarin: Until we measure things under load we don't know what the optimum number is, most examples start with a value of 1 and take it from there. Not sure if this will influence things that much since the process is I/O bound

hkeeler:
Looks good.  Merging...

Issue #75, opened 2016-02-17
Implement FI Data DAT parser test


jmarin:
https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/FIDataDatParserSpec.scala

schbetsy:
Since `FIData` class doesn't have a `toDAT` method, I'm not sure how to test this. Usually, I would use the same strategy as the other parser tests (convert to DAT, then parse that DAT and make sure the result matches the starting object), but that won't work here. 

Should we add a `toDAT` method to `FIData`, or use a different testing strategy?

cc @jmarin

jmarin:
I think we don't have a `toDAT` method in `FIData` because I totally forgot to add it :)
So yes, that testing strategy should work. Keep in mind (or rather, ignore for now) that this method might generate a huge string for large datasets

PR #88, opened 2016-03-04
Unit test for FIDataDatParser

schbetsy:
#75 

I'm looking for feedback on this one.
- Is this the kind of tests we're looking for? Should I include more specific tests that the data were imported correctly?
- Should I wait until we have more realistic sample DATs, and use those instead?


jmarin:
@schbetsy This is fine for now. I am going to prepare some sample files that we can use for additional tests

PR #89, opened 2016-03-07
WIP: Validation

jmarin:


validation/shared/src/test/scala/hmda/validation/rules/lar/syntactical/S020Spec.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import hmda.parser.fi.lar.LarGenerators
+import hmda.validation.dsl.Success
+import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.prop.PropertyChecks
+
+class S020Spec extends PropSpec with PropertyChecks with MustMatchers with LarGenerators {
+  property("Loan Application Register Agency Code must = 1,2,3,5,7,9") {
+    forAll(larGen) { lar =>
+      whenever(lar.id == 2) {
```
schbetsy: I don't understand why we need this line (`whenever(lar.id == 2)`). The LAR generator always assigns `id` of `2`, right?

jmarin: True. It's usually necessary to use `whenever` to tell ScalaCheck what type of data to create, in this case it's a bit redundant. There might be some cases down the line where we only generate LARs that have a certain type of property

validation/js/src/main/scala/hmda/validation/dsl/HmdaDsl.scala
```diff
@@ -0,0 +1,27 @@
...
+      case _ => false
+    }
+    override def failure: String = s"invalid timestamp format"
+  }
+
+  def checkDateFormat[T](s: String): Boolean = {
+    try {
+      val year = s.substring(0, 4)
+      val d = Date.parse(s)
+      d.asInstanceOf[AnyRef] match {
```
schbetsy: Is the purpose of this (lines 19-22) just to test whether `Date.parse` was successful? I'm surprised there's not a cleaner way to do that.

validation/shared/src/main/scala/hmda/validation/engine/ts/CommonTsValidation.scala
```diff
@@ -0,0 +1,74 @@
...
+    for {
+      a <- fs100
+      b <- fs013
+    } yield {
+      (
+        s010(ts)
+        |@| s020(ts)
+        |@| a
+        |@| b
+      )((_, _, _, _) => ts)
```
lgdean: @schbetsy and I wonder whether this tuple will eventually become a problem if there are more than 22 rules in here. We're also curious/confused about exactly what this code is doing, but that may be a conversation for later, and perhaps in real-time communication.

validation/jvm/src/test/scala/hmda/validation/engine/ts/TsValidationSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  property("Transmittal Sheet fails S028 (Timestamp Format)") {
+    forAll(tsGen) { ts =>
+      whenever(ts.id == 1) {
+        val badTs = for {
+          t <- badTimestamp
+          x = ts.copy(timestamp = t)
+        } yield x
+
+        failGenTs(badTs)
```
lgdean: @schbetsy and I noticed that this line doesn't do what we expected.  We thought it was supposed to be testing `TsValidationEngine`, but in fact it calls `CommonTsValidation.validate`.  So it seems that `TsValidationEngine` is not tested at all (even in the JVM version), so `S028` isn't being tested here.

validation/jvm/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
```diff
@@ -0,0 +1,15 @@
...
+import hmda.validation.dsl.{ HmdaDsl, CommonDsl, Result }
+
+/*
+ Timestamp must be numeric and in ccyymmddhhmm format
+ */
+object S028 extends CommonDsl with HmdaDsl {
+  def apply(ts: TransmittalSheet): Result = {
+    import scala.language.postfixOps
+    val timestamp = ts.timestamp
+    (timestamp is numeric) and (timestamp.toString is validTimestampFormat)
```
lgdean: @schbetsy and I noticed that `ts.timestamp` is always numeric, because it is a Long. As the parser currently stands, a TS string with a non-numeric timestamp would fail during parsing. Will the parser eventually be responsible for checks such as S028, or is this a sign that `TransmittalSheet.timestamp` should be a String, so that this check can happen during validation?

validation/js/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
```diff
@@ -0,0 +1,15 @@
+package hmda.validation.rules.ts.syntactical
+
+import hmda.model.fi.ts.TransmittalSheet
+import hmda.validation.dsl.{ Result, HmdaDsl, CommonDsl }
```
lgdean: We noticed that the js and jvm versions of these files are identical, except for the order of imports. Is there any way to move this code to the `shared` directory? (We see that each one depends on the respective version of date-related code, but we don't know whether that dependency prevents sharing the code.)

validation/shared/src/main/scala/hmda/validation/engine/ts/CommonTsValidation.scala
```diff
@@ -0,0 +1,74 @@
...
+      case Failure(msg) => ValidationError(s"S020 failed: $msg").failure.toValidationNel
+    }
+  }
+
+  //TODO: Implement S025 validation rule
+  //  protected def s025(t: TransmittalSheet): TsValidation = {
+  //
+  //  }
+
+  protected def s100(t: TransmittalSheet)(implicit ec: ExecutionContext): Future[TsValidation] = {
```
lgdean: We noticed that `s100` has an implicit `ExecutionContext` and `s013` does not (and seems not to need it, because of the one near the top of the file -- line 16 in this version). Why?

validation/js/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
lgdean: We have a similar observation about `TsValidationEngine.scala`. The two versions have differences, but we suspect that they're not substantive. This duplication seems likely to cause maintenance headaches, so we'd like to avoid it if possible.

validation/shared/src/main/scala/hmda/validation/engine/ts/CommonTsValidation.scala
jmarin: I see the validation being potentially broken down further, depending on type of validation performed. 
For explanation of Scalaz validation [this](https://www.innoq.com/en/blog/validate-your-domain-in-scala/) is a good resource

validation/jvm/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
jmarin: The division between parser and validation is totally arbitrary and defined by how the edits are prescribed in HMDA. If I were to design this from scratch, I would put a lot more intelligence through the type system into the data model / parser. But one design goal has always been "implement all the rules" and "make them flexible". So I think the parser can't be responsible for S028, unless those initial requirements change

validation/js/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
jmarin: I haven't found a way to share these implementations, which while identical call different implementations of the date formatting function, given that we're targeting two runtimes (JVM and JS). Now, this project would be a lot simpler to maintain if we forget about JS. If we keep it there is additional complexity and yes, and additional maintenance burden.

validation/jvm/src/test/scala/hmda/validation/engine/ts/TsValidationSpec.scala
jmarin: Seems like an oversight on my part. Will revisit

jmarin:
Travis has decided to stop working on the Scala.js portion of the project. I'm having second thoughts about having a JS implementation at all given the circumstances.

validation/jvm/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
lgdean: In that case, do we want the parser to return a string for the timestamp? Then S028 could pass or fail.

validation/shared/src/main/scala/hmda/validation/dsl/CommonDsl.scala
```diff
@@ -0,0 +1,76 @@
+package hmda.validation.dsl
+
+trait CommonDsl {
+  implicit class Rule[T](data: T) {
```
lgdean: @schbetsy had (and I like) the idea of calling this class Subject. It looks like it's not actually a Rule (though I can totally see how it might have started out with that intent).

validation/jvm/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
lgdean: This question now appears as its own issue: https://github.com/cfpb/hmda-platform/issues/257

schbetsy:
Any outstanding issues will be addressed when we remove the ScalaJS portion of the project.

Merging.

validation/js/src/main/scala/hmda/validation/rules/ts/syntactical/S028.scala
lgdean: (This is now moot, but I'm adding the comment just so we/I have a record of it.)  I played a bit with ways to remove the duplication, which naturally all boil down to passing some kind of dependency into the rule.  I didn't find a way to do it and preserve the nice DSL-ness of the syntax, but it was still educational (with regard to both scala and this system).  And it might not be the last time we encounter a question of running things differently in different contexts.  It does seem like traits mean you have to decide certain things at compile time.

PR #249, opened 2016-03-08
Test files

jmarin:
Adds test files (DAT and TXT formats)
Fixes a bug with DAT parser (not trimming values when being parsed)
Adds utility to convert DAT to TXT files (streaming)


platform-test/jvm/src/main/scala/util/Dat2Csv.scala
```diff
@@ -0,0 +1,60 @@
...
+import akka.util.ByteString
+import hmda.parser.fi.lar.LarDatParser
+import hmda.parser.fi.ts.TsDatParser
+
+import scala.concurrent.ExecutionContext
+
+/*
+Takes a .DAT file and converts to pipe delimited CSV (2017)
+ */
+object Dat2Csv {
```
schbetsy: Is `Dat2Csv` just for convenience in making test files, or do you expect to use it in the app somewhere?

jmarin: Just a convenience

schbetsy:
Looks good.

Issue #73, opened 2016-02-17
Implement TS DAT parser test


jmarin:
https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/test/scala/hmda/parser/fi/ts/TsDatParserSpec.scala

hkeeler:
Closing this out with the merge of #252.

PR #252, opened 2016-03-14
add test for the no-parent case

lgdean:
#73 
I think this finishes things up, as the actual error cases are explicitly to be handled in later work. Please let me know if I've missed things, though!

### DIFF
--- a/parser/shared/src/test/scala/hmda/parser/util/FITestData.scala
+++ b/parser/shared/src/test/scala/hmda/parser/util/FITestData.scala
@@ -4,6 +4,8 @@ object FITestData {
 
   val tsDAT =
     "100110132019201501171330 201320-1177984  551  SMALL BANK USA, NA            1850 TYSONS BLVD., SUITE 50             MCLEAN                   VA22102     BIGS USA, INC.                412 THIRD AVENUE                        NEW YORK                 NY10012     Bob Smith                     555-555-5555999-999-9999bob.smith@bank.com                                                "
+  val tsDAT_noParent =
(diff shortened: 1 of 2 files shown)


parser/shared/src/test/scala/hmda/parser/util/FITestData.scala
```diff
@@ -4,6 +4,8 @@ object FITestData {
 
   val tsDAT =
     "100110132019201501171330 201320-1177984  551  SMALL BANK USA, NA            1850 TYSONS BLVD., SUITE 50             MCLEAN                   VA22102     BIGS USA, INC.                412 THIRD AVENUE                        NEW YORK                 NY10012     Bob Smith                     555-555-5555999-999-9999bob.smith@bank.com                                                "
+  val tsDAT_noParent =
```
hkeeler: Scala variable names should be camelCase.

[Scala Style Guide](http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods) says:
>Method, Value and variable names should be in lower camel case

hkeeler:
Overall, this looks good, just the one nitpick about the variable name.

parser/shared/src/test/scala/hmda/parser/util/FITestData.scala
lgdean: I see your point. So, tsDATNoParent or tsDATnoParent? The former seems a little hard to skim, but the latter is inconsistent. (So I suppose I'd lean toward the former.)

hkeeler: Agreed.  I think `tsDATNoParent` is better...or `tsDatNoParent`, but then you have to change the other variable name to make it consistent. :)  Either is fine by me.

hkeeler:
Looks good.  Merging...

Closes #73

PR #259, opened 2016-03-18
actionTaken was renamed to actionTakenType

schbetsy:
Rename `actionTaken` to `actionTakenType` in specs to make them pass.


schbetsy:
Travis still fails on ScalaJS related errors, but I fixed the failing parser spec.

PR #260, opened 2016-03-21
Bye scala js

jmarin:
Removes Scala.js from the project configuration and takes this back to 100% JVM based implementation


jmarin:
@schbetsy @lgdean please take a look, this is ready for review. I've removed Scala.js from the project and tweaked a couple of tests a bit. I consider TsValidationEngine a proof of concept still, it will have to be refactored to accommodate a large number of rules and to scale. Thanks

project/HMDABuild.scala
```diff
@@ -37,6 +35,8 @@ object HMDABuild extends Build {
 
   val httpDeps = akkaDeps ++ Seq(akkaHttp, akkaHttpJson, akkaHttpTestkit)
 
+  val scalazDeps = scalaz
```
lgdean: Seems more consistent to have this value be `Seq(scalaz)` (with a corresponding removal on line 78).

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S010.scala
```diff
@@ -1,7 +1,7 @@
 package hmda.validation.rules.ts.syntactical
 
 import hmda.model.fi.ts.TransmittalSheet
-import hmda.validation.dsl.{ Result, CommonDsl }
+import hmda.validation.dsl.{ CommonDsl, Result }
```
lgdean: Looks like you cleaned up imports here and elsewhere. Do we want to have some sort of cleanup step as part of the build/pre-commit process, so we don't get spurious diffs? (Not as part of this PR, and perhaps a discussion for its own issue, but I wanted to say something while thinking about it.)

schbetsy:
Should this PR also remove the ScalaJS plugin for SBT?

https://github.com/jmarin/hmda-platform/blob/bye-scala-js/project/scalajs.sbt

schbetsy:
After the ScalaJS plugin, @lgdean and I give this a :+1:

PR #262, opened 2016-03-23
Validation engine

jmarin:
Light refactoring of the validation engine

### DIFF
--- a/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
+++ b/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
@@ -14,10 +14,10 @@ trait ValidationApi {
     }
   }
 
-  def validateAll[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
+  def validateAll[E, T](checks: List[ValidationNel[E, T]], input: T): ValidationNel[E, T] = {
     checks.sequenceU.map {
       case c :: _ => c
-      case Nil => t
+      case Nil => input
     }
   }
 
--- a/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
+++ b/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
@@ -7,7 +7,7 @@ import scalaz.Scalaz._
 
 trait ValidationApi {
 
-  def convertResult[A](input: A, result: Result, ruleName: String): ValidationNel[ValidationError, A] = {
+  def convertResult[T](input: T, result: Result, ruleName: String): ValidationNel[ValidationError, T] = {
     result match {
       case Success() => input.success
       case Failure(msg) => ValidationError(s"$ruleName  failed: $msg").failure.toValidationNel
--- a/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
+++ b/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
@@ -1,0 +1,24 @@
+package hmda.validation.api
+
+import hmda.validation.dsl.{ Failure, Result, Success }
+import hmda.validation.engine.ValidationError
+import scalaz._
+import scalaz.Scalaz._
+
+trait ValidationApi {
+
+  def convertResult[A](input: A, result: Result, ruleName: String): ValidationNel[ValidationError, A] = {
+    result match {
+      case Success() => input.success
+      case Failure(msg) => ValidationError(s"$ruleName  failed: $msg").failure.toValidationNel
+    }
+  }
+
+  def validateAllT[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
+    checks.sequenceU.map {
+      case c :: _ => c
+      case Nil => t
+    }
+  }
+
+}
--- a/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
+++ b/validation/src/main/scala/hmda/validation/api/ValidationApi.scala
@@ -14,7 +14,7 @@ trait ValidationApi {
     }
   }
 
-  def validateAllT[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
+  def validateAll[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
     checks.sequenceU.map {
       case c :: _ => c
       case Nil => t
--- a/validation/src/main/scala/hmda/validation/engine/ts/TsValidationEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/ts/TsValidationEngine.scala
@@ -1,58 +1,56 @@
 package hmda.validation.engine.ts
 
 import hmda.model.fi.ts.TransmittalSheet
-import hmda.validation.api.TsValidationApi
-import hmda.validation.dsl.{ Failure, Result, Success }
+import hmda.validation.api.ValidationApi
+import hmda.validation.api.ts.TsValidationApi
 import hmda.validation.engine.ValidationError
 import hmda.validation.rules.ts.syntactical._
-
 import scala.concurrent.{ ExecutionContext, Future }
-import scalaz.Scalaz._
 import scalaz._
+import scalaz.Scalaz._
 
-trait TsValidationEngine extends TsValidationApi {
+trait TsValidationEngine extends ValidationApi with TsValidationApi {
 
   type TsValidation = ValidationNel[ValidationError, TransmittalSheet]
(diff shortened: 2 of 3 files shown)


validation/src/main/scala/hmda/validation/api/ValidationApi.scala
```diff
@@ -0,0 +1,24 @@
...
+trait ValidationApi {
+
+  def convertResult[A](input: A, result: Result, ruleName: String): ValidationNel[ValidationError, A] = {
+    result match {
+      case Success() => input.success
+      case Failure(msg) => ValidationError(s"$ruleName  failed: $msg").failure.toValidationNel
+    }
+  }
+
+  def validateAllT[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
```
schbetsy: Could `validateAllT` be renamed to `validateAll`? If not, what does the `T` add?

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
```diff
@@ -0,0 +1,24 @@
+package hmda.validation.api
+
+import hmda.validation.dsl.{ Failure, Result, Success }
+import hmda.validation.engine.ValidationError
+import scalaz._
+import scalaz.Scalaz._
+
+trait ValidationApi {
+
+  def convertResult[A](input: A, result: Result, ruleName: String): ValidationNel[ValidationError, A] = {
```
schbetsy: What do you expect `A` to be? `TransmittalSheet` and `LoanApplicationRecord`?

I have the same question about `T` in the next method in this file. I'm wondering how `A` and `T` are different.

validation/src/main/scala/hmda/validation/engine/ts/TsValidationEngine.scala
```diff
@@ -1,59 +1,57 @@
...
 import hmda.validation.engine.ValidationError
 import hmda.validation.rules.ts.syntactical._
-
 import scala.concurrent.{ ExecutionContext, Future }
-import scalaz.Scalaz._
 import scalaz._
+import scalaz.Scalaz._
 
-trait TsValidationEngine extends TsValidationApi {
+trait TsValidationEngine extends ValidationApi with TsValidationApi {
```
schbetsy: I'm curious: How do you decide when to use `class` vs `object` vs `trait`? Does the fact that `TsValidationEngine` is a `trait` here mean that another class will eventually extend it?

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
jmarin: Yes, this is just semantics. `validateAllT` was referring to "validate all types". I will push a commit changing this.

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
jmarin: A is just a generic type, usually called `type parameters` in Scala. So the signature of this method is saying something along the lines of "for a given input of type `A`, a `Result` and a name, return a `ValidationError` with type parameters `ValidationError` and `A`"

A lot of Scala code uses `T` for generic types, as is done in Java. Some Scala naming conventions prefer to use a capital letter of the English alphabet, starting with `A`. I use `T` (as in "Type") when it's the only type parameter of a function, i.e. 

```scala
def function[T] (i: Int, s: String): T = ???
```
It there is more than one, I usually start with `A`, `B` (but it could also be `K`, `V` if you are doing keys and values for instance), etc. i.e.:

```scala
def function (a: A, b: B) = ???
```

This is all based on convention, we can use one or the other but I agree it should be consistent. Since there are more `T`s than `A`s in the codebase, if it's ok with you I will change the convertResult signature to use a `T`

validation/src/main/scala/hmda/validation/engine/ts/TsValidationEngine.scala
jmarin: This could be a long topic, and there is no one single answer to rule them all, but as a general rule of thumb I tend to think of these in the following way:

* Objects: these are singletons, which means one instance for that particular class. They are static, don't use the `new` keyword to get instantiated. They are the most basic, and are useful to encapsulate data and some functions. 

* Class. This is like a regular Java class (simplifying but it will do for now). You can have many instances of this class, they get instantiated with the `new` keyword and are used a lot if you are doing OO programming with Scala, a bit less if it's more functional (but they are still there; usually the entry point for a program, or some high level abstraction). There are abstract classes (need to be inherited, as in Java), and `case classes`, which are immutable and are one of the nicer features of Scala

* Traits: basically an "interface", with the particular feature that they allow multiple inheritance (unlike say, Java), and can also have an implementation (not just the function signature, but actual implementation that is inherited).

So, when to use which? Well, I guess it depends. Many things can be modelled with objects, specially if they are just data wrappers. The fact that `TsValidationEngine` is a trait is to give it more flexibility. You can see how the test (a ScalaTest class) extends it to have access to its functions and do testing, but that implementation is different from what will end up being the real engine, which will be based on `Akka` actors. Having traits in functions this way is very flexible for testing the business logic. In addition to this, you can start to see how the validation engine as a whole will be "composed" of several of these traits, that we can implement in a bit of isolation if needed. The division is a bit artificial right now (Transmittal Sheet vs LAR, syntactical, validity, etc.) but it's the way the domain is structured right now. Hopefully this rambling wasn't too long and this helped, but let me know otherwise.

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
schbetsy: Thanks for your response. In the spirit of our team discussion yesterday, I wanted to give you feedback: Please assume a higher level of scala knowledge from me, since I've taken a course in it. In this case, I do already know what type parameters are and that `T` is frequently used for them. I do appreciate your willingness to explain! This will be a great way for me to learn concepts in the future.

Maybe I should have asked, what specific types do you expect `T` to be a stand-in for? Or, in other words, where else will `ValidationApi` be used? Will we be validating anything besides `TransmittalSheet`s and `LoanApplicationRecord`s?

jmarin: Given that `T` is a type parameter, it can be "anything". So I don't expect a particular list of specific types to go here. The `convertResult`method was "less generic" initially but was refactored as part of [this](https://github.com/cfpb/hmda-platform/pull/89/commits/434ba98c032ef73c293b11c6f91d3871482f7d0a) commit to reduce boilerplate. 

Besides `TransmittalSheet` and `LoanApplicationRecord` there has been some talk recently of using a similar approach for validating Panel construction. So you can think of another set of "rules" for that process, in that case this could become very useful.

validation/src/main/scala/hmda/validation/engine/ts/TsValidationEngine.scala
schbetsy: This was very helpful, thanks! This definitely helps my understanding of how the `TsValidationEngine` will be used in the future.

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
```diff
@@ -0,0 +1,24 @@
...
+trait ValidationApi {
+
+  def convertResult[T](input: T, result: Result, ruleName: String): ValidationNel[ValidationError, T] = {
+    result match {
+      case Success() => input.success
+      case Failure(msg) => ValidationError(s"$ruleName  failed: $msg").failure.toValidationNel
+    }
+  }
+
+  def validateAll[E, T](checks: List[ValidationNel[E, T]], t: T): ValidationNel[E, T] = {
```
schbetsy: Now that this method is no longer in the Transmittal Sheet context, `t` is confusing as a variable name. Could `t` be renamed to `input`?

schbetsy:
One last comment about `t`, then this LGTM.

validation/src/main/scala/hmda/validation/api/ValidationApi.scala
jmarin: @schbetsy I've renamed the variable as per your suggestion, but `t` in this context could be read as "type". I can see how that can be confused with Transmittal Sheet, so I'm fine with changing it

schbetsy: Ah, I see. If it's idiomatic in Scala to name an instance of a class (like `t`) after the type variable (like `T`), I'm okay with it. But in general, I prefer descriptive names.

PR #263, opened 2016-03-23
V140

jmarin:
Implements V140, further refactors TS validation engine (splits traits by validation type). Adds first round of tests


validation/src/test/scala/hmda/validation/engine/ts/validity/TsValidityEngineSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  property("Transmittal Sheet must be valid") {
+    forAll(tsGen) { ts =>
+      whenever(respondentNotEmpty(ts.respondent)) {
+        validate(ts).isSuccess mustBe true
+      }
+    }
+  }
+
+  property("Transmittal Sheet fails V105 (Respondent Mailing Address)") {
```
lgdean: I believe this line and line 41 (the one naming a test that claims to be about V155) are switched.

validation/src/test/scala/hmda/validation/rules/ts/validity/V140Spec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  property("Wrong or missing respondent state code should fail") {
+    forAll(tsGen) { ts =>
+      val r1 = ts.respondent.copy(state = "")
+      val badTs1 = ts.copy(respondent = r1)
+      val r2 = ts.respondent.copy(state = "XXX")
+      val badTs2 = ts.copy(respondent = r2)
+      whenever(ts.id == 1) {
+        V140(badTs1) mustBe Failure("is not contained in valid values domain")
+        V140(badTs2) mustBe Failure("is not contained in valid values domain")
```
lgdean: Is this in fact the failure message we want? or would we want it to include the actual value provided? Maybe that change would be a future story, not in this PR (as it's not V140-specific), but the question seems worth asking.

validation/src/test/scala/hmda/validation/rules/ts/validity/V140Spec.scala
```diff
@@ -0,0 +1,31 @@
...
+        V140(ts) mustBe Success()
+      }
+    }
+  }
+
+  property("Wrong or missing respondent state code should fail") {
+    forAll(tsGen) { ts =>
+      val r1 = ts.respondent.copy(state = "")
+      val badTs1 = ts.copy(respondent = r1)
+      val r2 = ts.respondent.copy(state = "XXX")
```
lgdean: It'd be nice to have a test (here or elsewhere) covering the case of a 2-digit string that isn't a valid state code. (I'm not gonna insist on it before merging, but it's nice to have a minimally-different example.)

model/src/main/scala/hmda/model/census/Census.scala
```diff
@@ -0,0 +1,65 @@
+package hmda.model.census
+
+object Census {
+  val states = Map(
```
lgdean: At the moment this is more complicated than it needs to be, as the state names aren't used. I could imagine there's a future purpose in mind; is there?

model/src/main/scala/hmda/model/census/Census.scala
```diff
@@ -0,0 +1,65 @@
...
+    "WV" -> "West Virginia",
+    "WI" -> "Wisconsin",
+    "WY" -> "Wyoming",
+    "AS" -> "American Samoa",
+    "DC" -> "District of Columbia",
+    "FM" -> "Federated States of Micronesia",
+    "GU" -> "Guam",
+    "MH" -> "Marshall Islands",
+    "MP" -> "Northern Mariana Islands",
+    "PW" -> "Palau",
```
lgdean: This doesn't match the list of states in `FITestData`.

validation/src/test/scala/hmda/validation/rules/ts/validity/V155Spec.scala
```diff
@@ -12,15 +12,17 @@ class V155Spec extends PropSpec with PropertyChecks with MustMatchers with TsGen
...
 
   property("Invalid emails should be detected") {
-    forAll(badContactGen) { contact =>
-      whenever(contact.email != "") {
-        V155(contact) mustBe Failure("is not a valid email")
+    forAll(tsGen) { ts =>
+      val badContact = badContactGen.sample.getOrElse(Contact("", "", "", ""))
+      val badTs = ts.copy(contact = badContact)
+      whenever(badTs.contact.email != "") {
+        V155(badTs) mustBe Failure("is not a valid email")
```
lgdean: This `whenever
Shouldn't the same failure also happen when the contact email *is* the empty string? It seems worth covering that case too. (I could see wanting to put it into a separate test, but I don't see that test in this file.)

validation/src/test/scala/hmda/validation/rules/ts/validity/ValidityUtils.scala
```diff
@@ -0,0 +1,24 @@
+package hmda.validation.rules.ts.validity
+
+import hmda.model.fi.ts.Respondent
+
+trait ValidityUtils {
+  def respondentNotEmpty(r: Respondent): Boolean = {
```
lgdean: Based on the name, I'd expect this method to be the opposite (that is, the complement) of `respondentIsEmpty`.  In other words, I'm surprised that a Respondent with a name but no address (for example) would have both these functions return false: it's neither empty nor not-empty.  I'm not sure what names would make it clear: empty and full, perhaps?  (Or perhaps empty and not-empty are domain terms that mean exactly what they mean here, in which case, a comment might be in order.)

Also, I don't think all the cases of `V105` are covered, which feels somewhat related.  For example, I can change its first `and` to `or` and still have all the tests pass. (Not part of this PR, perhaps, but the code is here so I'm commenting here.)

model/src/main/scala/hmda/model/census/Census.scala
jmarin: Yes there is. We will need to provide data structures from Census for validation. I'm not sure what it will look like in the end, so this object might change. One possibility is to have a `State`, `County`, etc. object that encapsulates the necessary fields. This is probably going to be refactored at some point

model/src/main/scala/hmda/model/census/Census.scala
jmarin: I think this one includes territories, while the other one didn't. This will be addressed once we define the Census API needs (see above comment).

validation/src/test/scala/hmda/validation/rules/ts/validity/V155Spec.scala
jmarin: Regarding `2`, this is only half true. The generator always generates a non empty Contact with some email string in it, that is correct. The `badContactGen.sample` at line 22 call returns a `Some(contact)` or a `None`. When `None`, the `getOrElse` call chooses an empty Contact. This might take a few runs to have Scalacheck cover that use case, but it does happen.

lgdean: Yeah, that's exactly what happened: I ran the tests and it didn't happen. (Later, I ran `sbt test` 10 times, and it happened in 2 of those times.)  It seem, then, like property-based testing (or at least this particular test) is not the right way to test this rule (or specifically, this behavior).  The empty-string case is a known, expected case, and I'd want it to be covered every time I run the tests.  Otherwise, I can't be confident that running the tests lets me know I haven't broken the code.

On the plus side, that explains why `sample` has the signature it has! I'd wondered, since I had literally never seen it return `None` when I was using it to test (well, visually inspect) generators in the console.

validation/src/test/scala/hmda/validation/rules/ts/validity/V155Spec.scala
```diff
@@ -20,8 +20,9 @@ class V155Spec extends PropSpec with PropertyChecks with MustMatchers with TsGen
   property("Invalid emails should be detected") {
     forAll(tsGen) { ts =>
       val badContact = badContactGen.sample.getOrElse(Contact("", "", "", ""))
+      println(badContact)
       val badTs = ts.copy(contact = badContact)
-      whenever(badTs.contact.email != "") {
+      whenever(badTs.id == 1) {
```
lgdean: I don't understand the purpose of having a `whenever` here at all. It seems like extra code to read, that doesn't affect the behavior of the test. What am I missing?

validation/src/test/scala/hmda/validation/rules/ts/validity/V155Spec.scala
```diff
@@ -12,15 +12,17 @@ class V155Spec extends PropSpec with PropertyChecks with MustMatchers with TsGen
   property("A valid email address for the institution must be provided") {
```
lgdean: This phrase sounds like the test would be about making sure a failure happens when there's not a valid email address.

jmarin: This language is directly extracted from the edit checks document, it's the description of what this rule should do.

validation/src/test/scala/hmda/validation/rules/ts/validity/ValidityUtils.scala
lgdean: I still believe this to be true, but I intend to merge anyway. Perhaps if a later user of the code is confused by the inconsistency, they will have good ideas for what to do about it.

Issue #126, opened 2016-03-08
Validation (2017): V262 (Action Taken - Type)


hkeeler:
**Edit #:** V262
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Action Taken - Type

If date application received = NA, then action taken type must = 6.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

lgdean:
It looks like V262 already has an implementation, but the tests don't yet cover the failure case.  Let's be sure to cover that case before closing this issue.

It also looks like one of the tests will occasionally fail because its `whenever` isn't satisfied before ScalaCheck gives up.  I believe that's what happened in [this failure](https://travis-ci.org/cfpb/hmda-platform/builds/119109068), which was triggered (but not caused) by PR #266 .  We might want to go ahead and fix that separately -- or that might be a good reason to pick this issue up soon and deal with all of it.

lgdean:
code and test cleanup accomplished!

PR #266, opened 2016-03-28
refactor empty Predicate: simplify String case, throw error in others

lgdean:
This PR combines two changes, which is perhaps not ideal, but they are both small and I'd made them days ago (in the context of reviewing one of the larger validation-related PRs), so it seemed good to go ahead and get them in.  The first is pretty clear: simplify an `if (bool-expression) true else false` that had snuck in.

The second might be controversial (but if so, potentially educational about the intent of the code).  The type signature suggests that `empty` is eventually intended to apply to more than just Strings.  That's cool, if so.  (It's easy to imagine how the concept could apply to, say, a Contact or the like.)  However, those cases are not yet implemented.  So, with this change, instead of silently returning `false`, the code explicitly says that those other cases are not yet implemented.

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
@@ -59,7 +59,7 @@ trait CommonDsl {
   def numeric[T]: Predicate[T] = new Predicate[T] {
     override def validate: (T) => Boolean = _.asInstanceOf[AnyRef] match {
       case n: Number => true
-      case _ => false
+      case _ => throw new NotImplementedError("'numeric' doesn't handle string (or other) values yet")
     }
     override def failure: String = s"is not numeric"
   }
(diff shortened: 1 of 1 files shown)


jmarin:
Travis isn't happy about something in this change, needs to be checked (might be a temporary glitch). I'm ok with this change, but if we do this it should also be done [here](https://github.com/cfpb/hmda-platform/blob/master/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala#L62) for consistency

lgdean:
Oh, good point. With the `numeric` predicate, is the idea that it might at some point be used on both Strings and Numbers?

lgdean:
As for Travis, I don't think it's anything in this change; I think it's in the nature of `V262Spec` (if I read the failure correctly). See [comment](https://github.com/cfpb/hmda-platform/issues/126#issuecomment-203568245) on #126 for details.

schbetsy:
LGTM. Is this ready to merge?

PR #267, opened 2016-03-29
Test the failure case for V220

schbetsy:
LoanType must be 1,2,3, or 4.

Addresses #115

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/validity/V220Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/validity/V220Spec.scala
@@ -1,7 +1,9 @@
 package hmda.validation.rules.lar.validity
 
+import hmda.model.fi.lar.{ Loan, LoanApplicationRegister }
 import hmda.parser.fi.lar.LarGenerators
-import hmda.validation.dsl.Success
+import hmda.validation.dsl.{ Failure, Success }
+import org.scalacheck.Gen
 import org.scalatest.prop.PropertyChecks
 import org.scalatest.{ MustMatchers, PropSpec }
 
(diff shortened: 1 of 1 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/V220Spec.scala
```diff
@@ -13,4 +15,18 @@ class V220Spec extends PropSpec with PropertyChecks with MustMatchers with LarGe
...
+  val badLoanTypeGen: Gen[Int] = {
+    val belowRange = Gen.choose(Integer.MIN_VALUE, 0)
+    val aboveRange = Gen.choose(5, Integer.MAX_VALUE)
+    Gen.oneOf(belowRange, aboveRange)
+  }
+
+  property("Loan Type other than 1,2,3,4 is invalid") {
+    forAll(larGen, badLoanTypeGen) { (lar: LoanApplicationRegister, x: Int) =>
+      val invalidLoan: Loan = lar.loan.copy(loanType = x)
+      V220(invalidLoan) mustBe Failure("is not contained in valid values domain")
```
lgdean: This message does not match the "Loan type is missing or is not in range 1-4" that I see in one the 2016 edit document.  I'm guessing that is OK for now, but I'd like to ask explicitly: is that OK for now?  Do we have a plan for making the correct user-facing error messages at some point?

I ask partly because, in the absence of other constraints, it seems easier to implement a rule and its failure message together.  (But there may well be reasons for doing it in a different way.  For example: maybe the 2017 messages won't necessarily match the 2016 ones.  I don't know at the moment, but now is a good time to find out.)

jmarin: One of the things that is being discussed as a potential change is the error messages displayed to the user

jmarin:
@schbetsy Any idea why this PR doesn't include Codecov or Travis checks?

schbetsy:
Strange. No, I don't think I did anything to deactivate the checks. I'll try pushing the code up again.

jmarin:
@schbetsy It worked now. LGTM, merging

validation/src/test/scala/hmda/validation/rules/lar/validity/V220Spec.scala
schbetsy: Where is this being discussed? I'm interested in what options we're considering.

PR #268, opened 2016-03-30
implement V340

lgdean:
#158 

I raised one question in a comment (which we may or may not address now, but if not, doing things this way may become a default).  I also wonder about conventions for naming tests, as these names may not follow any existing one.  (There's should vs must vs just-plain does, for one thing.  Do we want to standardize yet? Ever?)

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
@@ -51,8 +51,8 @@ trait CommonDsl {
     override def failure: String = s"not greater than $that"
   }
 
-  def containedIn[T](list: List[T]): Predicate[T] = new Predicate[T] {
-    override def validate: (T) => Boolean = list.contains(_)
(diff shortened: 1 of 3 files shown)


jmarin:
I would suggest asking that question you have in a comment as a GitHub ticket for better tracking, using the `question` label created for that purpose. The question you are asking is more generic than the changes reflected in this PR.

jmarin:
On the test naming conventions, the discrepancies are a nature of organic growth. This doesn't bother me much, but if you want to prepare a pull request standardizing them that would be great. I would suggest using `MustMatchers` and `must` as language for them.

lgdean:
removed the comment and put its question into #270

schbetsy:
Since the LarValidityEngine is already merged (thanks!) I'd say add this to the engine, then LGTM.

lgdean:
I'd like to add it to the engine in a different PR if that's cool with you (since I have some other validity check work in progress).

schbetsy:
Sounds good. LGTM. @jmarin?

PR #273, opened 2016-03-31
V225 (Loan Purpose)

schbetsy:
Addresses #117


schbetsy:
The coverage goes down because the `LarValidityEngine` isn't tested yet. I've added #276 to make sure that happens.

Issue #160, opened 2016-03-08
Validation (2017): V375 (Type of Purchaser)


hkeeler:
**Edit #:** V375
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Type of Purchaser

If type of purchaser = 2, then loan type must = 2, 3, or 4.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

PR #274, opened 2016-03-31
V375 purchaser loan; V262 (tests and refactoring)

lgdean:
addresses #160 and #126 

with @schbetsy:
Implemented V375, which led to creating a `when` part of the DSL.  Then refactored V262 to use it, updating its tests first.

We'd take feedback on the `when` syntax (as we did try two different versions), or any other aspect.

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala
@@ -74,10 +74,6 @@ trait CommonDsl {
   }
 
   def when(condition: Result)(thenTest: => Result): Result = {
-    condition match {
-      case Success() => thenTest
-      case Failure(_) => Success()
-    }
+    condition.implies(thenTest)
   }
-
 }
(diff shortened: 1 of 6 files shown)


lgdean:
I made a branch with yet another option for `when` syntax.  I'm not convinced I like it better, but it didn't take long, so it seemed worth trying.  (And perhaps others do prefer it!  Opinions welcome.)

jmarin:
Of these two options, I like the one you have here better. [Here](https://github.com/jmarin/hmda-platform/tree/V375-when-then) is another option I quickly put together, tests seem to still pass. Relevant changes in [`CommonDsl`](https://github.com/jmarin/hmda-platform/blob/V375-when-then/validation/src/main/scala/hmda/validation/dsl/CommonDsl.scala#L77) and [`Result`](https://github.com/jmarin/hmda-platform/blob/V375-when-then/validation/src/main/scala/hmda/validation/dsl/Result.scala#L18)

lgdean:
Ooh, very interesting!  We had talked about when/then but hadn't tried it.  But yeah, once you move to having things on `Result` then adding `then` does make it flow in the more-typical English word order, and plus you can go back to having the `=>` which is a nice touch.

lgdean:
I liked the idea of having the logic in `Result` so much that I put it there.  I almost changed that method body to match @jmarin's version, but then I noticed that the failure message in that case would just be "Success() " prepended to the existing failure message, which wouldn't actually make sense.  So I kept it as is, because it's less code.

schbetsy:
LGTM

Issue #116, opened 2016-03-08
Validation (2017): V400 (Property Type)


hkeeler:
**Edit #:** V400
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Property Type

Property type must = 1, 2, or 3.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

schbetsy:
Read to close?

PR #275, opened 2016-04-01
V400 (Property Type)

schbetsy:
Addresses #116


jmarin:
@schbetsy GitHub complains about merge conflicts on this one, might be a result of not having merged master into it.

schbetsy:
Makes sense, since it conflicted with #273. Fixed.

Issue #258, opened 2016-03-18
Rename "Rule" implicit class?


schbetsy:
Up for discussion. See [this comment](https://github.com/cfpb/hmda-platform/pull/89#discussion_r56438023) for context.

We (@lgdean and I) want to suggest `Subject` as a possible new name for the `Rule` implicit class.
Our rationale: In the English sentence "X is numeric", X is the subject of the sentence.

(The reason we think this would be especially valuable is that we were envisioning a refactor where each edit check (e.g. `S010`, `s013`, etc) could inherit from a common class. `Rule` would be a logical name for this class. This could be a separate discussion.)

lgdean:
I am closing this because it seems like there isn't much more discussion to be had at this time.

PR #278, opened 2016-04-01
refactor: rename Rule to Subject

lgdean:
#258 

This turned out to be a one-line change. (I gave it a try, since I am quickly tiring of repetition as I implement validity checks, and am definitely pondering ways to refactor. I suspect `Rule` may not be the right name there either, but we might as well update this one.)

All this talk of rule and subjects feels quite regal!


jmarin:
Not a huge fan of the name change but accepting this anyway

Issue #159, opened 2016-03-08
Validation (2017): V347 (Type of Purchaser)


hkeeler:
**Edit #:** V347
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Type of Purchaser

If type of purchaser = 1, 2, 3, 4, 5, 6, 7, 8, or 9, then action taken type must be 1 or 6. 

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

PR #279, opened 2016-04-02
refactor: remove duplication in (some) engines

lgdean:
The goal here is to get rid of the duplication that causes us to add 4 lines each time we add a validation to `LarValidityEngine` (and others, but that's the current pain point).

It won't solve all the merge conflicts, but it'll at least mean we're adding just one new line for each new edit check.

There's more work to do here, including some really obvious parts, but I'd like to get something merged soonish, so we can stop doing extra work for each validation rule, and also make merges slightly easier.  (That said: it may be a good idea to merge the existing PRs first, then fix this one, and then merge this one -- ideally quickly, before others happen.)

Then, once the immediate `LarValidityEngine` pressure is off, we can finish up the refactoring.  (Assuming, that is, that people like it in the first place!)

### DIFF
--- a/validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala
@@ -3,14 +3,9 @@ package hmda.validation.engine.lar.validity
 import hmda.model.fi.lar.LoanApplicationRegister
 import hmda.validation.api.ValidationApi
 import hmda.validation.engine.lar.LarCommonEngine
-import hmda.validation.rules.EditCheck
 import hmda.validation.rules.lar.validity._
 
 trait LarValidityEngine extends LarCommonEngine with ValidationApi {
-
-  private def doCheck(check: EditCheck[LoanApplicationRegister], lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, check(lar), check.name)
-  }
 
   def validate(lar: LoanApplicationRegister): LarValidation = {
     val checks = List(
--- a/validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala
@@ -3,27 +3,21 @@ package hmda.validation.engine.lar.validity
 import hmda.model.fi.lar.LoanApplicationRegister
 import hmda.validation.api.ValidationApi
 import hmda.validation.engine.lar.LarCommonEngine
+import hmda.validation.rules.EditCheck
 import hmda.validation.rules.lar.validity.{ V262, V220, V225 }
 
 trait LarValidityEngine extends LarCommonEngine with ValidationApi {
-  private def v220(lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, V220(lar.loan), "V220")
-  }
 
-  private def v225(lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, V225(lar), "V225")
-  }
-
-  private def v262(lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, V262(lar), "V262")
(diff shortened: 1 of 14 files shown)


validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala
```diff
@@ -3,27 +3,21 @@ package hmda.validation.engine.lar.validity
...
-    convertResult(lar, V220(lar.loan), "V220")
-  }
-
-  private def v225(lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, V225(lar), "V225")
-  }
 
-  private def v262(lar: LoanApplicationRegister): LarValidation = {
-    convertResult(lar, V262(lar), "V262")
+  private def doCheck(editCheck: EditCheck[LoanApplicationRegister], lar: LoanApplicationRegister): LarValidation = {
```
jmarin: Could we change this to `check`? I think it reads better and is more in line with the rest of the codebase.

validation/src/main/scala/hmda/validation/engine/ts/validity/TsValidityEngine.scala
```diff
@@ -2,29 +2,25 @@ package hmda.validation.engine.ts.validity
...
+import scalaz._
 
-  private def v140(t: TransmittalSheet): TsValidation = {
-    convertResult(t, V140(t), "V140")
-  }
+trait TsValidityEngine extends TsCommonEngine with ValidationApi {
 
-  private def v155(t: TransmittalSheet): TsValidation = {
-    convertResult(t, V155(t), "V155")
+  private def doCheck[T](editCheck: EditCheck[T], input: T): ValidationNel[ValidationError, T] = {
```
jmarin: Same comment as above, plus an additional observation. This method uses a parameterized type, while the one in `LarValidityEngine` doesn't. Why the difference? Is this method a candidate for abstraction for both? (i.e. in `ValidationApi`)

validation/src/main/scala/hmda/validation/rules/EditCheck.scala
```diff
@@ -0,0 +1,10 @@
+package hmda.validation.rules
+
+import hmda.validation.dsl.{ CommonDsl, Result }
+
+abstract class EditCheck[T] extends CommonDsl {
+
+  def name: String
+
+  def apply(lar: T): Result
```
jmarin: Variable name should probably be called something other than `lar`?

validation/src/main/scala/hmda/validation/engine/ts/validity/TsValidityEngine.scala
lgdean: Mostly because it was the end of the day/week in my timezone. This is definitely some of the duplication I mentioned. I'm happy to go ahead and fix it if frees things up to let merging happen. (But these lines are pretty far from the ones that are likely to cause conflicts and such, so I could see it going either way.)

validation/src/main/scala/hmda/validation/rules/EditCheck.scala
lgdean: Done. (I went with `input` for lack of another idea, but would take suggestions.)

Do you have thoughts on where this class should live? I thought about putting it in the same package with `ValidationApi` (since yes, I agree that the function currently called `doCheck` will end up there), but then it felt weird to have that package refer to the DSL, and... well, that's as far as I got with thinking about it, as I decided to focus first on the class-level stuff.

lgdean:
OK, I now have this in a state where it can be merged automatically. I would like to do that, even though I haven't yet addressed the duplication between the two engines, because then we can continue work on this refactoring independently, without worrying (as much) about conflicts from additional edits as we keep implementing them. (In particular, then I can bring these changes into the branch where I'm implementing `V347` (for #159) and then be done with that one.)

That said, I am open to hearing other concerns too.

schbetsy:
LGTM

PR #280, opened 2016-04-04
Add persistence to file upload

jmarin:
This PR introduces Akka Persistence to the project. It is fairly limited in scope so far, and does not address all the persistence requirements. It only focuses on storing the uploaded HMDA file into persistent storage as a series of events that contain text data in raw format while at the same time updating state about the filing process (i.e. number of records uploaded).


hkeeler:
Looks good.  Merging...

For those following along, the official Akka docs for [Persistence](http://doc.akka.io/docs/akka/2.4.3/scala/persistence.html) and [Testing Actors](http://doc.akka.io/docs/akka/2.4.3/scala/testing.html) (especially the [Testkit](http://doc.akka.io/docs/akka/2.4.3/scala/testing.html#Asynchronous_Integration_Testing_with_TestKit) section) were very helpful to understanding how all of this works.

PR #282, opened 2016-04-05
V347 (Type of Purchaser)

lgdean:
addresses #159

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/validity/PurchaserTypeUtils.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/validity/PurchaserTypeUtils.scala
@@ -1,0 +1,16 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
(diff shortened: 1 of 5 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/PurchaserTypeUtils.scala
```diff
@@ -0,0 +1,16 @@
+package hmda.validation.rules.lar.validity
```
schbetsy: Making a separate utility for generating bogus data makes sense (especially since it will probably be useful for future edit check tests). I like it.

I haven't thought through where they should live. Want to tell me about your reasoning for putting it here?

lgdean: I put it there (and with that name) because at the moment, that's the only place where it's used, and the only kind of thing it contains. I would definitely be open to generalizing it and/or putting it somewhere more obvious, though.

schbetsy:
LGTM

PR #283, opened 2016-04-05
Add V340 and V375

lgdean:
should let us close out #158 and #160 

Note that this branch also has the changes from PR #282 (since these changes are in the very same lines of a changed file) so maybe do that one first.  (Or, review all three in this one, and close out #159 here, all in one go.)


schbetsy:
LGTM

Issue #276, opened 2016-04-01
Test LarValidityEngine


schbetsy:
We were in a hurry to add the LAR Validity Engine a few days ago, and did so without test coverage.

Add tests to cover the `LarValidityEngine`. They should ensure that each check that should get run does get run. Refer to #271 for a discussion of test record strategy.

schbetsy:
I checked out the ScalaMock library as a way to test the Lar Validity Engine without duplicating the logic of the individual edits. I hoped to override each check individually to return a failure and make sure that that would result in the engine returning a failure. Unfortunately, in order to use mocking the way it is intended to be used, we would have to architect the validity engine quite differently.

I'm open to suggestions about how else to implement validation engine tests that aren't redundant to the tests of individual edits.

schbetsy:
Here are the edits that should be covered by the engine:

- [ ] V210
- [ ] V215
- [ ] V220
- [ ] V225
- [ ] V230
- [ ] V250
- [ ] V255
- [ ] V260
- [ ] V262
- [ ] V265
- [ ] V275
- [ ] V280
- [ ] V285
- [ ] V290
- [ ] V295
- [ ] V300
- [ ] V310
- [ ] V315
- [ ] V317
- [ ] V320
- [ ] V325
- [ ] V326
- [ ] V330
- [ ] V335
- [ ] V338
- [ ] V340
- [ ] V347
- [ ] V355
- [ ] V360
- [ ] V375
- [ ] V385
- [ ] V400
- [ ] V410
- [ ] V415
- [ ] V425
- [ ] V430
- [ ] V435
- [ ] V440
- [ ] V445
- [ ] V447
- [ ] V450
- [ ] V455
- [ ] V460
- [ ] V463
- [ ] V465
- [ ] V470
- [ ] V475
- [ ] V480
- [ ] V485
- [ ] V490
- [ ] V495
- [ ] V500
- [ ] V505
- [ ] V520
- [ ] V525
- [ ] V535
- [ ] V540
- [ ] V545
- [ ] V550
- [ ] V555
- [ ] V560
- [ ] V565
- [ ] V570
- [ ] V575

PR #284, opened 2016-04-05
Engine

jmarin:
This PR adds a single validation engine for both Transmittal Sheet and Loan Application Register. It does so by combining the existing validation engines (syntactical and validity), refactoring along the way to make this possible. When possible, tests are also added (missing in aggregate LAR validation engine, will be covered once we have test files)

### DIFF
--- a/validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
@@ -16,7 +16,7 @@ trait LarSyntacticalEngine extends LarCommonEngine with ValidationApi {
     validateAll(checks, lar)
   }
 
-  def checkSyntacticalIterable(lars: Iterable[LoanApplicationRegister]): LarsValidation = {
+  def checkSyntacticalCollection(lars: Iterable[LoanApplicationRegister]): LarsValidation = {
     val checks = List(
       S011,
       S040
--- a/validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
@@ -1,0 +1,28 @@
+package hmda.validation.engine.lar.syntactical
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.api.ValidationApi
+import hmda.validation.engine.lar.LarCommonEngine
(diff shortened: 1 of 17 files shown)


validation/src/test/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngineSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+    extends PropSpec
+    with PropertyChecks
+    with MustMatchers
+    with LarGenerators
+    with LarSyntacticalEngine {
+
+  property("A LAR must pass syntactical checks") {
+    forAll(larGen) { lar =>
+      whenever(lar.id == 2) {
+        checkSyntactical(lar).isSuccess mustBe true
```
schbetsy: We'll need a test for `checkSyntacticalIterable` too. If it's not feasible to add it now, we could add an issue for it, or add a note to #276.

jmarin: I will add it here, thanks

validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
```diff
@@ -0,0 +1,28 @@
...
+  def checkSyntactical(lar: LoanApplicationRegister): LarValidation = {
+    val checks = List(
+      S010,
+      S020
+    ).map(check(_, lar))
+
+    validateAll(checks, lar)
+  }
+
+  def checkSyntacticalIterable(lars: Iterable[LoanApplicationRegister]): LarsValidation = {
```
schbetsy: Could we rename this method so it doesn't expose the implementation? Maybe we could use `Collection` instead of `Iterable`.

lgdean: Could they both be `checkSyntactical` (with different signatures)?  (I'm not sure I like that idea, but I'll still put it out there.)

To me `Collection` still sounds like it refers to implementation. (I blame Java.)

Hm. Most things I can think of that read well (to me) end up either involving a lot of words, or having the Validity and Syntactical engines both be classes that get instantiated, rather than traits that get mixed in. (That is, they end up having names that would be the same for both of those engines.)

jmarin: I'm going to go with `Collection`, since it's a pretty common English term anyway

Issue #168, opened 2016-03-08
Validation (2017): V575 (Rate Spread)


hkeeler:
**Edit #:** V575
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Rate Spread

If lien status = 2, then rate spread must be ≥ 3.5% and ≤ 99.99% or NA.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

lgdean:
could've closed this one last sprint; closing now. there's code cleanup to do, but it's happening as part of #167 because they are so similar.

lgdean:
(I apparently didn't close this when I said I was. This time I am definitely hitting the "Close and comment" button very carefully.)

PR #285, opened 2016-04-07
V575 (Rate Spread)

lgdean:
addresses #168 

I will likely refactor a bit when implementing similar/related rules (particularly V570), but this can stand on its own if needed. (And merging it means that we won't keep having to deal with differing commas in `LarValidationEngine` which is a nice bonus.)


jmarin:
LGTM

PR #287, opened 2016-04-07
Single row api

jmarin:
Adds single LAR validation HTTP API. It currently returns `List[ValidationError]` serialized to `JSON` (this needs to be augmented with more details in the future, but it's what we currently have in the model). Current API is as follows:

* POST `/lar/parse` accepting a single row in `CSV` pipe delimited format. Returns LAR formatted in `JSON`
* POST `/lar/validate` accepts LAR in `JSON` format (see above) and returns list of validation errors, if present. Can be filtered with optional query parameter `check` to return only `syntactical` or `validity` errors


api/src/test/resources/logback.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
```
lgdean: I noticed that this puts a bunch of stuff into the terminal when `sbt test` runs.  Can we have it go into a file instead, so there's not as much clutter in with the test output?  (Right now the output actually seems to come from logback reporting its own configuration.)

Also, it's warning that it finds `logback.xml` in two places.

I'm seeing similar output from slf4j, too.

api/src/main/scala/hmda/api/processing/lar/SingleLarValidation.scala
```diff
@@ -0,0 +1,59 @@
...
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.engine.ValidationError
+import hmda.validation.engine.lar.LarEngine
+
+object SingleLarValidation {
+  def props: Props = Props(new SingleLarValidation)
+
+  case class CheckLar(lar: LoanApplicationRegister)
+  case class CheckSyntacticalLar(lar: LoanApplicationRegister)
+  case class CheckValidityLar(lar: LoanApplicationRegister)
```
lgdean: I'd expect these to be called something like `CheckLarValidity` (etc).  Or, since they're already in the context of an object with `Lar` in its name, then perhaps even just `CheckValidity`, `CheckSyntactical`, and `Check` (or `CheckEverything`). Is there a reason it's helpful to keep repeating the `Lar` part?

api/src/main/scala/hmda/api/processing/lar/SingleLarValidation.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    case _ =>
+      log.error(s"Unsupported message sent to ${self.path}")
+  }
+
+  private def validationErrors(lar: LoanApplicationRegister, validationType: ValidationType): List[ValidationError] = {
+    val validation = validationType match {
+      case LarFull => validateLar(lar).disjunction
+      case LarSyntactical => checkSyntactical(lar).disjunction
+      case LarValidity => checkValidity(lar).disjunction
```
lgdean: Could each ValidationType delegate to the appropriate function? It seems like this block mirrors the one above without adding anything.

api/src/main/scala/hmda/api/processing/lar/SingleLarValidation.scala
```diff
@@ -0,0 +1,59 @@
...
+    val validation = validationType match {
+      case LarFullValidation => validateLar(lar).disjunction
+      case LarSyntacticalValidation => checkSyntactical(lar).disjunction
+      case LarValidityValidation => checkValidity(lar).disjunction
+    }
+    if (validation.isRight) {
+      Nil
+    } else {
+      val lErrors = validation.toEither.left.get
+      val errors = lErrors.head :: lErrors.tail.toList
```
lgdean: This line could be just `lErrors.list.toList` (good find, @hkeeler). Any reason not to?

api/src/main/scala/hmda/api/processing/lar/SingleLarValidation.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    case _ =>
+      log.error(s"Unsupported message sent to ${self.path}")
+  }
+
+  private def validationErrors(lar: LoanApplicationRegister, validationType: ValidationType): List[ValidationError] = {
+    val validation = validationType match {
+      case LarFullValidation => validateLar(lar).disjunction
+      case LarSyntacticalValidation => checkSyntactical(lar).disjunction
+      case LarValidityValidation => checkValidity(lar).disjunction
```
lgdean: It seems like this block mirrors the one above without adding anything.
Could each ValidationType delegate to the appropriate function instead?

Or you could pass each method in above, and change this method to:
```scala
  private def validationErrors(lar: LoanApplicationRegister, checker: LoanApplicationRegister => LarValidation): List[ValidationError] = {
    val validation = checker(lar).disjunction
```
and then `ValidationType` is unused.

(Note: I made a similar comment before, but GitHub hides it because it's outdated, so I said it again, plus an additional code option.)

jmarin: I'm not sure I follow. Can you send a pull request with the changes you are proposing?

hkeeler:
@lgdean and I paired on a review yesterday.  I think with the latest commit, the code is in pretty good shape.

The only question I have at this point is with the API design itself.  Is it really useful to have a `/parse` resource?  What's the use case for just parsing?  Our _current_ use is validating the pipe-delimited format, which requires two calls to validate in this scheme.  It seems more useful to `/validate` with multiple formats.  We could do this via different `Content-Type`s (`application/json` and...`text/plain`? (there is no official mime for pipe-delimited)).  We'd probably also want a default, though not sure which that would be.

Thoughts?  If this seems like something we want to do, we could move this over to a separate issue so we can get this one merged as-is now.

hkeeler:
Moved discussion on supporting multiple formats for validation to https://github.com/cfpb/hmda-platform/issues/299.

Merging.

PR #294, opened 2016-04-25
V290

### LINKED ISSUE #135: Validation (2017): V290 (MSA/MD,State,County Codes)
**Edit #:** V290
**Type:** Validity
**Column:** MSA/MD,State,County Codes

If MSA/MD does not = NA, then MSA/MD, state, and county codes must = a valid combination.

Additional resources:

jmarin:
Closes #135


validation/src/test/scala/hmda/validation/rules/lar/validity/V290Spec.scala
```diff
@@ -0,0 +1,42 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.validation.dsl.{ Failure, Success }
+import hmda.validation.rules.lar.LarEditCheckSpec
+
+class V290Spec extends LarEditCheckSpec {
+
+  property("Succeeds when valid MSA/MD, state and county codes are present") {
+    forAll(larGen) { lar =>
+      whenever(lar.geography.msa != "NA") {
```
lgdean: This line seems unnecessary, as the next line sets the geography values appropriately, including `msa`.

lgdean:
It's pretty late here, so I'll review the rest tomorrow, unless someone else picks it up first.

validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
```diff
@@ -1,17 +1,23 @@
...
-  property("Engine returns success for valid LAR") {
-    //TODO: test using a file of sample LARs, since the generator does not
-    //  produce records that pass all edit checks
+  "LAR Validity engine" must {
+    "pass validation on valid sample file" in {
+      val lines = Source.fromFile(new File("parser/src/test/resources/txt/FirstTestBankData_clean_407_2017.txt")).getLines()
+      val lars = lines.drop(1).map(l => LarCsvParser(l))
+
+      lars.foreach { lar =>
+        val validity = checkValidity(lar)
```
schbetsy: This line seems unnecessary, since `validity` isn't used.

validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
```diff
@@ -1,17 +1,23 @@
...
 
-class LarValidityEngineSpec extends PropSpec with PropertyChecks
-    with MustMatchers with LarGenerators with LarValidityEngine {
+class LarValidityEngineSpec extends WordSpec with MustMatchers with LarValidityEngine {
 
-  property("Engine returns success for valid LAR") {
-    //TODO: test using a file of sample LARs, since the generator does not
-    //  produce records that pass all edit checks
+  "LAR Validity engine" must {
+    "pass validation on valid sample file" in {
```
schbetsy: Thank you for adding a spec for this engine. It's valuable and will improve our test coverage.

But, I don't think this spec is sufficient to close #276. One of the requirements of that issue is "ensure that each check that should get run does get run."

model/src/main/scala/hmda/model/census/CBSATractLookup.scala
```diff
@@ -0,0 +1,53 @@
...
+}
+
+case class CBSATract(
+  name: String,
+  metDivName: String,
+  state: String,
+  countyFips: String,
+  county: String,
+  tracts: String,
+  geoidMsa: String,
```
lgdean: I found `geoIdMsa` a little easier to read.  With this version (when I saw it used elsewhere) I read it as geoid, which is also a word but different.

validation/src/test/scala/hmda/validation/rules/lar/validity/V290Spec.scala
```diff
@@ -0,0 +1,40 @@
...
+      val validGeography = lar.geography.copy(msa = "48424", state = "12", county = "086")
+      val validLar = lar.copy(geography = validGeography)
+      V290(validLar) mustBe a[Success]
+    }
+  }
+
+  property("Fails for invalid combination of MSA/MD, state and county") {
+    forAll(larGen) { lar =>
+      val invalidGeography = lar.geography.copy(msa = "17020")
+      val invalidLar = lar.copy(geography = invalidGeography)
```
lgdean: Couldn't this test accidentally pass, if the other two values happened to match this MSA? (Unlikely, yes, but seems possible and would be confusing if it happened.)

validation/src/test/scala/hmda/validation/rules/lar/validity/V290Spec.scala
```diff
@@ -0,0 +1,40 @@
...
+      val validLar = lar.copy(geography = validGeography)
+      V290(validLar) mustBe a[Success]
+    }
+  }
+
+  property("Succeeds when code is MD (not MSA) with valid state and county combination") {
+    forAll(larGen) { lar =>
+      val validGeography = lar.geography.copy(msa = "48424", state = "12", county = "086")
+      val validLar = lar.copy(geography = validGeography)
+      V290(validLar) mustBe a[Success]
```
lgdean: I want to make sure I understand the business logic correctly.  It looks like, if the LAR has an MD instead of an MSA, then it's OK to have a state+county combination that matches anything within the MSA that contains that MD -- even if the county is outside of the given MD.

So in this example, we have MD 48424, which is for county 099 (Palm Beach County), in state 12 (Florida).  Its larger MSA is 33100, which also includes counties 011 (Broward) and 086 (Miami-Dade), in their own separate MDs.

This test says that it's OK for a LAR to have MD 48424 with state 12 and county 086, because the MSA matches, even though the specific MD doesn't.  Am I understanding correctly, and is this the intended business logic?

If so, then please:
 1. add a test for the "normal" case of `12|086|48424`, and
 2. maybe say something a bit more explicit in the description of this test, because it took me a while to figure out what it meant.

If not, then let's discuss here or in real-time chat tomorrow, so I can understand the requirement and finish the review.

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
```diff
@@ -0,0 +1,38 @@
...
+
+    val combination = (state, county, msa)
+
+    when(msa not equalTo("NA")) {
+      combination is containedIn(validCombinations)
+    }
+  }
+
+  private def msaCode(code: String): String = {
+    val md = cbsaTracts.filter(m => m.metDivFp == code)
```
lgdean: I'm a little concerned about the performance of this search (and the one above). 70k is not a huge number, but it's not tiny either. It might be premature to optimize it now, but we might want to keep a list of things to investigate first if performance becomes a problem when we start measuring it.

validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
jmarin: Removed the closes comment for #276

model/src/main/scala/hmda/model/census/CBSATractLookup.scala
jmarin: `geoidMSA` is how the lookup table had it, didn't bother me at all but I've changed this to `geoIdMsa`

validation/src/test/scala/hmda/validation/rules/lar/validity/V290Spec.scala
jmarin: Hasn't happened in my tests, and it's extremely unlikely. I've changed it anyway

validation/src/test/scala/hmda/validation/rules/lar/validity/V290Spec.scala
jmarin: After checking with @lesgou, the logic should be:

1. If MSA/MD = "NA", edit check passes
2. If MSA/MD != "NA, check for valid (state, county, msa) combination or valid (state, county, md) combination
I've updated the logic to reflect this

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
jmarin: I am not concerned about this at this point. I would be surprised if this became our bottleneck. For some context, the HMDA pilot had this loaded into a `MongoDB` database and this edit check required a REST API call. That implementation was fast enough for 98% of HMDA filers according to their tests. Avoiding the network hop is already making this much faster. The tradeoff is memory of course. In any case I agree that we will need to measure when the system is deployed (in a cluster)

PR #296, opened 2016-04-27
V295

### LINKED ISSUE #136: Validation (2017): V295 (State/County Codes)
**Edit #:** V295
**Type:** Validity
**Column:** State/County Codes

State and county must = a valid combination or (county = NA where MSA/MD = NA).

Additional resources:

jmarin:
Closes #136


validation/src/main/scala/hmda/validation/rules/lar/validity/V295.scala
```diff
@@ -0,0 +1,38 @@
...
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V295 extends EditCheck[LoanApplicationRegister] {
+
+  val cbsaTracts = CBSATractLookup.values
+
+  val validCombination = cbsaTracts.map { cbsa =>
+    (cbsa.state, cbsa.county)
```
schbetsy: Here's another point that we could optimize in the future if performance becomes an issue. 

Since the CBSA tracts will not change frequently, we could generate and save the set of valid combinations beforehand, instead of generating it each time this validation runs. If we did that, we could remove the duplicates, making the search faster.

schbetsy:
LGTM

PR #297, opened 2016-04-27
V285

### LINKED ISSUE #134: Validation (2017): V285 (State Code)
**Edit #:** V285
**Type:** Validity
**Column:** State Code

State must = a valid FIPS code or (NA where MSA/MD = NA).

Additional resources:

jmarin:
Closes #134


parser/src/test/scala/hmda/parser/fi/lar/LarGenerators.scala
```diff
@@ -91,8 +91,7 @@ trait LarGenerators extends FIGenerators {
       "30", "31", "32", "33", "34", "35", "36", "37", "38", "39",
       "40", "41", "42", "44", "45", "46", "47", "48", "49",
       "50", "51", "53", "54", "55", "56",
-      "60", "66", "69",
-      "72", "74", "78"
```
schbetsy: I assume you removed 69, 74, and 78 (Northern Mariana Islands, U.S. Minor Outlying Islands, and U.S. Virgin Islands) because they are not contained in the CBSA mapping. So now this list deviates from the list linked in the comment above ([census.gov](http://www2.census.gov/geo/docs/reference/state.txt))

Could you update the comment to say what your source is? I'm also interested in why the list deviates from the [list of FIPS state codes](https://en.wikipedia.org/wiki/Federal_Information_Processing_Standard_state_code), but that may be a question for a business analyst.

lesgou: Yes, not all territories are HMDA filers - only the 50 states + PR + DC.

schbetsy: @lesgou I see. So the CBSA contains every state code that can file HMDA? And it doesn't have any extra locations that shouldn't file HMDA?

@jmarin Would you update the comment on line 86 of this file to reflect where the list actually comes from?

jmarin: @schbetsy I've updated the comment to reflect what @lesgou said in hers. The list comes from Census, but only states, DC and PR need to file. I would assume that is written somewhere in the HDMA rule.

validation/src/main/scala/hmda/validation/rules/lar/validity/V285.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object V285 extends EditCheck[LoanApplicationRegister] {
+
+  val cbsaTracts = CBSATractLookup.values
+
+  override def name: String = "V285"
+
+  override def apply(input: LoanApplicationRegister): Result = {
+    val state = input.geography.state
+    val stateCodes = cbsaTracts.map(c => c.state)
```
lgdean: @schbetsy noticed that this could be `cbsaTracts.map(c => c.state).distinct` instead, and could be declared on line 11, outside the `apply` method.  We tried it out, and it did make the tests run noticeably faster (1 second) on the command line.

schbetsy:
I wanted to rephrase my [poorly-phrased comment](https://github.com/cfpb/hmda-platform/pull/297#discussion_r61791716) from earlier, now that I've thought about it a bit more:

The text of edit V285 (#134) says "State must = a valid FIPS code", but this code uses the CBSA tract mapping, rather than a canonical list of FIPS codes. Is that an issue?

validation/src/test/scala/hmda/validation/rules/lar/validity/V285Spec.scala
```diff
@@ -0,0 +1,47 @@
...
+    forAll(larGen) { lar =>
+      val validGeography = lar.geography.copy(state = "NA", msa = "NA")
+      val validLar = lar.copy(geography = validGeography)
+      V285(validLar) mustBe a[Success]
+    }
+  }
+
+  property("Fails when state is NA but MSA/MD is not NA") {
+    forAll(larGen) { lar =>
+      whenever(lar.geography.msa != "NA") {
```
lgdean: Running on the command line, I just got this failure:
[info] - Fails when state is NA but MSA/MD is not NA *** FAILED ***
[info]   Gave up after 1 successful property evaluation. 10 evaluations were discarded.

It's unlikely, but it did happen.

validation/src/main/scala/hmda/validation/rules/lar/validity/V285.scala
lgdean: or `toSet` for that matter.

validation/src/test/scala/hmda/validation/rules/lar/validity/V285Spec.scala
jmarin: @schbetsy This should not be an issue. The CBSA lookup table uses the same list of states, while the [MSA](http://www.census.gov/population/metro/) is defined by [OMB](https://www.whitehouse.gov/omb/inforeg_statpolicy) the list of geographies that conform them is a list of counties, identified by their FIPS code, i.e. example [here](http://www.census.gov/population/metro/files/lists/2015/List1.xls)

jmarin: @lgdean I've tweaked the configuration of the Property Checks to decrease the likelihood of this happening. In theory it could still happen, but I got tired of rerunning tests and not seeing failures on any tests for the validation project.

schbetsy:
LGTM

PR #298, opened 2016-04-27
V470

### LINKED ISSUE #145: Validation (2017): V470 (Applicant Race)
**Edit #:** V470
**Type:** Validity
**Column:** Applicant Race

If applicant race = 1-5 in applicant race field 1, then all other applicant race fields must = blank or 1-5.

Additional resources:

schbetsy:
Closes #145 

As usual, I'm looking for feedback on how to make my Scala more idiomatic/efficient/correct.

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V470.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V470.scala
@@ -1,0 +1,34 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V470 extends EditCheck[LoanApplicationRegister] {
+
+  def apply(lar: LoanApplicationRegister): Result = {
+    val applicant = lar.applicant
+    when(applicant.race1 is containedIn(1 to 5)) {
+      (validRace(applicant.race2) is equalTo(true)) and
+        (validRace(applicant.race3) is equalTo(true)) and
+        (validRace(applicant.race4) is equalTo(true)) and
+        (validRace(applicant.race5) is equalTo(true))
+    }
+  }
+
+  private def validRace(input: String): Boolean = {
+    (input == "") || inRange(1 to 5, input)
+  }
+
+  private def inRange(domain: Seq[Int], input: String): Boolean = {
+    try {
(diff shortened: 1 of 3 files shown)


validation/src/main/scala/hmda/validation/rules/lar/validity/V470.scala
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  private def inRange(domain: Seq[Int], input: String): Boolean = {
+    try {
+      val x = input.toInt
+      domain.contains(x)
+    } catch {
+      case ex: NumberFormatException => false
+    }
+  }
```
jmarin: @schbetsy Nothing wrong with your approach, but I think your implementation can be made a bit more idiomatic. Here is a somewhat lengthy explanation of what I would propose. 

_Usually_, the `try...catch` approach to exception handling is an anti pattern in `Scala`. There are cases where you have to use it, but in this instance you are not doing anything with the exception, just capturing it and returning `false`. In general, optional values / things that can throw exceptions are usually managed in `Scala` by using `Option`, `Try` or `Either` (all of them available in the standard Scala library). Here is a more idiomatic way using `Try`, which is the one that makes most sense here: 

```scala
private def inRange(domain: Seq[Int], input: String): Try[Boolean] = {
  Try(domain.contains(input.toInt))
}
```

`Try` returns a `Success[A]` or a `Failure[Throwable]`, and it has "monadic" properties, which means we can get to the wrapped value in a consistent way using `getOrElse`, basically the way it works is if it has a successful value (`Success[A]`) it will return the wrapped value, if it is a `Failure` it will return what you pass as a parameter (`false` in this case) . Your validRace function would now look like this:

```scala
private def validRace(input: String): Boolean = {
  (input == "") || inRange(1 to 5, input).getOrElse(false)
}
```

schbetsy: Thanks for the thorough explanation, that was really helpful. 👍 

 I implemented your suggestion in 8bf2e61.

jmarin:
Done reviewing, LGTM. I'd like to wait until #294 gets merged in, it includes more testing coverage with real data (see [this ] (https://github.com/jmarin/hmda-platform/blob/V290/validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala))

PR #300, opened 2016-04-27
V280

### LINKED ISSUE #133: Validation (2017): V280 (MSA/MD Number)
**Edit #:** V280
**Type:** Validity
**Column:** MSA/MD Number

MSA/MD must = a valid Metropolitan Statistical Area or Metropolitan Division (if appropriate) code for period being processed or NA.

Additional resources:

jmarin:
Closes #133


validation/src/main/scala/hmda/validation/rules/lar/validity/V280.scala
```diff
@@ -0,0 +1,29 @@
...
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V280 extends EditCheck[LoanApplicationRegister] {
+
+  val cbsaTracts = CBSATractLookup.values
+
+  val validMSAs = cbsaTracts.map(cbsa => cbsa.geoIdMsa)
+
+  val validMDs = cbsaTracts.map(cbsa => cbsa.metDivFp)
```
lgdean: I played with this and put `.toSet` at the end of these two lines, and added an appropriate method to CommonDsl (taking a `Set` instead of a `Seq`), and found that it made the tests over a second faster.  This is probably worth doing.

There may be a way to do it with just one `containedIn` that works for both cases; I didn't spend much time trying.

Also, I notice we're gaining a lot of code that uses the `cbsaTracts` data in various ways, in various places.  And certainly we can safely do this, thanks to immutable data structures, but I wonder whether it'd be nicer to centralize them.

lgdean:
LGTM; I'll create a separate PR for the change suggested above.

PR #302, opened 2016-04-28
V570 and V520 rate spread

### LINKED ISSUE #167: Validation (2017): V570 (Rate Spread)
**Edit #:** V570
**Type:** Validity
**Column:** Rate Spread

If lien status = 1, then rate spread must be > 1.5% and ≤ 99.99% or NA.

Additional resources:

### LINKED ISSUE #169: Validation (2017): V520 (Rate Spread)
**Edit #:** V520
**Type:** Validity
**Column:** Rate Spread

If lien status = 3, then rate spread must = NA

Additional resources:

lgdean:
closes #169 
closes #167


validation/src/test/scala/hmda/validation/rules/lar/validity/V570Spec.scala
```diff
@@ -0,0 +1,47 @@
...
+
+class V570Spec extends RateSpreadEditCheckSpec {
+
+  override val check = V570
+
+  property("succeeds if lien status is 1 and rate spread is NA") {
+    succeedsWhen(lienStatus = 1, rateSpread = "NA")
+  }
+
+  property("fails if lien status = 1 and rate spread is < 1.5%") {
```
schbetsy: Should there be an equivalent test to show that the check fails if the rate spread is `"100"` or greater?

validation/src/main/scala/hmda/validation/rules/lar/validity/V570.scala
```diff
@@ -0,0 +1,21 @@
...
+import hmda.validation.dsl.{ Failure, Result }
+import hmda.validation.rules.EditCheck
+
+object V570 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V570"
+
+  override def apply(lar: LoanApplicationRegister): Result = {
+    when(lar.lienStatus is equalTo(1)) {
+      lar.rateSpread is equalTo("NA") or {
+        try {
```
schbetsy: According to @jmarin's [comment on my recent PR](https://github.com/cfpb/hmda-platform/pull/298#discussion_r61296170), using `try ... catch` isn't preferred in Scala. A more idiomatic way is to use `Try()` and `.getOrElse`. Maybe something like this?:
```scala
  override def apply(lar: LoanApplicationRegister): Result = {
    when(lar.lienStatus is equalTo(1)) {
      lar.rateSpread is equalTo("NA") or {
        rateSpreadInRange(lar.rateSpread)
          .getOrElse(Failure(s"can't parse ${lar.rateSpread} as decimal"))
      }
    }
  }

  private def rateSpreadInRange(input: String): Try[Result] = {
    Try(BigDecimal(input) is between(BigDecimal("1.50"), BigDecimal("99.99")))
  }
```

This is still new to me, though, so don't take this as law. :)

validation/src/main/scala/hmda/validation/rules/lar/validity/V575.scala
```diff
@@ -1,21 +1,17 @@
...
 import hmda.validation.rules.EditCheck
 
-object V575 extends EditCheck[LoanApplicationRegister] with RegexDsl {
+object V575 extends EditCheck[LoanApplicationRegister] {
   override def name: String = "V575"
 
   override def apply(lar: LoanApplicationRegister): Result = {
     when(lar.lienStatus is equalTo(2)) {
       lar.rateSpread is equalTo("NA") or {
         try {
```
schbetsy: Same comment here as the `try ... catch` block in V570.

validation/src/main/scala/hmda/validation/rules/lar/validity/V570.scala
lgdean: Yeah, I meant to raise (ha) that question in the PR, but forgot to do so. Or actually a related question: why do we even have these failure messages? Are they going to be part of the API (that is, the behavior of the actual program we're writing) in any way?

For now I will assume it's worth keeping them, and yeah, I'll clean up the code.  (And honestly, the fact that you're making the same comment twice, makes it pretty clear that I also need to remove more duplication.  I was going to postpone, but if I'm cleaning up anyway, might as well go ahead.)

validation/src/test/scala/hmda/validation/rules/lar/validity/V570Spec.scala
lgdean: oops, yes. For V575, that case is covered in among the other badly-formatted ones (because anything other than NN.NN is considered badly-formatted by other rules), but I didn't do that here. Will do.

validation/src/main/scala/hmda/validation/rules/lar/validity/V570.scala
schbetsy: Good question. If I remember correctly, we eventually need to be sending the official HMDA-sanctioned error message for each edit, and we aren't currently set up to do that. So, my understanding is that the current error messages will be removed via refactor in the future (but not sure when).

lgdean: OK, I did a thing. It doesn't feel entirely beautiful, but it is in one place, and I think even the right place.

schbetsy:
LGTM. Nice refactor 👍 .

Merging.

PR #306, opened 2016-05-02
V485

### LINKED ISSUE #149: Validation (2017): V485 (Co-Applicant Race)
**Edit #:** V485
**Type:** Validity
**Column:** Co-Applicant Race

If co-applicant race = 1-5 in co-applicant race field 1, then all other co-applicant race fields must = blank or 1-5.

Additional resources:

schbetsy:
Closes #149.

This edit is very similar to #145 , but for co-applicant instead of applicant. Compare to #298. I've used a different strategy for coding this edit, so I'm looking for feedback about which way is preferable (if any). If one is significantly better, I can refactor so that both edits use the same strategy.

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
@@ -1,0 +1,30 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+import scala.util.Try
+
+object V485 extends EditCheck[LoanApplicationRegister] {
+
+  def apply(lar: LoanApplicationRegister): Result = {
+    val applicant = lar.applicant
+    when(applicant.coRace1 is containedIn(1 to 5)) {
+      (validRace(applicant.coRace2) is equalTo(true)) and
+        (validRace(applicant.coRace3) is equalTo(true)) and
+        (validRace(applicant.coRace4) is equalTo(true)) and
+        (validRace(applicant.coRace5) is equalTo(true))
+    }
+  }
+
+  private def validRace(input: String): Boolean = {
+    (input == "") || inRange(1 to 5, input).getOrElse(false)
+  }
+
+  private def inRange(domain: Seq[Int], input: String): Try[Boolean] = {
+    Try(domain.contains(input.toInt))
+  }
+
+  def name: String = "V485"
+
+}
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
@@ -1,7 +1,7 @@
 package hmda.validation.rules.lar.validity
 
 import hmda.model.fi.lar.LoanApplicationRegister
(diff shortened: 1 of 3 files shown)


validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
```diff
@@ -0,0 +1,31 @@
...
+    val applicant = lar.applicant
+    when(applicant.coRace1 is containedIn(1 to 5)) {
+      (applicant.coRace2 is validRace) and
+        (applicant.coRace3 is validRace) and
+        (applicant.coRace4 is validRace) and
+        (applicant.coRace5 is validRace)
+    }
+  }
+
+  def validRace: Predicate[String] = new Predicate[String] {
```
jmarin: I see how this makes the language for this edit a little nicer, but I don't think it belongs here. It leaks implementation of the language into its usage. This also breaks the typeclass pattern, which we are using for the DSL (there is a refactor that I am working on about this topic that will explain this in much further detail). If you want to keep it this way, I suggest putting this function in [HmdaDsl](https://github.com/cfpb/hmda-platform/blob/master/validation/src/main/scala/hmda/validation/dsl/HmdaDsl.scala) since this language is pretty specific to HMDA (and possibly including `blankOr1to5` as an inner function to `validRace`)

validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
```diff
@@ -0,0 +1,31 @@
...
+object V485 extends EditCheck[LoanApplicationRegister] {
+
+  def apply(lar: LoanApplicationRegister): Result = {
+    val applicant = lar.applicant
+    when(applicant.coRace1 is containedIn(1 to 5)) {
+      (applicant.coRace2 is validRace) and
+        (applicant.coRace3 is validRace) and
+        (applicant.coRace4 is validRace) and
+        (applicant.coRace5 is validRace)
+    }
```
lgdean: If you'd like, you can do something like `List(applicant.coRace2, applicant.coRace3, applicant.coRace4, applicant.coRace5).map(_ is validRace).reduce(_ and _)` to remove some of the duplication.  Does it lose readability? I can't tell at the moment.

Likewise in V470: `List(applicant.race2, applicant.race3, applicant.race4, applicant.race5).forall(validRace(_)) is equalTo(true)`

I don't know whether @jmarin's changes will make one version make more sense than the other.

jmarin: I think the one liner loses readability, especially for less experienced Scala developers.

schbetsy:
Thanks for your detailed comments, @jmarin and @lgdean. 

I decided to revert the commit that adds a new predicate for `validRace`. I didn't want to add a predicate to the DSL that was only useful in this very specific context.

validation/src/main/scala/hmda/validation/rules/lar/validity/V485.scala
schbetsy: I think `List(applicant.race2, applicant.race3, applicant.race4, applicant.race5).forall(validRace(_)) is equalTo(true)` is decently readable, FWIW.

But I think it doesn't necessarily need to be rewritten, so I'm leaving it as is (after reverting the refactor)

PR #307, opened 2016-05-02
V450 & V460

schbetsy:
Closes #138 
Closes #139

PR #309, opened 2016-05-03
refactor: extract method to generate int outside range

lgdean:
Extract duplicated logic into a single handy method.
Rename a Utils class to something more general, and put method there.
If it grows too large in the future, we can re-evaluate.

This change was inspired by PR #307 (issues #138 and #139) but has a scope beyond those rules.


schbetsy:
Great idea. LGTM.

PR #310, opened 2016-05-03
Typeclass refactor

jmarin:
This PR deals with refactoring the `DSL` to better conform to the `Typeclass` pattern in `Scala`. Here is an introduction to this pattern, why we are using it and how it is used in `Scala` (all of which motivate in some form this refactor). 

Type classes are present in the standard `Scala` library and are the foundation of many third party libraries in the `Scala` ecosystem. One of the main purposes of using type classes is to provide extensibility to existing code without the need to write adapter objects. Type classes allow ad-hoc and [retroactive polymorphism] (https://en.wikipedia.org/wiki/Ad_hoc_polymorphism). As such, this differs from the classic object oriented technique of extending a class or implementing an interface (type classes have nothing to do with OO classes; in `Scala` some of these concepts can be confusing because we have to use classes, objects and traits to implement this pattern). 

Type classes don’t have native support in `Scala`, but this pattern can be created by using some features already present in the language, as follows (example here is providing `Json` serialization to arbitrary types; code inspired by the `spray-json` library)

1) Define a trait. This is the actual type class, and in general uses parameterized types to make it as flexible as possible. It represents the functionality that we want to implement. In this case we want a generic way to serialize objects to `Json`.

```scala
trait JsonWriter[T] {
  def write(value: T): JsValue
}
```

2) Type class instances. Provide particular implementations of the type class instance. This uses __implicits__ so that when these instances are brought into scope the compiler can inject this implementation. They are usually implemented as an object wrapping the implicit instances for our type class. To illustrate this, here we provide two instances to serialize `String` and an arbitrary and fictional domain object, `Customer`.

```scala
object DefaultJsonWriter {
  implicit val stringJsonWriter = new JsonWriter[String] {
    def write(value: String): Json = ???
  }
  implicit val customerJsonWriter = new JsonWriter[Customer] {
    def write(value: Customer): Json = ???
  }
}
```

In many `Scala` libraries it is common to use this naming convention of “Default”, “Defaults” or “Instances” in addition to the type class name. 

3) Interfaces. This is what is exposed to users, they are generic methods that take an instance of a type class as an implicit parameter, and they represent the operations that can be done with our type classes. While this can be done in different ways, many libraries wrap this into a “Syntax” or “Protocol” object that includes an “Ops” implicit class that is brought into scope when used. 

```scala
object JsonSyntax {
  implicit class JsonWriterOps[T](value: T) {
    def toJson(implicit writer: JsonWriter[T]): Json = {
      writer.write(value)
    }
  }
}
```

To use this pattern, we need to import the syntax and the instances that we want to use:

```scala
import JsonSyntax._
import DefaultJsonWriter._

val json: JsValue = Customer(“customer”, “email@email.com”).toJson
```

Note that as a user of this functionality the developer doesn’t have to deal with implicits directly, but they are there (so while it’s useful to understand how this happens, it’s not required to use the functionality created). 

The same `toJson` function can be applied to any type class instance for which there is an implicit `JsonWriter` implementation in scope. 

In the `HMDA` code base, and following these conventions, this PR intends to better follow this pattern by performing the following changes:

1) renaming the `DSL` related objects as follows (type class trait `Predicate` remains as is). This reflects better that they are a “unit”

`CommonDsl` —> `PredicateCommon`
`HmdaDsl` —> `PredicateHmda`
`RegexDsl` —> `PredicateRegex`

2) Extracting the implicit class into `PredicateSyntax` and renaming it to `PredicateOps` (“operations with predicates”) as per the most utilized naming conventions out there. 

3) Adapting every validation rule to use this pattern by importing the relevant artifacts (in most cases, `PredicateDefaults` and `PredicateSyntax`)


validation/src/main/scala/hmda/validation/rules/lar/syntactical/S010.scala
```diff
@@ -1,11 +1,14 @@
...
 import hmda.model.fi.lar.LoanApplicationRegister
-import hmda.validation.dsl.{ CommonDsl, Result }
+import hmda.validation.dsl.Result
 import hmda.validation.rules.EditCheck
 
 object S010 extends EditCheck[LoanApplicationRegister] {
 
+  import hmda.validation.dsl.PredicateSyntax._
+  import hmda.validation.dsl.PredicateDefaults._
+
```
lgdean: What's the reasoning behind having these imports here, rather than up with the others?  I know they are technically different scopes, but for practical purposes they seem the same, since all the code in the file is in the one object.  Having two chunks of imports feels like clutter in this case.

But I can imagine that there may be a good reason to do it.  If so, would that reason also apply to importing `Result`?

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S013.scala
```diff
@@ -1,11 +1,14 @@
 package hmda.validation.rules.ts.syntactical
 
 import hmda.model.fi.ts.TransmittalSheet
-import hmda.validation.dsl.{ CommonDsl, Result }
-
+import hmda.validation.dsl.Result
 import scala.concurrent.{ ExecutionContext, Future }
 
-object S013 extends CommonDsl {
+object S013 {
```
hkeeler: Why does this not extend `EditCheck[TransmittalSheet]`?

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S100.scala
```diff
@@ -1,14 +1,17 @@
...
-import hmda.validation.dsl.{ CommonDsl, Result }
+import hmda.validation.dsl.Result
 
 import scala.concurrent.{ ExecutionContext, Future }
 
 /*
  Activity year must = year being processed (i.e. = 2016)
  */
-object S100 extends CommonDsl {
+object S100 {
```
hkeeler: Same, _why does this not extend `EditCheck[TransmittalSheet]`?_ question.

hkeeler:
Not specific to this PR, just something that caught my eye when scanning through all the edits.  We're not consistent in our use of `override` in those classes.  Some `override` on `apply`, some on `name`, some neither, some both.

hkeeler:
I'm all reviewed up.  Overall, looks pretty good, just a couple questions/comments.

validation/src/main/scala/hmda/validation/dsl/PredicateDefaults.scala
```diff
@@ -1,92 +1,68 @@
...
-    }
-
-    def not(predicate: Predicate[T]): Result = {
-      testNot(predicate)
-    }
-
-  }
-
-  def equalTo[T](that: T): Predicate[T] = new Predicate[T] {
+object PredicateDefaults {
```
lgdean: I don't understand this name. In what sense are these defaults? I understand the `Default` in the example from the explanation, because I could imagine an application needing to use a different `JsonWriter` instead, but here it just seems like a group of common predicates.

Are you intending to imply that these default behaviors might be overridden in some contexts?  That's how I read that word.  So if so, then we're good.  But if not, then I (currently) find it confusing.

(Note: I don't mean "overridden" in the OO-method-implementation-inheritance sense, but simply in the sense of providing an alternative implementation in some situations.)

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S013.scala
jmarin: The `apply` method for this edit has a different signature than the parent class. It is not clear what is going to happen with this edit, since it doesn't make much sense given how we are building the submission process.

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S100.scala
jmarin: Same reason as above, the `apply` method has a different signature, taking two parameters

validation/src/main/scala/hmda/validation/dsl/PredicateDefaults.scala
jmarin: Other libraries use this notation, so that's why I chose it. Would `PredicateInstances` or `PredicateCommon` work for you? I don't have a preference

lgdean: Yes, I'd like `PredicateCommon` or `CommonPredicates`; `PredicateInstances` is also fine though a little vague.  (I suppose "common" is also vague, but it expresses an intent, even if not super-precise, so I prefer it.)

jmarin:
@lgdean @hkeeler I've addressed all the comments in your review. Can either one of you take a last look and merge?

lgdean:
LGTM. Good to merge once @hkeeler gets a chance to look.

hkeeler:
I know its already merged, but yep, looks good.

PR #313, opened 2016-05-09
V500 and some test code work

### LINKED ISSUE #165: Validation (2017): V500 (Rate Spread)
**Edit #:** V500
**Type:** Validity
**Column:** Rate Spread

Rate spread must be in the NN.NN format or NA, must not = blank.

Additional resources:

lgdean:
Closes #165 

Also includes some changes to test code, with the goal of removing some boilerplate without reducing readability.  Right now it just introduces methods to say that a given LAR passes or fails with the relevant edit; there may be more we can do in the future, but I wanted to go ahead and get the PR in, since this week has other stuff going on.

Also, there may be nicer ways to write it with Matchers, but I was away from the network when writing this code, so couldn't look into best practices for that.  (I'd definitely like to investigate it in the future.)

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/RegexDsl.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/RegexDsl.scala
@@ -14,10 +14,18 @@ trait RegexDsl {
     override def failure: String = s"is not a valid email"
   }
 
-  def numericMatching: Predicate[String] = new Predicate[String] {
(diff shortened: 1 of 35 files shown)


schbetsy:
Very smart refactor of the RegexDsl. Looks great.

Implementation of V500 also LGTM.

The refactor of `LarEditCheckSpec` I think is also positive. Let's make sure to point out `mustPass` and `mustFail` to the rest of team so they remember to use them for future edit check specs.

Issue #155, opened 2016-03-08
Validation (2017): V330 (Income)


hkeeler:
**Edit #:** V330
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Income

Income must be numeric and > 0, or equal NA.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

PR #316, opened 2016-05-16
V330: Income Validation

nickgrippin:
Implements V330 check.

Let me know if I violated any style guidelines, or missed an incredibly easy way to make this code more streamlined.


nickgrippin:
Closes #155

PR #317, opened 2016-05-17
V335: Income Validation

### LINKED ISSUE #156: Validation (2017): V335 (Income)
**Edit #:** V335
**Type:** Validity
**Column:** Income

If property type = 3, then income must = NA.

Additional resources:

nickgrippin:
Closes #156 

Implements V335 income validation check.

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/validity/V335Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/validity/V335Spec.scala
@@ -1,7 +1,8 @@
 package hmda.validation.rules.lar.validity
 
-import hmda.model.fi.lar.{ Loan, LoanApplicationRegister }
+import hmda.model.fi.lar.LoanApplicationRegister
 import hmda.validation.dsl.{ Failure, Success }
+import hmda.validation.rules.EditCheck
 import hmda.validation.rules.lar.LarEditCheckSpec
 import org.scalacheck.Gen
 
@@ -33,4 +34,6 @@ class V335Spec extends LarEditCheckSpec {
       V335(invalidLar) mustBe a[Failure]
     }
   }
+
+  override def check: EditCheck[LoanApplicationRegister] = V335
 }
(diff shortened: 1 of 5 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/V335Spec.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import hmda.model.fi.lar.{ Loan, LoanApplicationRegister }
+import hmda.validation.dsl.{ Failure, Success }
+import hmda.validation.rules.lar.LarEditCheckSpec
+import org.scalacheck.Gen
+
+class V335Spec extends LarEditCheckSpec {
+  property("Succeeds when property type is not 3") {
+    forAll(larGen) { lar =>
+      whenever(lar.loan.propertyType != 3) {
```
nickgrippin: Quick question:  why is this line necessary in the test?  The validation should pass for all LARs, not just ones when the property type is not equal to 3.  I've seen this pattern followed for other validations (410, as one example), and it made my tests pass, but I'm not sure if I understand why.

lgdean: The validation should not pass for all LARs.  If the property type is 3, and the income is numeric (or, for that matter, "nonsense_string" or the like), then it should fail.
(But if the property type is not 3, then it should pass no matter what.  And those are the cases covered by this test.)

This answer may not have addressed your question.  If not, feel free to ping me in chat (or here, if you prefer).  Or if you haven't already (which you may have), try removing the line and see whether/how/why it fails.  And definitely feel free to ask if you get stuck on the "why" part, as that may be part of what you're asking in the first place, and my goal is not to make you go around in circles. :)

nickgrippin: Resolved, but adding a comment here for future readers.

The LARs generated in `LarGenerators.scala` are not supposed to pass all the validations, only the formatting checks.  So, a generated LAR might have the property type 3 and a numeric income (which is checked in a test later in this file).  In this test, we want to make sure the conditional statement works at filtering out all cases where `property type != 3`

nickgrippin:
Closes #156

lgdean:
LGTM, but I believe it's now in a state where it won't work due to other changes merged to master. Happy to merge once that is resolved. Thanks, and sorry for the post-conference churn.

nickgrippin:
Got a little too messy, closing this one and opening a fresh PR

PR #321, opened 2016-05-17
Add check method to V330Spec

jmarin:


lgdean:
LGTM, and I'm also fine with you merging your own PR if you happen to notice Travis passing before I do.

PR #323, opened 2016-05-17
V338: Income

### LINKED ISSUE #157: Validation (2017): V338 (Income)
**Edit #:** V338
**Type:** Validity
**Column:** Income

If applicant ethnicity, race, and sex = 4, 7, and 4, respectively, and co- applicant ethnicity, race, and sex = 5, 8, and 5, respectively, and  action taken = 1-5 or 7-8, then applicant income must = NA.

Additional resources:

nickgrippin:
addresses #157

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
@@ -3,11 +3,10 @@ package hmda.validation.rules.lar.validity
 import hmda.model.fi.lar.LoanApplicationRegister
 import hmda.validation.dsl.Result
 import hmda.validation.rules.EditCheck
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax._
 
 object V338 extends EditCheck[LoanApplicationRegister] {
-  import hmda.validation.dsl.PredicateCommon._
-  import hmda.validation.dsl.PredicateSyntax._
-
   override def name: String = "V338"
 
   override def apply(lar: LoanApplicationRegister): Result = {
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
@@ -7,7 +7,7 @@ import hmda.validation.rules.EditCheck
 object V338 extends EditCheck[LoanApplicationRegister] {
   override def name: String = "V338"
 
-  def apply(lar: LoanApplicationRegister): Result = {
+  override def apply(lar: LoanApplicationRegister): Result = {
     when(applicantOk(lar)) {
       lar.applicant.income is equalTo("NA")
     }
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
@@ -1,0 +1,27 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V338 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V338"
+
+  def apply(lar: LoanApplicationRegister): Result = {
+    when(applicantOk(lar)) {
+      lar.applicant.income is equalTo("NA")
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/V338Spec.scala
```diff
@@ -0,0 +1,51 @@
...
+      val applicant = lar.applicant.copy(ethnicity = 4, race1 = 7, sex = 4, coEthnicity = 5,
+        coRace1 = 8, coSex = 5, income = income.toString)
+      val invalidLar = lar.copy(applicant = applicant, actionTakenType = x)
+      invalidLar.mustFail
+    }
+  }
+
+  override def check: EditCheck[LoanApplicationRegister] = V338
+
+  private def applicantNotOk(lar: LoanApplicationRegister): Boolean = {
```
lgdean: I don't buy this approach.  I changed an `and` to an `or` in the code under test, and then ran the test 5 times.  It failed 2 out of 5 times, which is better than 0, but not as good as 5.  I'm not sure what to suggest instead.  The obvious suggestion would be 7 different tests, but that seems like a lot of code.  When I think about it, it leads quickly to deeper questions of what we're trying to do and what we're trying to test.  (It's a rule.  Does the test need to be an [ugly mirror](http://jasonrudolph.com/blog/2008/07/30/testing-anti-patterns-the-ugly-mirror/) of the rule?  Would actual examples be better, instead of generated values?  That sort of question.)

validation/src/test/scala/hmda/validation/rules/lar/validity/V338Spec.scala
```diff
@@ -0,0 +1,51 @@
...
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.rules.EditCheck
+import hmda.validation.rules.lar.LarEditCheckSpec
+import org.scalacheck.Gen
+
+class V338Spec extends LarEditCheckSpec {
+  property("All applicants not meeting the criteria must pass") {
+    forAll(larGen) { lar =>
+      whenever(applicantNotOk(lar)) {
+        lar.mustPass
```
lgdean: The "Ok" naming is a bit confusing.  It's confusing in both the production code and the test code, but I think these two lines are the best example of what confused me.  When the applicant is not OK, the LAR must pass the validation?

Unfortunately I do not yet have a better name to suggest.  Something that means "the kind of LAR to which this rule even applies," but ideally with fewer words.  I wonder whether that concept already appears elsewhere in the code, and if so, what we called it there.  @schbetsy, do you know of any examples?

validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
```diff
@@ -0,0 +1,27 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V338 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V338"
+
+  def apply(lar: LoanApplicationRegister): Result = {
```
jmarin: This needs an `override` modifier. @hkeeler pointed this out, it's very inconsistent throughout the codebase. Both `name` and `apply` override from base abstract class, and should have this notation. There is a fix for all previous inconsistencies in another PR

validation/src/main/scala/hmda/validation/rules/lar/validity/V338.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def apply(lar: LoanApplicationRegister): Result = {
+    when(notANaturalPerson(lar)) {
+      lar.applicant.income is equalTo("NA")
+    }
+  }
+
+  private def notANaturalPerson(lar: LoanApplicationRegister): Result = {
+    val okAction = List(1, 2, 3, 4, 5, 7, 8)
+
+    (lar.applicant.ethnicity is equalTo(4)) and
```
nickgrippin: Should I put these values into a List and compare it with `List(4,7,4,5,8,5)`?  I wasn't sure if that would be a better approach.

lgdean: I've read this question a few times and not come up with a solid answer. So let's keep it as is.

lgdean:
This looks good. I have an idea for reducing duplication in the test code, but that's no reason to keep this PR open. I'll merge it and then assign #157 to myself. If my idea works out, I'll open another PR; if not, I'll close the issue.

PR #325, opened 2016-05-17
V230

### LINKED ISSUE #119: Validation (2017): V230 (Occupancy)
**Edit #:** V230
**Type:** Validity
**Column:** Occupancy

Occupancy must = 1, 2, or 3.

Additional resources:

kgudel:
closes #119


validation/src/main/scala/hmda/validation/rules/lar/validity/V230.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object V230 extends EditCheck[LoanApplicationRegister] {
+
+  val occupancyStatusTypes = List(1, 2, 3)
+
+  def apply(lar: LoanApplicationRegister): Result = {
+    lar.loan.occupancy is containedIn(occupancyStatusTypes)
+  }
+
+  def name = "V230"
```
jmarin: Both `apply` and `name` need an `override` modifier, since they extend from `EditCheck` abstract base class that has those methods

schbetsy: My understanding was that the `override` modifier is optional when extending an abstract class. Is it really necessary, or is it just stylistically preferable? (Just curious)

jmarin: Our codebase was inconsistent about this one, with some `apply` and `name` methods having `override` while others didn't. It's more explicit and shows clear intent if we use the override modifier. You are right that technically it's optional if the parent abstract class has no concrete implementation.

validation/src/main/scala/hmda/validation/rules/lar/validity/V230.scala
```diff
@@ -0,0 +1,16 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.{ LoanApplicationRegister }
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V230 extends EditCheck[LoanApplicationRegister] {
+
+  val occupancyStatusTypes = List(1, 2, 3)
```
schbetsy: Small comment: I don't think "Types" is an accurate way to name this list. Maybe `occupancyStatuses` or `occupancyValues` instead? Or you could use the list directly inline on line 12 so you don't have to name it.

validation/src/test/scala/hmda/validation/rules/lar/validity/V230Spec.scala
```diff
@@ -0,0 +1,26 @@
...
+  property("Occupancy other than 1, 2, or 3 is invalid") {
+    forAll(larGen, badLienStatusGen) { (lar: LoanApplicationRegister, x: Int) =>
+      val invalidLoan = lar.loan.copy(occupancy = x)
+      val invalidLar = lar.copy(loan = invalidLoan)
+      invalidLar.mustFail
+    }
+  }
+
+  override def check: EditCheck[LoanApplicationRegister] = V230
+}
```
schbetsy: This 🚫 ↩️  symbol happens when your text editor doesn't add a newline character at the end of the file. I'm not sure why it added one correctly in `V230.scala`, but not in `V230Spec.scala`. Any idea?

kgudel: I do not.

schbetsy:
LGTM

PR #329, opened 2016-05-18
V545

### LINKED ISSUE #173: Validation (2017): V545 (HOEPA Status)
**Edit #:** V545
**Type:** Validity
**Column:** HOEPA Status

If lien status = 3, then HOEPA status must = 2.

Additional resources:

kgudel:
closes #173

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/validity/V545Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/validity/V545Spec.scala
@@ -1,0 +1,36 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.rules.EditCheck
+import hmda.validation.rules.lar.LarEditCheckSpec
+import org.scalacheck.Gen
+
+class V545Spec extends LarEditCheckSpec with BadValueUtils {
+
+  property("Valid when HOEPA status is 2") {
+    forAll(larGen) { lar =>
+      whenever(lar.hoepaStatus == 2) {
+        lar.mustPass
+      }
+    }
+  }
+
+  property("Valid when Lien status != 3") {
+    forAll(larGen) { lar =>
+      whenever(lar.lienStatus != 3) {
+        lar.mustPass
+      }
+    }
+  }
+
+  val invalidHoepaGen: Gen[Int] = intOutsideRange(2, 2)
+
+  property("Invalid when lien status is 3 and HOEPA status not 2") {
+    forAll(larGen, invalidHoepaGen) { (lar: LoanApplicationRegister, x: Int) =>
+      val invalidLar = lar.copy(lienStatus = 3, hoepaStatus = x)
+      invalidLar.mustFail
+    }
+  }
+
+  override def check: EditCheck[LoanApplicationRegister] = V545
+}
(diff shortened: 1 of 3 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/V545Spec.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  property("Valid when Lien status != 3") {
+    forAll(larGen) { lar =>
+      whenever(lar.lienStatus != 3) {
+        lar.mustPass
+      }
+    }
+  }
+
+  val invalidHoepaGen: Gen[Int] = intOutsideRange(2, 2)
```
nickgrippin: I feel like `intOutsideRange()` doesn't fit here.  Try something like this instead:
```
Gen.choose(Int.MinValue, Int.MaxValue) filter(_ != 2)
```

validation/src/test/scala/hmda/validation/rules/lar/validity/V545Spec.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  property("Valid when Lien status != 3") {
+    forAll(larGen) { lar =>
+      whenever(lar.lienStatus != 3) {
+        lar.mustPass
+      }
+    }
+  }
+
+  val invalidHoepaGen: Gen[Int] = Gen.choose(Int.MinValue, Int.MaxValue).filter(_ != 2)
```
schbetsy: [The API documentation for `Gen.filter`](https://www.scalacheck.org/files/scalacheck_2.11-1.13.1-api/index.html#org.scalacheck.Gen@filter(p:T=>Boolean):org.scalacheck.Gen[T]) says "If the condition is not fulfilled, the generator fails (returns None)."  So it's possible for the `invalidHoepaGen` to fail, which will break the test (I think). 

Might be better to use `intOutsideRange(2,2)` instead.

schbetsy: Caveat: I haven't tested the suggestion I made above, and I'm not sure exactly what happens when the generator returns None. If you find out, or if you want to investigate together, let me know.

I didn't notice that when I was reviewing https://github.com/cfpb/hmda-platform/pull/327, unfortunately. Once we figure out a good solution here, would you update that one as well?

nickgrippin: I don't think that it will return None unless the conditions are very specific or impossible (see [this documentation](https://github.com/rickynils/scalacheck/blob/master/doc/UserGuide.md)).  The docs aren't completely clear, but that seems to be the case.
If we still think this is a danger, we can always just use `retryUntil(some condition)`, which will try the generator again until it passes the condition

nickgrippin: Oops, correction:  It will return `Option[None]` if it happens to pick 2.  

I still think that `retryUntil()` would be a more efficient option in this case (since there's such a small chance of it equalling 2).

schbetsy: @kgudel Ignore everything we just said. Your version works well. `forAll` only runs the test when the generator succeeds. So inside the test, you never have to worry about a `None`.

schbetsy:
LGTM

PR #336, opened 2016-05-18
V555

### LINKED ISSUE #175: Validation (2017): V555 (Lien Status)
**Edit #:** V555
**Type:** Validity
**Column:** Lien Status

If loan purpose = 1 or 3, then lien status must = 1, 2, or 4.

Additional resources:

kgudel:
Closes #175

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V555.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V555.scala
@@ -1,0 +1,19 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V555 extends EditCheck[LoanApplicationRegister] {
+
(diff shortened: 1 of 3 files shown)


validation/src/main/scala/hmda/validation/rules/lar/validity/V555.scala
```diff
@@ -0,0 +1,21 @@
...
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax._
+
+object V555 extends EditCheck[LoanApplicationRegister] {
+
+  val loanPurposeTypes = List(1, 3)
+  val lienStatusTypes = List(1, 2, 4)
```
jmarin: I think I've seen this commented elsewhere as well, but the notation "Types" is potentially confusing, since we always talk about types in `Scala` to refer to the type system. I would read this variable and expect a type alias or something similar. Maybe change to `loanPurposeValues` or similar (look around in the codebase, not sure if we are already using some convention)

PR #338, opened 2016-05-19
V360: Denial reasons not equal

### LINKED ISSUE #163: Validation (2017): V360 (Reasons for Denial)
**Edit #:** V360
**Type:** Validity
**Column:** Reasons for Denial

Responses for reasons for denial must not be the same (i.e. 1,1; 1,1,2; 1,2,2;...........).

Additional resources:

nickgrippin:
Closes #163

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
@@ -54,7 +54,7 @@ object PredicateCommon {
   def numeric[T]: Predicate[T] = new Predicate[T] {
     override def validate: (T) => Boolean = _.asInstanceOf[AnyRef] match {
       case n: Number => true
-      case _ => throw new NotImplementedError("'numeric' doesn't handle non-number/string values yet")
+      case _ => throw new NotImplementedError("'numeric' doesn't handle string (or other) values yet")
     }
     override def failure: String = s"is not numeric"
   }
--- a/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
@@ -1,71 +1,47 @@
 package hmda.validation.dsl
 
+import scala.language.implicitConversions
 import scala.util.Try
 
-trait CommonDsl {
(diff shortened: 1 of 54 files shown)


validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
```diff
@@ -54,7 +54,8 @@ object PredicateCommon {
   def numeric[T]: Predicate[T] = new Predicate[T] {
     override def validate: (T) => Boolean = _.asInstanceOf[AnyRef] match {
       case n: Number => true
-      case _ => throw new NotImplementedError("'numeric' doesn't handle string (or other) values yet")
+      case s: String => (s forall Character.isDigit) && !s.isEmpty
```
schbetsy: Looks like this case (`case s: String => ...`) is no longer necessary

nickgrippin: True, but I feel like this code could be helpful in the future.  Would it be better to make this change as a separate, one-line pull request (which might satisfy codecov)?

schbetsy: Sure. Then we can discuss it without blocking this PR.

validation/src/test/scala/hmda/validation/rules/lar/validity/V360Spec.scala
```diff
@@ -0,0 +1,37 @@
...
+import hmda.model.fi.lar.{ Denial, LoanApplicationRegister }
+import hmda.validation.dsl.{ Failure, Success }
+import hmda.validation.rules.EditCheck
+import hmda.validation.rules.lar.LarEditCheckSpec
+import org.scalacheck.Gen
+
+class V360Spec extends LarEditCheckSpec {
+  property("Denials with all different values must pass") {
+    forAll(larGen) { lar =>
+      val goodDenial = lar.denial.copy(reason1 = "1", reason2 = "2", reason3 = "")
```
schbetsy: I'd like to see this test be a bit more generic. For instance, you could use a generator like `badOptionGen` below to generate three values, then just use a `whenever` clause to make sure they're all different.

schbetsy:
LGTM

PR #341, opened 2016-05-19
V480, V495 + spec refactor

### LINKED ISSUE #147: Validation (2017): V480 (Applicant Race)
**Edit #:** V480
**Type:** Validity
**Column:** Applicant Race

Applicant race must not be the same (i.e. 1,1; 1,1,2; 1,2,3,4,1), when more than one applicant race is designated.

Additional resources:

### LINKED ISSUE #151: Validation (2017): V495 (Co-Applicant Race)
**Edit #:** V495
**Type:** Validity
**Column:** Co-Applicant Race

Co-Applicant race must not be the same (i.e. 1,1; 1,1,2; 1,2,3,4,1), when more than one co-applicant race is designated.

Additional resources:

schbetsy:
Closes #147 
Closes #151

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V495.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V495.scala
@@ -1,0 +1,17 @@
+package hmda.validation.rules.lar.validity
+
+import hmda.model.fi.lar.{ Applicant, LoanApplicationRegister }
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+
+object V495 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V495"
+
+  override def apply(lar: LoanApplicationRegister): Result = {
+    val appl: Applicant = lar.applicant
+    val coRaceFields: List[String] = List(appl.coRace1.toString, appl.coRace2, appl.coRace3, appl.coRace4, appl.coRace5)
+    val coRaces = coRaceFields.filterNot(_.isEmpty)
+
+    coRaces.distinct.size is equalTo(coRaces.size)
+  }
+}
--- a/validation/src/main/scala/hmda/validation/rules/lar/validity/V495.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/validity/V495.scala
@@ -3,6 +3,8 @@ package hmda.validation.rules.lar.validity
 import hmda.model.fi.lar.{ Applicant, LoanApplicationRegister }
 import hmda.validation.dsl.Result
 import hmda.validation.rules.EditCheck
+import hmda.validation.dsl.PredicateCommon._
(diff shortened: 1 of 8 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/RaceEditCheckSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  def failsWithRace(vals: List[String]) = {
+    forAll(larGen) { lar: LoanApplicationRegister =>
+      val testApplicant = applicantWithRaceVals(lar.applicant, vals)
+      val testLar = lar.copy(applicant = testApplicant)
+      testLar.mustFail
+    }
+  }
+  def failsWithCoRace(vals: List[String]) = {
```
lgdean: I very much like having these methods. Can you say more about why they take a List rather than 5 args? I can see that it makes it easier to pass the chunk of vals down into the next method, and maybe 5 different names wouldn't be meaningful, but I also notice that it does mean you're writing `List` each time, and that it means the compiler can't yell if you pass the wrong number of values. I'm not sure how to weigh these things, and I suspect you have reasons that I haven't considered (and have spent more time thinking about those I have), so I'd like to hear what you say.

validation/src/main/scala/hmda/validation/rules/lar/validity/V495.scala
```diff
@@ -0,0 +1,19 @@
...
+object V495 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V495"
+
+  override def apply(lar: LoanApplicationRegister): Result = {
+    val appl: Applicant = lar.applicant
+    val coRaceFields: List[String] = List(appl.coRace1.toString, appl.coRace2, appl.coRace3, appl.coRace4, appl.coRace5)
+    val coRaces = coRaceFields.filterNot(_.isEmpty)
+
+    coRaces.distinct.size is equalTo(coRaces.size)
+  }
```
lgdean: V480 and V495 are extremely similar. It feels like a `Races` object is waiting to be born.

validation/src/test/scala/hmda/validation/rules/lar/validity/RaceEditCheckSpec.scala
schbetsy: I had pretty much the same assessment of the pros and cons.

- Pros of the list: I knew I would be passing the values to a helper method (`applicantWithRaceVals`), and passing five named variables straight through seemed ugly and a hassle.
- Pros of named variables: enforce correct number of values and correct types (race1 = `Int`, others = `String`).

Another issue with the current implementation is that it only allows hard-coded values. Maybe switching to named variables would make it easier to use a generator to fill in the desired field with randomized values...

kgudel:
LGTM

nickgrippin:
LGTM as well

Issue #113, opened 2016-03-08
Validation (2017): V210 (Date Application Received)


hkeeler:
**Edit #:** V210
**Type:** Validity
**Category:** Loan/Application Register (only)
**Column:** Date Application Received

Date application received must be in CCYYMMDD format; month and day must be in the range 01-12 and 01-31, respectively or equal NA. CCYY must be numeric and CC (century) must equal 20.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

PR #343, opened 2016-05-20
String case for numeric check

nickgrippin:


validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
```diff
@@ -54,7 +54,8 @@ object PredicateCommon {
   def numeric[T]: Predicate[T] = new Predicate[T] {
     override def validate: (T) => Boolean = _.asInstanceOf[AnyRef] match {
       case n: Number => true
-      case _ => throw new NotImplementedError("'numeric' doesn't handle string (or other) values yet")
+      case s: String => Try(s.toDouble).isSuccess
```
jmarin: Is this currently being used in the codebase?

nickgrippin: No, it was something I wrote for an approach that was later dropped.  I figured it might be useful, but I can just close the ticket if this looks unnecessary

jmarin: Maybe save this branch for later, in case it's needed? I don't think we should introduce code that is not going to be used, it will linger in the codebase without being properly tested.

nickgrippin: Actually, I will (potentially) be using this in #113 to parse the month and day

PR #344, opened 2016-05-20
V120, V135, and tsSpec edit refactor

kgudel:
closes #110 
closes #111 
The refactor to the transmittal sheet editspecs is just to bring it in step with how we do lar editspecs.

PR #346, opened 2016-05-20
V108

### LINKED ISSUE #51: Validation (2017): V108 (Parent Name)
If parent name is reported, then parent name cannot equal institution name

schbetsy:
Closes #51 

Depends on #344

### DIFF
--- a/validation/src/main/scala/hmda/validation/engine/ts/validity/TsValidityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/ts/validity/TsValidityEngine.scala
@@ -3,13 +3,14 @@ package hmda.validation.engine.ts.validity
 import hmda.model.fi.ts.TransmittalSheet
 import hmda.validation.api.ValidationApi
 import hmda.validation.engine.ts.TsCommonEngine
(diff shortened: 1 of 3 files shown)


schbetsy:
I removed the commit that was duplicating #344

api/src/test/scala/hmda/api/processing/HmdaFileUploadSpec.scala
```diff
@@ -2,7 +2,7 @@ package hmda.api.processing
 
 import java.time.Instant
 import akka.testkit.TestProbe
-import hmda.api.processing.HmdaFileUpload.{AddLine, GetState, HmdaFileUploadState}
+import hmda.api.processing.HmdaFileUpload.{ AddLine, GetState, HmdaFileUploadState }
```
lgdean: I made a separate PR (#352) for this, because I didn't notice it here. I'd checked a few other PRs, but didn't happen to look at this one. Is there a better way to do that kind of search?

I am in any case fine with merging this one instead of mine. The whitespace change is a separate commit, so the history will be clear.

lgdean:
LGTM. I'm not merging simply because it's late in my time zone, and I wouldn't be here to watch it go through. (And neither would anyone else.)

PR #315, opened 2016-05-14
Actor refactor

jmarin:
This pull request refactors existing actors to:

* Use factory methods for creating actors in order to facilitate testing
* Replace Akka's `TestKit` with `TestProbe` for testing expected messages

PR #352, opened 2016-05-24
whitespace cleanup

lgdean:
When I run sbt, it changes the whitespace to look like what you see in this commit.  Perhaps the same is not true on other people's machines (and in particular on @jmarin's as of #315), in which case, let's find the discrepancy rather than getting into a commit war.

But if it's true on other people's machines too, then please merge.


lgdean:
merging #346 would be another way to accomplish this goal.

nickgrippin:
Thank you!  I've been undoing these spaces every time I push, just to avoid having the extra file changes.

PR #354, opened 2016-05-24
`Relevant` Keyword

schbetsy:
This is an example of how I would use `relevant` in edit checks. 

Per our discussion today, I think it will be useful to designate conditions that must be met for an edit to run.

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/validity/V540Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/validity/V540Spec.scala
@@ -7,14 +7,14 @@ import org.scalacheck.Gen
 
 class V540Spec extends LarEditCheckSpec with BadValueUtils {
 
-  val relevantActionTakenList = List(2, 3, 4, 5, 7, 8)
+  val relevantActions = List(2, 3, 4, 5, 7, 8)
 
-  val irrelevantActionTakenGen: Gen[Int] =
+  val irrelevantActionGen: Gen[Int] =
     Gen.choose(Int.MinValue, Int.MaxValue)
-      .filter(!relevantActionTakenList.contains(_))
+      .filter(!relevantActions.contains(_))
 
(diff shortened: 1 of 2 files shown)


validation/src/test/scala/hmda/validation/rules/lar/validity/V540Spec.scala
```diff
@@ -7,30 +7,31 @@ import org.scalacheck.Gen
...
-  property("Valid if HOEPA status is 2") {
+  property("Valid when HOEPA status is 2") {
     forAll(larGen) { lar =>
       val validLar = lar.copy(hoepaStatus = 2)
       validLar.mustPass
     }
   }
 
-  val badActionTakenGen: Gen[Int] = Gen.oneOf(badActionTakenList)
+  val actionTakenGen: Gen[Int] = Gen.oneOf(relevantActionTakenList)
```
kgudel: Should this be relevantActionTakenGen?

schbetsy: Sure, that would be an appropriate use of `relevant`. Not sure why I didn't use it here--maybe because I thought it wasn't necessary, and that this Generator's purpose was clear enough.

kgudel:
My understanding from this is that if that we have classified variables into two categories, those in a `when` statement and those not in a/following a `when` statement.
- variables in a `when`:
   - `relevant` when used in a way that causes the when to be true
   - `irrelevant` when used in a way that causes the when to be false
- variables not in a/after a `when`:
   - `valid` when it causes the edit to pass given the when being true
   - `invalid` when it causes the edit to fail given the when being true

We also have variables representing a whole, or larger part of a lar, such as the loan section, these are marked as valid or invalid depending on whether the resulting lar they build are .mustPass or .mustFail respectively.

Is this correct?

schbetsy:
@kgudel That looks correct to me!

PR #358, opened 2016-05-26
Get rid of type erasure compiler warning

jmarin:
`BadValueUtils` has a compiler warning due to type erasure


nickgrippin:
[Good article](http://www.artima.com/scalazine/articles/scalas_type_system.html) that explains why this works (see section on Existential types)

PR #361, opened 2016-05-26
Date parsing fix

nickgrippin:
The `SimpleDateFormat` class will, when parsing, automatically trim any leading zeroes on the day input.  For example, a date input of `201603026` will correctly be formatted by the parser, but should not be able to pass our validation check.  

Shoutout to @kgudel for discovering this bug.


schbetsy:
Oh, weird. So, in your example above, `201603026` will be parsed as 2016-03-26?

nickgrippin:
Yes, that seems to be the case from the documentation that I've read, along with some local testing of my own.  Actually, the formatter `yyyyMMdd` can be simplified to just `yyyyMd`.  I made the decision to keep it in the longer format both for better readability and a more obvious tie-in to the actual validation requirements.

schbetsy:
LGTM

PR #363, opened 2016-05-27
V300

### LINKED ISSUE #137: Validation (2017): V300 (Census Tract)
**Edit #:** V300
**Type:** Validity
**Column:** Census Tract

Census tract must = a valid census tract number for the MSA/MD, state, county combination or (NA if county is classified as small) or (where MSA/MD = NA the census tract must = a valid census tract for the state/county combination or NA). Valid census tract format must be NNNN.NN or NA.

Additional resources:

kgudel:
closes #137


validation/src/main/scala/hmda/validation/dsl/PredicateRegEx.scala
```diff
@@ -4,8 +4,20 @@ import scala.language.implicitConversions
 import scala.util.matching.Regex
 
 object PredicateRegEx {
+  implicit def validCensusTractFormat: Predicate[String] = new Predicate[String] {
+
+    override def validate: (String) => Boolean = {
+      val censusTractRegEx = "^[0-9]{4}.[0-9]{2}$"
```
nickgrippin: `\d` is a shortcut for `[0-9]`.  Actually, with the additional backslash for the escape character (so, `\\d`), it's only two characters shorter, but I think that will better fit the pattern of our regular expressions.

Also, depending on who gets their PR merged in first, we'll need tests for this regex.

kgudel: I'll race you!!! but also make that change.

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
```diff
@@ -0,0 +1,66 @@
...
+    val allCombination = (msa, state, county, tract)
+    val stateCountyCombination = (state, county, tract)
+
+    val validCensusTractCombination = when(msa not equalTo("NA")) {
+      (allCombination is containedIn(validCombination)) or
+        (allCombination is containedIn(validMdCombination))
+    }
+
+    val tractStateCountyCombination = when(msa is equalTo("NA")) {
+      (stateCountyCombination is containedIn(validStateCountyCombination))
```
nickgrippin: I don't think you need the parentheses here (or on line 53)

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
```diff
@@ -0,0 +1,66 @@
...
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax._
+import hmda.validation.dsl.PredicateRegEx._
+
+object V300 extends EditCheck[LoanApplicationRegister] {
+
+  val cbsaTracts = CBSATractLookup.values
+
+  val validCombination = cbsaTracts.map { cbsa =>
+    (cbsa.geoIdMsa, cbsa.state, cbsa.county, cbsa.tractDecimal)
```
schbetsy: I wanted to check my understanding. It looks like the difference between `validCombination` and `validMdCombination` is that the former uses MSA and the latter uses MD. Is that right?

If so, I think `validMsaCombination` would be a clearer name.

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  val validMdCombination = cbsaTracts.map { cbsa =>
+    (cbsa.metDivFp, cbsa.state, cbsa.county, cbsa.tractDecimal)
+  }
+
+  val validStateCountyCombination = cbsaTracts.map { cbsa =>
+    (cbsa.state, cbsa.county, cbsa.tractDecimal)
+  }
```
schbetsy: I'm not familiar enough with the CBSA data file to know whether there will be any duplicates in `validStateCountyCombination`, `validMdCombination`, or `validCombination`. If so, adding `.distinct` to the end would improve performance at runtime.

validation/src/main/scala/hmda/validation/dsl/PredicateRegEx.scala
```diff
@@ -4,8 +4,20 @@ import scala.language.implicitConversions
 import scala.util.matching.Regex
 
 object PredicateRegEx {
+  implicit def validCensusTractFormat: Predicate[String] = new Predicate[String] {
```
schbetsy: This should get a test like the ones @nickgrippin wrote in #364 , but that doesn't need to block merging this PR. I created an issue to remind us: https://github.com/cfpb/hmda-platform/issues/367

validation/src/test/scala/hmda/validation/rules/lar/validity/V300Spec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  property("Fails for valid MSA not NA in a small county") {
+    forAll(larGen) { lar =>
+      val invalidGeography = Geography("962100", "46", "045", "9621.00")
+      val invalidLar = lar.copy(geography = invalidGeography)
+      invalidLar.mustFail
+    }
+  }
+
+  property("Succeeds for NA track in a small county") {
```
schbetsy: `track` -> `tract`

validation/src/test/scala/hmda/validation/rules/lar/validity/V300Spec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  property("Succeeds for MD, state, county and tract all = NA") {
+    forAll(larGen) { lar =>
+      val validGeography = Geography("NA", "NA", "NA", "NA")
+      val validLar = lar.copy(geography = validGeography)
+      validLar.mustPass
+    }
+  }
+
+  property("Fails for valid MSA not NA in a small county") {
```
schbetsy: I believe the edit says it should fail if _tract_ is not NA for a small county. (Maybe a small county can still have an MSA; I'm not familiar enough with the domain to know.)

validation/src/test/scala/hmda/validation/rules/lar/validity/V300Spec.scala
kgudel: Oh semantic satiation

validation/src/test/scala/hmda/validation/rules/lar/validity/V300Spec.scala
kgudel: I meant to say that it is a valid combination of fields but tract is not NA and the county is small one. I'll fix the language.

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
kgudel: Unfortunately there are no duplicates in there.

lgdean: I'd suggest experimenting with `.toSet` even so.  It'll likely lead to faster lookup times, as you don't have to search the whole list.  See the changes in `ef8167c775d6df76d3721b4d9bcf69e347e93ddd` for examples.

On that note, I notice that we have a bunch of similar lookup methods happening to the `CBSATractLookup` values, and scattered around the rules.  I wonder whether there's something common and useful to extract.  Doesn't necessarily have to be in this PR, but seems worth considering.

schbetsy: Added an issue for this task: https://github.com/cfpb/hmda-platform/issues/369

schbetsy:
LGTM. 
@lgdean LGTY?

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
kgudel: Hey Laura, if there are advantages to be gotten I will happily switch over! However, would you mind explaining a bit more why it would be faster or pointing me to some documentation? Thanks!

lgdean: Strangely, a quick search didn't find anything handy. So I'll take a shot, with some (over-?)simplification.

A Scala List is basically a linked list. (What I'm about to say would also apply to an array.)  So when you're doing a `contains` in it, you end up traversing the whole list (in the `false` case) or on average half the list (in the `true` case).  In other words, it takes linear time.

A Scala Set, on the other hand, is basically* a hash table. The idea there is that it takes every element and does math (it's really cool, and worth reading about) so that lookup can happen in (effectively, usually) constant time.  For a large number of elements, it's usually a win.  Yes, "large" is a vague term!

*not exactly; for more info go [here](http://docs.scala-lang.org/overviews/collections/sets#sorted-sets) and scroll up a couple of paragraphs.

PR #364, opened 2016-05-27
Regex testing

### LINKED ISSUE #359: Test validation Regex
See [this comment](https://github.com/cfpb/hmda-platform/pull/350#discussion_r64759370) for more info.

Any regex used for validating data should have tests proving it matches the correct patterns.

nickgrippin:
These are all the test cases I could think of.  Let me know if there's any more I should add.

Closes #359

### DIFF
--- a/validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
+++ b/validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
@@ -1,21 +1,15 @@
 package hmda.validation.dsl
 
-import hmda.parser.fi.lar.LarGenerators
 import hmda.parser.fi.ts.TsGenerators
-import hmda.validation.engine.lar.syntactical.LarSyntacticalEngine
-import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.PropSpec
 import org.scalatest.prop.PropertyChecks
 import hmda.validation.dsl.PredicateRegEx._
-import hmda.validation.dsl.PredicateCommon._
 import hmda.validation.dsl.PredicateSyntax._
 import org.scalacheck.Gen
 
 class PredicateRegExSpec
     extends PropSpec
     with PropertyChecks
-    with MustMatchers
-    with LarGenerators
-    with LarSyntacticalEngine
     with TsGenerators {
 
   // Email **************
--- a/validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
+++ b/validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
@@ -8,6 +8,7 @@ import org.scalatest.prop.PropertyChecks
 import hmda.validation.dsl.PredicateRegEx._
 import hmda.validation.dsl.PredicateCommon._
 import hmda.validation.dsl.PredicateSyntax._
+import org.scalacheck.Gen
 
 class PredicateRegExSpec
     extends PropSpec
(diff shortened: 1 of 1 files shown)


validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+      phone not validPhoneNumber
+    }
+  }
+
+  property("An empty string will fail the phone regex") {
+    "" not validPhoneNumber
+  }
+
+  property("An improperly formatted phone number must fail the phone number regex") {
+    val testCases = List("123-456-789", "--", " - - ", "0-0-0", "abc-def-ghij")
```
kgudel: maybe include other common phone number formats such as (123) 456-7890.

nickgrippin: Good one!  I'll throw that in

validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+class PredicateRegExSpec
+    extends PropSpec
+    with PropertyChecks
+    with TsGenerators {
+
+  // Email **************
+
+  property("A valid email must pass the email regex") {
+    forAll(emailGen) { email =>
+      email is validEmail
```
schbetsy: The HMDA DSL doesn't work in tests. `email is validEmail` returns an _EditCheck_ `Success` or `Failure`, but that isn't the same as having a _spec_ expectation pass or fail. This line couldn't make the test fail, regardless of the input.

`validEmail.validate(email) mustBe true` would be a way to get the tests to reflect whether the validations pass or fail.

nickgrippin: Thanks!  Sorry about the oversight on my part, didn't catch that return type.

validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  property("An alphanumeric string will fail the email regex") {
+    forAll(Gen.alphaStr) { email =>
+      validEmail.validate(email) mustBe false
+    }
+  }
+
+  property("An empty string will fail the email regex") {
+    "" not validEmail
```
schbetsy: Looks like you missed this one when you were converting to `mustBe` syntax.

validation/src/test/scala/hmda/validation/dsl/PredicateRegExSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+
+class PredicateRegExSpec
+    extends PropSpec
+    with PropertyChecks
+    with TsGenerators
+    with MustMatchers {
+
+  // Email **************
+
+  property("A valid email must pass the email regex") {
```
schbetsy: Since the `emailGen` generator only generates a small subset of valid emails, I'd like to see another spec that shows specific examples, with things like numerals, symbols, etc. A few I thought of that pass:

``` 
    val testCases = List("what.e.v.e.r@example.co.uk", "6237468@example.gov",
      "f@a.ke", "lol@12345.lol", "hi__mom@example.club")
```

nickgrippin: Sure thing!  I'll throw those in

schbetsy:
LGTM

PR #357, opened 2016-05-25
simplify phoneGen and avoid compiler warning

lgdean:
addresses #110 
addresses #111 

I'd mentioned compiler warnings. Fixed one of them while in transit, by simplifying the code.  Came from PR #344 so I'm mentioning the same issues.

PR #366, opened 2016-05-27
cleanup: use stringOfN more widely in generators

lgdean:
Following on the heels of #357, I remembered we have a method for that.


nickgrippin:
This is great!  LGTM

PR #370, opened 2016-06-01
RegEx predicate refactor

lgdean:
Removed some code. The failure messages are all the same now! If you disagree with that plan, then we can add a message arg to `stringMatching`, but I didn't see the point.


validation/src/main/scala/hmda/validation/dsl/PredicateRegEx.scala
```diff
@@ -54,10 +54,10 @@ object PredicateRegEx {
 
   private def regExFor(pattern: String): String = {
     val result = pattern.map {
-      case 'N' => "[0-9]"
+      case 'N' => "\\d"
       case '.' => "\\."
     }
-    "^" + result.reduce(_ + _) + "$"
+    result.mkString("^", "", "$")
```
lgdean: I prefer this to the previous version. (Of course I do, or I wouldn't have made the change! Just commenting so it doesn't get lost in the noise.)

schbetsy:
Since #363 merged, `PredicateRegex` now has a `validCensusTractFormat` method. I'm assuming that's what's causing the merge conflicts, so we may as well change it to use the new format.

Nice refactor 🎉

nickgrippin:
LGTM!  This code looks so much nicer 👍

PR #371, opened 2016-06-01
Allow requests with an empty Host header

jmarin:
See [reference documentation](http://doc.akka.io/docs/akka/2.4.6/scala/http/configuration.html) (`default-host-header` setting)


schbetsy:
For anyone curious, I found this helpful for learning a bit more about the Host header: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

schbetsy:
LGTM

jmarin:
Also for context, this is necessary for health checks performed by our runtime platform on the root endpoint. Without a default `Host` header, they fail.

lgdean:
Thanks to you both for the context! Merging.

Issue #319, opened 2016-05-17
Describe Institutions Data Model


jmarin:

ericspry:
Thought that #381 was going to auto close this.

Issue #213, opened 2016-03-08
Validation (2017): Q595 (MSA/MD Number)


hkeeler:
**Edit #:** Q595
**Type:** Quality
**Category:** Loan Application Register (only)
**Column:** MSA/MD Number

If action taken type = 1-5, 7, or 8, then MSA/MD must = a corresponding respondent, MSA/MD combination on respondent panel, or NA. @NOTE: This edit is not applied to mortgage banking subsidiaries or independent mortgage companies. Their Metropolitan Statistical Areas or Metropolitan Divisions (if appropriate) will be determined at the end of the cycle based on the data reported. For depository institutions, this edit will produce the Q595 report but will not be counted in any error statistics.

Additional resources:
- [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
- [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
- [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

kgudel:
Marking this as blocked since it requires panel

lesgou:
Additional explanation: 
1) If the filer is a depository institution, and 
2) The reported loan is not a purchased loan, then

The reported MSA/MD for this loan should be listed on the panel as where the bank has a branch. Please indicate (yes/no) if you have a branch in this msa/md.

hkeeler:
Thanks @lesgou!  That's much clearer.

lesgou:
Updating requirements:

If institution type <9
and action taken type = 1, 2, 3, 4, 5, 7, or 8

Then list the MSA/MD where the filer does not have a branch office.

Include the following information:
MSA/MD ID
MSA/MD Name
Number of loans in this MSA/MD

nickgrippin:
Closing this ticket, edit will not be checked.

lesgou:
This edit is no longer relevant due to changes in architecture.

Issue #178, opened 2016-03-08
Validation (2017): Q033 (Parent Mailing Address)


hkeeler:
**Edit #:** Q033
**Type:** Quality
**Category:** Transmittal Sheet (only)
**Column:** Parent Mailing Address

If respondent is a bank, savings association, or independent mortgage company, and if any parent company exists, then parent name, address, city, state and zip code should not = blank.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

schbetsy:
Not sure how we'll know what kind of institution the respondent is (bank, savings association, mortgage co, etc). Will that information come from Panel? I didn't see it in the transmittal sheet.

jmarin:
This will come from panel, yes.

nickgrippin:
Adding a blocked label, as this requires panel

Issue #105, opened 2016-03-08
Validation (2017): V110 (Parent Mailing Address)


hkeeler:
**Edit #:** V110
**Type:** Validity
**Category:** Transmittal Sheet (only)
**Column:** Parent Mailing Address

If respondent is a mortgage banking subsidiary (MBS) or a nondepository institution that has an affiliation with a depository institution, then parent name of the MBS or the depository institution that the nondepository institution is affiliated with along with the address, city, state and zip code must not = blank.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

kgudel:
This seems like it requires panel.

jmarin:
@kgudel It does

nickgrippin:
Blocked by panel requirement

hkeeler:
Can we give this description a bit of love?  I'm having a hard time parsing the run-on sentence.

>If respondent is a mortgage banking subsidiary (MBS) or a nondepository institution that has an affiliation with a depository institution, then parent name of the MBS or the depository institution that the nondepository institution is affiliated with along with the address, city, state and zip code must not = blank.

First off, can we define what _affiliated with_ means?  Does that mean it's _parent_?  The term seems broader than that, but that's what makes the most sense in context with the rest of the description.

I _think_ the logic for this is supposed to be (hope this pseudo-code makes sense):

```python
if (respondent.type == 'mbs' or (respondent.type == 'non-di' and respondent.parent.type == 'di')) and
   (ts.parent_name == '' or ts.parent_address == '' or ts.parent_city == '' or ts.parent_state == '' or ts.parent_zip == ''):
        
    fail
else:
    pass
```
...because I _could_ read the first statement as...

```python
if ((respondent.type == 'mbs' or respondent.type == 'non-di') and respondent.parent.type == 'di') and...
```

...though I'm pretty sure this isn't the case after reading this a few dozen times.

I think I finally get it.  Can this be simplified to the following?

[...]

PR #381, opened 2016-06-06
For discussion: Institutions prototype

jmarin:
This PR is meant to gather additional requirements for how the _Institutions Panel_ is used. Initially, just documentation about the data model is provided, to elicit comments from subject matter experts on what is missing or should be corrected. After a few iterations a prototype with possible approaches to coding the institutions database APIs will be presented for discussion.

### DIFF
--- a/Documents/Institutions.md
+++ b/Documents/Institutions.md
@@ -1,0 +1,62 @@
+# Institutions Data Model
+
+This document describes the data model for Financial Institutions in HMDA. 
+A Financial Institution Database is created in HMDA for the purposes of 
+filing mortgage data. The information in this database is used in some edits, 
+as complementary information for publication and analysis of mortgage data, 
+and in the creation of the HMDA Panel. The HMDA Panel is the list of institutions
+that are required to file HMDA data, every year. 
+
+The Institutions Database needs to keep historical information for analysis, 
+and at least have information readily available for the two previous years in order
+to be able to perform the necessary checks when performing validation on the HMDA
+data being submitted or when composing the HDMA Panel. In addition to this, it must
+maintain the relationships between institutions (i.e. parent/child relationships)
+ and the actions that modify them (i.e. mergers, acquisitions). 
+ 
+ 
+The end product is a list of institutions that contains the necessary information 
+so that the HMDA Panel can be processed automatically. This table contains the following
+information:
+
+
+* `Activity Year` (integer length 4, i.e. 2017)
+* `Respondent ID` (string length 10)
+  Unique identifier for the financial institution for HMDA Panel
+* `Respondent RSSD ID` (string)
+  Unique identifier for financial institution in NIC. Present for depository
+  institutions, may not be present for non-depository institutions
+* `Agency Code` (integer, i.e. 9=CFPB)
+* `Name` (string)
+* `Assets` (double)
+* `Other Lender Code` (i.e. 1=Mtg Bkg Sub(MBS))
+* `Region` (integer length 2; Agency Office Code)
+* `Number of Loans` (integer)
+* `Charter Type`
+* `Charter Authorization`
+* `Deposit Insurance Code`
+* `FRB Membership Status`
+* `NCUA Membership Status`
+* `Legal Entity Identifier (LEI)` (string) (*)
+* `Respondent Address`
+* `Respondent City`
+* `Respondent State`
+* `Respondent FIPS State` number (string length 2)
+* `Parent Respondent ID`
+* `Parent RSSD ID`
+* `Parent Name`
+* `Parent Address`
+* `Parent City`
+* `Parent State`
+* `Top Holder RSSD ID`
+* `Top Holder Name`
+* `Top Holder City`
+* `Top Holder State`
+* `Top Holder Country`
+* `HMDA Panel` (integer length 1; 0=no, 1=yes)
+
+
+
+
+
+(*) These fields will most likely be empty in the 2017 collection.


Kibrael:
+1

wpears:
Any thoughts on whether filing status information should be included here? Say `Submission Count` or `Filing Status`?

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+ and the actions that modify them (i.e. mergers, acquisitions). 
+ 
+ 
+The end product is a list of institutions that contains the necessary information 
+so that the HMDA Panel can be processed automatically. This table contains the following
+information:
+
+
+* `Activity Year` (integer length 4, i.e. 2017)
+* `Respondent ID` (string length 10)
```
Kibrael: The respondent ID is not guaranteed to be unique unless it is concatenated with the agency code. This is because the RID can be one of the following: ID OCC, ID FDIC CERT, ID RSSD, Tax ID, NCUA ID, or in rare cases FHA number. There is no control that prevents duplication of a number in one set into another.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+  Unique identifier for the financial institution for HMDA Panel
+* `Respondent RSSD ID` (string)
+  Unique identifier for financial institution in NIC. Present for depository
+  institutions, may not be present for non-depository institutions
+* `Agency Code` (integer, i.e. 9=CFPB)
+* `Name` (string)
+* `Assets` (double)
+* `Other Lender Code` (i.e. 1=Mtg Bkg Sub(MBS))
+* `Region` (integer length 2; Agency Office Code)
```
Kibrael: Agency Region seems to be used to sort out work responsibility for panel updates inside the Federal Reserve System. I have not come across another use case for this field.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+* `Number of Loans` (integer)
```
Kibrael: I would recommend splitting this field into two fields, count of closed end loans and count of open end lines of credit. Each count has a different regulatory requirement for reporting: 25 for closed end, 100 for HELOC. Currently these counts are only available in LAR data and through the NCUA for its members.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+* `Parent Address`
+* `Parent City`
+* `Parent State`
+* `Top Holder RSSD ID`
+* `Top Holder Name`
+* `Top Holder City`
+* `Top Holder State`
+* `Top Holder Country`
+* `HMDA Panel` (integer length 1; 0=no, 1=yes)
+
```
Kibrael: Recommend adding a flag for Large FI as determined by this procedure: https://github.cfpb.gov/data/ss-supervised_entity/blob/master/Procedures/dco.GenerateSupervisedEntityList.StoredProcedure.sql

This is the data team's procedure that creates a list of entities supervised by the CFPB. This list is used in panel creation to move banks, their affiliates, and subsidiaries from other regulator's panel sections (agency codes) to the CFPB (agency code 9). This list is also published externally.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+
+
```
Kibrael: Recommend including entity_type from NIC as it is useful in working with Mortgage Banking Subsidiaries (MBS) and holding companies.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+
```
Kibrael: It may be worth having a discussion about the Federal Reserve System's "Institution Type". This field is derived internally at the FRS, not made public, but is used in segmenting the panel institutions to roll forward some but not others. For definitions and logic (not complete) of institution type please see Panel Data Spec at https://drive.google.com/drive/u/1/folders/0B667BGHwjTznMGR4MEE4eVJFdjQ

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+maintain the relationships between institutions (i.e. parent/child relationships)
+ and the actions that modify them (i.e. mergers, acquisitions). 
+ 
+ 
+The end product is a list of institutions that contains the necessary information 
+so that the HMDA Panel can be processed automatically. This table contains the following
+information:
+
+
+* `Activity Year` (integer length 4, i.e. 2017)
```
hkeeler: Isn't this coming from HMDA transmittal sheet?  It doesn't seem appropriate here.  We're not going to have this data coming from NIC and other non-HMDA data sources, right?

However, an `is_active` does seem useful.

Documents/Institutions.md
hkeeler: Do we want a pre-calculated `Respondent ID`?...and could we even pre-calculate accurately it if we wanted to?  It seems more appropriate to have the underlying IDs as @Kibrael indicated above, and attempt the `id + agency_code` resolution against those at filing time.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+* `FRB Membership Status`
+* `NCUA Membership Status`
+* `Legal Entity Identifier (LEI)` (string) (*)
+* `Respondent Address`
+* `Respondent City`
+* `Respondent State`
+* `Respondent FIPS State` number (string length 2)
+* `Parent Respondent ID`
+* `Parent RSSD ID`
+* `Parent Name`
```
hkeeler: Why do we have the parent's data here?  Why not just the identifier.  This could probably just be the `parent_rssd_id` for the first iteration, though we may want our own identifier if we think there's a chance we'll have filers that aren't in NIC.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+* `Parent Address`
+* `Parent City`
+* `Parent State`
+* `Top Holder RSSD ID`
+* `Top Holder Name`
```
hkeeler: Same as above.  Probably just an identifier would suffice.  We may not even need this if we structure the data such that this can be derived easily through parent hierarchy.

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+* `Top Holder City`
+* `Top Holder State`
+* `Top Holder Country`
+* `HMDA Panel` (integer length 1; 0=no, 1=yes)
```
hkeeler: This seems more appropriate as a `required_to_file` flag on what we're currently calling `period` in the API.

hkeeler:
Per @wpears:
>Any thoughts on whether filing status information should be included here? Say `Submission Count` or `Filing Status`?

To me, that seems more appropriate as data about the `period`.  You'll want that data for each filing, not just the most recent.

Documents/Institutions.md
Kibrael: @hkeeler matching institutions (especially to NMLS) can require address information as there are very limited keys between datasets.

Documents/Institutions.md
debseidner: We should store history of Respondent ID to help with publication as part of the panel data. If an institution files in 2014 under one respondent id and agency code combo and then another in 2015 and then another in 2016 (maybe goes from being  a member bank to not being a member bank to being a member bank), we will need to be able to link all that history to track HMDA filings for historic reporting.

Kibrael: Regarding Debbie's comment: if we want to go far down this thread we may find good use for the Avery file.

debseidner:
@wpears 

Any thoughts on whether filing status information should be included here? Say Submission Count or Filing Status?

I would agree that we need to store whether an institution filed or not (so filing status) for each year (1) for reporting that an institution did or did not report for enforcement purposes and (2) to know that an institution was on panel for a given year and (3) I'm sure there are benefits for historical reporting - to know if you should or should not include that institution in historic reporting.

Documents/Institutions.md
debseidner: Is activity year more of an indicator that the institution sent a HMDA filing for a particular year? Are we envisioning one row for each institution per year? It makes sense to track history of which years institutions file.

Do we need a date for when this information is as of? For instance, we needs assets as of 12/31 of the prior year (so 12/31/2016 for 2017 data). Even if we get an update to assets for an institution on say a refresh on 4/30/2017, we do not want to update the assets for an institution because all that matters is what they were on 12/31 of the prior year. The same goes for MSA - it is based on which branches were in existence on 12/31 of the prior year. Just something to consider if we end up updating data monthly or quarterly that we will need to indicate as of what date the information contained on each record is.

Documents/Institutions.md
debseidner: +1 Agree with David on this; was going to suggest the same thing.

jmarin:
The key word is _`update`_. With an event sourcing model, we shouldn't have to lose information, and can precisely tell what the value of the assets field is, for any point in time (going back to a specific date, and up to current state).

debseidner:
Should we discuss whether we want to store any of these indicators (maybe in a separate table that links to this information) but this way we have a link of what identifier we used to link this institution to the data source? Here are some sample identifiers that we may use - FI RSSD, FI FDIC Certification Number, FI ID OCC, FI ID NCUA, Thrift ID, NMLSID, Filing_ID for MCR report. The filing_id for the MCR report may need to be tied to a quarter too as MCR data is updated quarterly.

debseidner:
Do we need to consider home or branch office information for each institution to pair with the metadata about an institution?

Documents/Institutions.md
```diff
@@ -0,0 +1,62 @@
...
+  institutions, may not be present for non-depository institutions
+* `Agency Code` (integer, i.e. 9=CFPB)
+* `Name` (string)
+* `Assets` (double)
+* `Other Lender Code` (i.e. 1=Mtg Bkg Sub(MBS))
+* `Region` (integer length 2; Agency Office Code)
+* `Number of Loans` (integer)
+* `Charter Type`
+* `Charter Authorization`
+* `Deposit Insurance Code`
```
debseidner: Is this meant to store information on if an institution federally insured or regulated?

Documents/Institutions.md
debseidner: @Kibrael Do we need to also store dollar value of loans for depositories since we don't have count?

debseidner: We might need to store loan origination volume in addition to loan counts as for non-DIs, we may need to calc if the institution’s home purchase loan originations (including refinancings of home purchase loans) equal or exceed 10 percent of its loan origination volume, measured in dollars, or equal $25 million or more.

debseidner:
Do we want to store CRA filer flag (y/n) on each institution record by activity year with the metadata for each institution?

debseidner:
Do we need to have fields dedicated to tracking merger and acquisition (M&A) activity - like whether the institution is no longer active because it was consumed by another institution or whether there used to be an older parent but the parent info here is the latest parent? Do we need to add a field that specifies who the other institution is that the panel participant merged with or who the surviving institution was (like a related institution ID field)? Or if a new institution and formed from two past institutions, which were the two past institutions that this new one is the survivor of (this would be a situation where tax-id changes and there is a new institution that supercedes both).

debseidner:
We should consider adding tax ID as an indicator in addition to LEI, respondent ID and agency code. We might need that to transition from agencycode/respondentID to LEI.

Documents/Institutions.md
hkeeler: @Kibrael, interesting.  So, we use these as pseudo foreign keys for fuzzy matching when we have no other identifiers?  Well, that's gross, but I can see how we may need that.

I still don't think that data needs to be _here_, though.  Can't it remain on the parent record, and just have a real foreign key back to the parent.  I'm concerned that every time a given institution's address info changes, you have to track down all the child records that must be updated as well.  I think things could get out of sync pretty easily.

Of course, under the model I'm proposing, we would need to create our own identifier since there is no reliable id that spans all institutions types.  Perhaps this could be replaced by LEI once that's stable, but I suspect there's going to be holes there as well for some time.

Documents/Institutions.md
Kibrael: Storing aggregate loan value would only be necessary if we decided to use that as a proxy for count, in which case we should talk about that methodology (after gently speaking with Feng) and consider if we want to make that process transparent and if we will store our estimated count as well.

Documents/Institutions.md
Kibrael: @hkeeler matching parent with address and contact information is not great, but sometimes it's what we have. 

I'm not partial to where parent information sits as long as it is available to the system when we need it.

Documents/Institutions.md
Kibrael: Yes, the deposit insurance code shows under which program the institutions deposit accounts are insured.

Kibrael:
@debseidner, what do you mean by using Tax ID as an indicator?

I would recommend keeping all of the respondent ID dependancies in the Panel. Having them available makes it much easier for data users to back out of the panel process. The Avery file contains all these indicators. @feomike had previously requested that we include Avery file components when doing so was not difficult.

Documents/Institutions.md
debseidner: That is a good idea too to store the estimated loan count separately from the loan count if we decide to store it.

debseidner:
@Kibrael Sorry, was using indicator instead of identifier. I meant that we should store tax id as an identifier - something that could link between different data sets like MCR and NIC and the panel info or a way to identify an institution when a new applicant tries to create an account.

Kibrael:
@debseidner +1 for Tax ID as an identifier.

hkeeler:
I think we're trying to solve too many problems in this one issue.  It is impossible to model for everyone's needs in this one issue.  I propose that we scope down the initial version, and gradually add institution attributes and refactor as we go.

I think the initial version should be purely for supporting the needs of HMDA filing.  We understand what data elements we need for this much better than what all is needed to _build the panel_.

I'm going to model this in terms of the API as I don't know what the underlying datastore will be and what implications Akka Persistence adds to the mix.

## Institution Model

#### Option 1 - bare bones

This may be all we need to get started.  All the end user needs to see, in most cases, is the list of institutions (or just a single institution), and then they can start filing.

* `id` - A unique CFPB generated ID.  This is going under the assumption that there is no other unique ID present across all other datasets.
* `name` - Bank's name
* `status` - Seems useful for filtering out old institutions.  This may not be necessary for our first year, since we probably won't have any old banks in the system, but I'll leave it here for conversation.  Also, if the only two states are `active` and `inactive`, this _could_ be switched to `is_active: true|false`...but I generally like the more flexible `status`.

```json
{
  "id": 12345678,
  "name": "Bank of Springfield",
  "status": "active"
}
```

#### Option 2 - w/ IDs and regulator

This option adds additional relevant IDs.  This seems useful for a couple reason:

* The user is associated with multiple institutions, perhaps with similar/same names, and they need more info to disambiguate the records.
* It may help clarify _Respondent ID_  issues if they know what _we_ think their IDs and regulator should be.

```json
{
  "id": 12345678,
  "name": "Bankerton Bank",
  "status": "active",
  "federal_regulator": {
    "code": "fdic",
    "name": "Federal Deposit Insurance Corporation (FDIC)"
  },
  "external_ids": [
    {
      "code": "fdic-certificate",
      "name": "FDIC Certificate Number",
      "value": "12345",
    },
    {
      "code": "ein",
      "name": "Federal Tax Id / Employee Identification Number (EIN)",
      "value": "123456789",
    },
    {
      "code": "ffiec-nic-rssd-id",
      "name": "FFIEC's NIC RSSD Identifier",
      "value": "1234567"
    }
  ]
}
```

#### Option 3 - w/ embedded filings (aka periods)

It would be convenient for the frontend to have `filings` embedded within an `institution`.  This is not directly related to modeling an _institution_, and also isn't _required_ by the frontend, but I thought it'd be useful to include it here.

See https://github.com/cfpb/hmda-platform/pull/411#issuecomment-226312782 for further discussion on this topic.

jmarin:
@hkeeler I've implemented Option 3 as part of #422

hkeeler:
We've decided that we're going to focus only on data elements required by the edit checks.

I believe the following are the minimum required:

* Identifiers
    * FFIEC's RSSD ID
    * Federal Tax ID (EIN)
    * FDIC Certification Number
    * OCC Charter ID
    * NCUA Charter ID
* Regulating Agency (CFPB, FDIC, FRS, HUD, NCUA, OCC)
* Institution Type
* Location
    * Street Address
    * Metropolitan Statistical Area / Metropolitan Division (MSA/MD)
* Parent Institution
* Branch Locations
* Community Reinvestment Act (CRA) Flag

In addition, we will also include the institution's name and our own unique identifier per institution record.

Does this seem accurate?  Are there any other fields I'm missing required by the remaining edit checks?

lesgou:
:+1:

Kibrael:
Hey Hans:
Are you going to put Respondent ID in this section? 

I'd like to add Thrift ID (these are savings and loan banks that were absorbed by other regulators after OTS failed).

Regarding Institution Type and Regulating Agency, there are many fields that need to be checked to determine these (see the Derived Logic tab of the Panel Data Spec.xlsx in the Drive Panel folder).

Is it worth clarifying that location is likely head office location?

hkeeler:
@Kibrael, good questions.

>Are you going to put Respondent ID in this section?

Isn't _Respondent ID_ derived from the fields above?  You just have to know which identifier to pull from based on the appropriate agency, right?  If that is the case, I don't think we should persist it as its own data field, but we should probably create a standard function for deriving it so we don't have that logic scattered all over the place.

>I'd like to add Thrift ID (these are savings and loan banks that were absorbed by other regulators after OTS failed).

I meant to ask about this one.  I left it off because it didn't seem like we would be receiving any new submissions where we'd need to check _Thrift ID_ as part of any edits.  It's not in the [_1.4 - Respondent Identification Numbers for 2017 HMDA Filers_](http://www.consumerfinance.gov/data-research/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf#page=3) section of the 2017 file spec.  I'm fine adding it though if we do actually need it or it provides extra clarity in some way.

>Regarding Institution Type and Regulating Agency, there are many fields that need to be checked to determine these (see the Derived Logic tab of the Panel Data Spec.xlsx in the Drive Panel folder).

I did sort of assume that I was oversimplifying _Institution Type_ a bit, and that it was likely multiple fields, or derived from multiple fields in the source data.  We need data for implementing the following rules:

* [Q959](https://github.com/cfpb/hmda-platform/issues/213) - _...this edit is not applied to **mortgage banking subsidiaries** or **independent mortgage companies**...._ and _...for **depository institutions**, this edit will produce the Q595 report but will not be counted in any error statistics..._.
* [Q033](https://github.com/cfpb/hmda-platform/issues/178) - _If respondent is a **bank**, **savings association**, or **independent mortgage company**..._
* [V110](https://github.com/cfpb/hmda-platform/issues/105) - _If respondent is a **mortgage banking subsidiary (MBS)** or a **nondepository institution** that has an affiliation with a **depository institution**, then parent name of the **MBS** or the **depository institution** that the **nondepository institution** is affiliated with ..._

...so we'll definitely need whichever fields are necessary for deriving those **bolded** attributes.  Similar goes for _Regulating Agency_.  I'll read up a bit more on the doc you mentioned, and try to figure out the best path forward.

>Is it worth clarifying that location is likely head office location?

I'm not sure.  Can you expand on why?  Is there a specific edit you're thinking of?

I'm currently thinking we'd structure this similar to NIC data where all _entities_ (FIs, parents, branches, etc.) are the same data structure, with references to parents/children, each of which would have their own location data, and could be looked up that way.  Does that make sense?

Kibrael:
@hkeeler 

I think I'm confused in the difference between the fields you posted and the inputs required to get them. 

The bold fields listed in the edits section will require a lot of input fields as determining institution type is a step beyond determining agency code. 

Thrift ID is probably just for backwards compatibility. 

The head office was in relation to location. I thought there were checks to make sure we got the actual head office location.

lesgou:
There are a few edits that check that parent address is not blank and <> institution address. Q033 <https://github.com/cfpb/hmda-platform/issues/178> is a little different.

hkeeler:
After much discussion on the edits requiring institution data, I think we're finally clear to close this one out.  I've updated _Institution Type_ to include the types explicitly mentioned in the edits, and noted that each will need further categorization of _depository / non-depository_.

* Identifiers
    * FFIEC's RSSD ID
    * Federal Tax ID (EIN)
    * FDIC Certification Number
    * OCC Charter ID
    * NCUA Charter ID
* Regulating Agency (CFPB, FDIC, FRS, HUD, NCUA, OCC)
* Institution Type (Bank, Savings and Loan, Credit Union, etc.)
    * Each having further categorization of: Depository / Non-depository
* Location
    * Street Address
    * Metropolitan Statistical Area / Metropolitan Division (MSA/MD)
* Parent Institution
* Branch Locations
* Community Reinvestment Act (CRA) Flag

Further discussion on this can play out on the PRs to come. :)

ericspry:
closes #319

PR #386, opened 2016-06-06
Q013, Q036, Q037

kgudel:
Closes #189 
Closes #190 
Closes #191 

Further makes LarHttpApiSpec only check for syntax and validity

PR #389, opened 2016-06-06
Fix V300

### LINKED ISSUE #385: V300 should not enforce tract="NA" for small county

kgudel:
Closes #385


kgudel:
Requires #386

nickgrippin:
Code LGTM.  @debseidner, does this fix look good to you?

debseidner:
@nickgrippin @kgudel Have one question

I see this:
 when(tract not equalTo("NA")) {
        validFormat and
          validCensusTractCombination and
 -        tractStateCountyCombination and
 -        smallCounty

Is there a check that MSA, Census, County and State are all a valid combination if MSA does not = NA? Also, what does smallcounty mean there?

kgudel:
1. Yes. `validCensusTractCombination` checks whether all geo fields are valid when MSA not "NA".

2. Small county is true if the county is marked as small and false if not.

debseidner:
@kgudel Thanks for the explanations. @nickgrippin All good.now. Thanks again for checking.

PR #390, opened 2016-06-07
Q032

nickgrippin:
Closes #194 

Includes modification of test files to pass this quality check.

Depends on #386 to update SingleLarValidationSpec

PR #387, opened 2016-06-06
Q025, Q038

nickgrippin:
Closes #192 
Closes #193

PR #391, opened 2016-06-07
Q014 fix

kgudel:
Small fix to problem causing tests to fail.


kgudel:
Should solve the problem in PRs #390 and #387

PR #393, opened 2016-06-07
call DSL containedIn without wrapping domain values in a List

lgdean:
Basically, I got tired of seeing `List(1, 2, 3)` everywhere; it felt like clutter.  So I added this to the DSL.

Overall I think it's a good idea, but I'm open to discussion of whether it's always good to use.  In particular, please give `V347` a look and decide what you think.  Hard to judge one's own code sometimes!

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/PredicateCommon.scala
@@ -41,6 +41,8 @@ object PredicateCommon {
     override def failure: String = s"not between $lower and $upper (inclusive)"
   }
 
+  implicit def containedIn[T](first: T, rest: T*): Predicate[T] = containedIn(first +: rest)
+
   implicit def containedIn[T](domain: Seq[T]): Predicate[T] = new Predicate[T] {
     override def validate: (T) => Boolean = domain.contains(_)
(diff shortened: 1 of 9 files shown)


nickgrippin:
I can definitely see how `V347` could be confusing.  Defining the `List` there makes it clear that we're only looking for 1 or 6, but if a user sees `containedIn`, they might wonder if we were looking for any values 1 through 6.

This might be a weird naming suggestion, but maybe `oneOf`?  It reads fairly well, and might be a bit clearer to the user.  I'm open to other suggestions though, or just keeping it as is.

kgudel:
I feel like its a nice way to clean things up a bit but I am all aboard with Nick. I also like `oneOf`.

lgdean:
I also like `oneOf`, and then I don't have to do the awkward first/rest thing with the args.  I'll leave it alone for a little while to give others a chance to weigh in while I review other PRs, because we have a lot of inventory right now.  (If nobody weighs in, that's fine, and I declare `oneOf` the consensus in that case.)

jmarin:
+0 from me. I see this as syntactic sugar, which _can_ be nice, but imho is not critical.

lgdean:
Renamed to `oneOf`; thanks for the suggestion!

nickgrippin:
LGTM

PR #394, opened 2016-06-07
minor cleanup: use states.keySet instead of .keys.toList

lgdean:
This change is not motivated by performance concerns (though it may be slightly better on that front). it just seems a bit cleaner.


nickgrippin:
Looks great!

PR #399, opened 2016-06-08
Q027

### LINKED ISSUE #199: Validation (2017): Q027 (Income)
**Edit #:** Q027
**Type:** Quality
**Column:** Income

If action taken type = 1-5, 7 or 8, and property type = 1 or 2, then applicant income should not = NA.

Additional resources:

schbetsy:
Closes #199

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/quality/Q027Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/quality/Q027Spec.scala
@@ -1,0 +1,42 @@
+package hmda.validation.rules.lar.quality
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.rules.EditCheck
+import hmda.validation.rules.lar.LarEditCheckSpec
+
+class Q027Spec extends LarEditCheckSpec {
+  property("fails when actionTaken = 1,2,3,4,5,7,8 and propertyType = 1,2 and income NA") {
+    forAll(larGen) { lar =>
+      whenever(lar.actionTakenType != 6 && lar.loan.propertyType != 3) {
+        val invalidApplicant = lar.applicant.copy(income = "NA")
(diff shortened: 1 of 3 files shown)


validation/src/test/scala/hmda/validation/rules/lar/quality/Q027Spec.scala
```diff
@@ -0,0 +1,42 @@
...
+    forAll(larGen) { lar =>
+      whenever(lar.applicant.income != "NA") {
+        lar.mustPass
+      }
+    }
+  }
+
+  property("passes when action taken not 1,2,3,4,5,7,8") {
+    forAll(larGen) { lar =>
+      val validLar = lar.copy(actionTakenType = 6)
```
kgudel: I think that the only issue would be if we think that the set of possible values for a field might change. This would be a question for @debseidner .

debseidner: @schbetsy @kgudel It is unlikely that the valid values will change but we have the ability to change them by taking the data spec out of the reg so we cannot rule that out. You can probably leave as-is with the caveat (or team understanding) that if valid values ever change, the team needs to go back and update tests like this to be correct for the new valid values. Hope this helps!

kgudel: Maybe something to bring up in discussion then.

jmarin: @schbetsy I think that is an acceptable way to test this edit. For those that have values not contained in a field (i.e. a fixed $ amount) we decided to put them in configuration, as they do change through time. As @debseidner said, if the file spec changes and we accept new or different values, our codebase would need to be updated.

jmarin:
LGTM

PR #402, opened 2016-06-09
Q029 + Predicate Geo

### LINKED ISSUE #211: Validation (2017): Q029 (MSA/MD,State,County,Census Tract)
**Edit #:** Q029
**Type:** Quality
**Column:** MSA/MD,State,County,Census Tract

If the reported state/county combination (when county is small and reported tract equals NA), or state/county/census tract combination is valid then the MSA/MD should not = NA.

Additional resources:

### LINKED ISSUE #369: Improve Geography Edit Checks
Taken from https://github.com/cfpb/hmda-platform/pull/363#discussion_r65217557:

> I notice that we have a bunch of similar lookup methods happening to the `CBSATractLookup` values, and scattered around the rules. I wonder whether there's something common and useful to extract. Doesn't necessarily have to be in this PR, but seems worth considering.

kgudel:
I'm putting this up as WIP because its a bit of a doozy and the code is all there given a correct interpretation. This pull request moves all the geo logic into PredicateGeo and closes #211 . 

For Q029 I have coded against the intention, as I understand it, with help from Debbie, instead of against the edit text. The edit as I have coded should say similar to:
If the report state/county combination is in an MSA then MSA should not = "NA". I have removed the tract combination entirely because, in my understanding, MSAs are comprised of counties, and so all tracts within a county will be in the same MSA or they will all not be in an MSA. @debseidner it would be great if you could confirm my interpretation of the edit.

I believe that this also closes #369


debseidner:
@kgudel LARs should appear on this edit if the loan was "reported with NA in the MSA/MD field; However, according to the reported state, county and census tract, the property is in a valid MSA/MD." It therefore seems like tract should factor into this. Can you add that back in?

Something else to consider is that we need to confirm that the census tract/state/county combo is not in a valid micropolitan statistical area. Micropolitan statistical areas are not part of HMDA. So if there is a valid combination but the MSA is actually a micropolitan statistical area, the NA in the MSA/MD field is valid. This was something we had wrong in Pilot.

Per the edit document:
"NA combination can be reported on property located outside the metropolitan statistical area/metropolitan division in which an institution has a home or branch office, or outside any metropolitan statistical area/metropolitan division...NA is not valid except in those cases where the property is located in a Micropolitan Statistical Area or is not located in any Metropolitan or Micropolitan Statistical Area; and in those examples, NA would only be reported in the MSA/MD column and the appropriate state, county, and census tract numbers would be reported."

kgudel:
@feomike are tracts important to determine MSA/MD?

Kibrael:
I think you can determine MSA/MD from the state/county combination.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,114 @@
...
+    (cbsa.geoIdMsa, cbsa.state, cbsa.county)
+  }.filter(x => x._1 != "").map { x =>
+    (x._2, x._3)
+  }.toSet
+
+  val hasMdSet = cbsaTracts.map { cbsa =>
+    (cbsa.metDivFp, cbsa.state, cbsa.county)
+  }.filter(x => x._1 != "").map { x =>
+    (x._2, x._3)
+  }.toSet
```
lgdean: `hasMsaSet` and `hasMdSet` seem unused. Leftovers from a previous approach, or something else?

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,114 @@
...
+    (cbsa.GEOIOD, cbsa.MEMI)
+  }.filter(x => x._2 == 1).map { x =>
+    x._1
+  }
+
+  val hasMsaNotMicroSet = cbsaTracts.map { cbsa =>
+    (cbsa.geoIdMsa, cbsa.state, cbsa.county)
+  }.filter(x => MsaNotMicro.contains(x._1)).map { x =>
+    (x._2, x._3)
+  }.toSet
```
lgdean: I propose something like:
```scala
  private val msaNotMicro = cbsaMetroMicro.filter(x => x.MEMI == 1).map { x => x.GEOIOD }.toSet
  val hasMsaNotMicroSet = cbsaTracts.filter(x => msaNotMicro.contains(x.geoIdMsa)).map { x =>
    (x.state, x.county)
  }.toSet
```
Maybe the second can also be private, and maybe you'd want to use underscore instead of `x` in some of those places, but that's the basic idea: do the mapping and then the tuple-ing so the names are still meaningful, and make sure the first one is a Set.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,114 @@
...
+  val smallCounties = cbsaTracts
+    .filter { cbsa => cbsa.smallCounty == 1 }
+    .map { cbsa => (cbsa.state, cbsa.county) }
+    .toSet
+
+  implicit def smallCounty: Predicate[Geography] = new Predicate[Geography] {
+    override def validate: (Geography) => Boolean = _.asInstanceOf[AnyRef] match {
+      case geo: Geography => smallCounties.contains((geo.state, geo.county))
+      case _ => false
+    }
```
lgdean: I prefer something like this (or perhaps even less code could work, but this certainly does):
```scala
    override def validate: (Geography) => Boolean = (geo) => smallCounties.contains((geo.state, geo.county))
```
Same deal for all the following predicates.
This may also improve the code coverage numbers. ;-)

validation/src/main/scala/hmda/validation/rules/lar/quality/Q029.scala
```diff
@@ -0,0 +1,19 @@
...
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax._
+import hmda.validation.dsl.PredicateGeo._
+
+object Q029 extends EditCheck[LoanApplicationRegister] {
+  override def apply(lar: LoanApplicationRegister): Result = {
+    val geo = lar.geography
+    when(geo is shouldHaveMsa) {
```
lgdean: Hm. This doesn't quite read well, and it puts the logic of the rule outside of the rule. (If I understand correctly, the only reason it should have an MSA is because of Q029.) Is there a way to name the predicate that describes what it actually checks? (And, perhaps more challenging: ... while not being ridiculously long?)

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
```diff
@@ -1,36 +1,19 @@
...
-  val cbsaTracts = CBSATractLookup.values
-
-  val validCombinations = cbsaTracts.map { cbsa =>
-    (cbsa.state, cbsa.county, cbsa.geoIdMsa)
-  }.toSet
-
-  val validMdCombinations = cbsaTracts.map { cbsa =>
-    (cbsa.state, cbsa.county, cbsa.metDivFp)
-  }.toSet
-
```
lgdean: I very much like that you moved this stuff out of here, and into a shared place with similar code.

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
```diff
@@ -1,36 +1,19 @@
...
-    val county = input.geography.county
-
-    val combination = (state, county, msa)
-
-    when(msa not equalTo("NA")) {
-      (combination is containedIn(validCombinations)) or
-        (combination is containedIn(validMdCombinations))
+  override def apply(lar: LoanApplicationRegister): Result = {
+    when(lar.geography.msa not equalTo("NA")) {
+      lar.geography is validStateCountyMsaCombination
```
lgdean: I'm somewhat less convinced of this part (just this one line, I mean -- the rest is good). I think "valid" is a bit vague in that predicate, because it's specifically about this one rule, V290.  I don't know whether I'd prefer just a different naming, or two predicates, or what. But I do feel like the logic is a bit obscured. (You may be able to convince me it's fine, though.)

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,114 @@
...
+  }.toSet
+
+  val validStateCountyCombinationSet = cbsaTracts.map { cbsa =>
+    (cbsa.state, cbsa.county)
+  }.toSet
+
+  val smallCounties = cbsaTracts
+    .filter { cbsa => cbsa.smallCounty == 1 }
+    .map { cbsa => (cbsa.state, cbsa.county) }
+    .toSet
```
lgdean: I find this one quite readable, for example.

And here's a question: do we want it private? I vote yes, because it's pairs of strings, which doesn't seem great for outside code to deal with (state and county? county and state?), but there may be reasons to do it this way.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
kgudel: These are left over, will remove.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
kgudel: I like this change a lot, good to remove the unnecessary command in msaNotMicro, and making the val private is a good change that I had not considered.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
kgudel: You are a rockstar. I was looking for exactly this!

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
kgudel: What would you say to the word `existing` as in `existingStateCountyCombination`? What I want to say is that that combination of State, County, and MSA is a real place.

schbetsy:
@kgudel Would you mind putting your current interpretation of Q029 into #211 ? That would make it easier to make sure we have consensus about how this edit should work.

validation/src/main/scala/hmda/validation/rules/lar/validity/V290.scala
lgdean: Hm. Now I see what you mean, and I think it's OK. I am however starting to wonder about the various meanings of `Msa` in the predicates, but that's a comment for another line.

model/src/main/scala/hmda/model/census/CBSAMetroMicroLookup.scala
```diff
@@ -0,0 +1,54 @@
...
+        INTPTLAT,
+        INTPTLON
+      )
+    }.toSeq
+  }
+}
+
+case class CBSAMetroMicro(
+  CBSAFP: String,
+  GEOIOD: String,
```
lgdean: Is this a value that is (potentially) the same sort of thing as a `CBSATract.geoIdMsa`? Because if so, then I suspect it is mis-spelled (extra `O`), and something like `geoId` would be better.

And probably in general, we want to use a camel case naming convention more like in `CBSATract`.  (And we may want to start using an explicit style guide, but that's outside the scope of this PR.)

model/src/main/scala/hmda/model/census/CBSAMetroMicroLookup.scala
```diff
@@ -0,0 +1,54 @@
...
+  name: String,
+  nameLSAD: String,
+  LSAD: String,
+  MEMI: Int,
+  MTFCC: String,
+  ALAND: String,
+  AWATER: String,
+  INTPTLAT: String,
+  INTPTLON: String
```
lgdean: What are these? I was able to find MEMI and it makes sense, particularly in context, but the rest are mysterious (and not explained in the same place, alas). I understand it's something about land vs water (area, perhaps, if my search helped), and I'd guess latitude and longitude, but it's still pretty opaque. I suspect a quick pointer to definitions would clear things up. (I might still ask whether we need it in the code right now, but you may have looked ahead in ways that I have not. Normally I'm all about YAGNI, but if we *are* gonna need it then yes please, let's document what each field is, in code.)

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
```diff
@@ -1,66 +1,35 @@
...
-    val validCensusTractCombination = when(msa not equalTo("NA")) {
-      (allCombination is containedIn(validMsaCombination)) or
-        (allCombination is containedIn(validMdCombination))
-    }
-
-    val tractStateCountyCombination = when(msa is equalTo("NA")) {
-      stateCountyCombination is containedIn(validStateCountyCombination)
-    }
-
     val validFormat = tract is validCensusTractFormat
```
lgdean: `validFormat` is calculated here, but then used in only one of the branches below.

validation/src/main/scala/hmda/validation/rules/lar/validity/V300.scala
```diff
@@ -1,66 +1,35 @@
...
-
     when(tract not equalTo("NA")) {
       validFormat and
-        validCensusTractCombination and
-        tractStateCountyCombination
+        when(msa not equalTo("NA")) {
+          (lar.geography is validCompleteCombination)
+        } and
+        when(msa is equalTo("NA")) {
+          (lar.geography is validStateCountyTractCombination)
```
lgdean: My IDE complains about unnecessary parens, here and on line 25. I guess I agree, though not as strongly as in some earlier cases.

model/src/main/scala/hmda/model/census/CBSAMetroMicroLookup.scala
```diff
@@ -0,0 +1,54 @@
...
+import java.io.File
+
+import scala.io.Source
+
+object CBSAMetroMicroLookup {
+  val values: Seq[CBSAMetroMicro] = {
+    val lines = Source.fromFile(new File("model/src/main/resources/tl_2013_us_cbsa.csv")).getLines
+    lines.map { line =>
+      val values = line.split('|').map(_.trim)
+      val CSAFP = values(0)
```
lgdean: This val is unused. I may have more to say about that based on the outcome of other parts, but I'm noting it now so we don't forget.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  implicit def validStateCountyMsaCombination: Predicate[Geography] = new Predicate[Geography] {
+    override def validate: (Geography) => Boolean = (geo) =>
+      validMsaCombinationSetNoTract.contains((geo.msa, geo.state, geo.county)) ||
+        validMdCombinationSetNoTract.contains((geo.msa, geo.state, geo.county))
+    override def failure: String = "state, county, msa, and census tract combination is not valid"
+  }
+
+  implicit def stateCountyCombinationInMsa: Predicate[Geography] = new Predicate[Geography] {
```
lgdean: So. Here, `Msa` seems to mean a metropolitan, not micropolitan, statistical area.
But in the previous predicate, `Msa` seems to mean MSA (either kind) or MD.
We probably want different names for these concepts.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,87 @@
...
+    (cbsa.geoIdMsa, cbsa.state, cbsa.county)
+  }.toSet
+
+  private val validMdCombinationSetNoTract = cbsaTracts.map { cbsa =>
+    (cbsa.metDivFp, cbsa.state, cbsa.county)
+  }.toSet
+
+  private val MsaNotMicro = cbsaMetroMicro
+    .filter(_.MEMI == 1)
+    .map { _.GEOIOD }
```
lgdean: This probably should be a Set, though I haven't measured anything.  (And for that matter, I don't know whether anything here is the most efficient algorithm, and it'd be premature to do much optimizing.  Still, I suspect a Set would be good, and certainly the code would look simpler if everything is a Set. Otherwise it invites the question, why not this one?)

Also, names.  I was about to suggest making this one start with a lower-case letter, but perhaps these are actually all constants and it'd be better to change all the others!  Hm.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,87 @@
...
+  implicit def validStateCountyTractCombination: Predicate[Geography] = new Predicate[Geography] {
+    override def validate: (Geography) => Boolean = (geo) =>
+      validStateCountyTractCombinationSet.contains((geo.state, geo.county, geo.tract))
+    override def failure: String = "state, county, and census tract combination is not valid"
+  }
+
+  implicit def validCompleteCombination: Predicate[Geography] = new Predicate[Geography] {
+    override def validate: (Geography) => Boolean = (geo) =>
+      validMsaCombinationSet.contains((geo.msa, geo.state, geo.county, geo.tract)) ||
+        validMdCombinationSet.contains((geo.msa, geo.state, geo.county, geo.tract))
```
lgdean: I kinda want to pull that tuple out into a variable. I'm not sure it'd actually read better, though.

For that matter, a lot of this code is really very similar. But I don't currently have a suggestion for what (if anything) to do about that.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
```diff
@@ -0,0 +1,87 @@
...
+    override def failure: String = "state, county, msa, and census tract combination is not valid"
```
lgdean: These failure messages are not all correct. But see #409; it may not matter.

schbetsy: Yep, #409 is merged, so when you pull in changes from master, you can remove the `override def failure` for each `Predicate`.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
kgudel: I looked at both and I think I find it more readable without pulling the tuple out but its pretty ambiguous.

validation/src/main/scala/hmda/validation/dsl/PredicateGeo.scala
kgudel: MSA now always mean metro or micro area.

model/src/main/scala/hmda/model/census/CBSAMetroMicroLookup.scala
```diff
@@ -0,0 +1,56 @@
...
+      val areaWater = values(9)
+      val internalPointLat = values(10)
+      val internalPointLon = values(11)
+
+      CBSAMetroMicro(
+        combinedCode,
+        metroOrMicroCode,
+        geoId,
+        name,
+        legalStatname,
```
schbetsy: Should the 'n' should be capitalized, to make this `legalStatName`? (I'm unsure, since there are lots of complexities of geocoding that I don't understand yet. But using words for these variables instead of acronyms is really helpful!)

schbetsy:
LGTM

PR #404, opened 2016-06-10
Q022

### LINKED ISSUE #182: Validation (2017): Q022 (Date Application Received)
**Edit #:** Q022
**Type:** Quality
**Column:** Date Application Received

Date application received should =activity year or (activity year minus 1 or activity year minus 2).

Additional resources:

nickgrippin:
I'm not sure if I made this approach correctly, please let me know of any improvements you can think of.

Closes #182

### DIFF
--- a/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
@@ -7,8 +7,8 @@ import hmda.validation.rules.lar.quality._
 
 trait LarQualityEngine extends LarCommonEngine with ValidationApi {
 
-  private def q022(lar:LoanApplicationRegister): LarValidation = {
-    Q022(lar, 2017)
+  private def q022(lar: LoanApplicationRegister): LarValidation = {
+    convertResult(lar, Q022(lar, 2017), "Q022")
   }
 
   def checkQuality(lar: LoanApplicationRegister): LarValidation = {
@@ -20,7 +20,6 @@ trait LarQualityEngine extends LarCommonEngine with ValidationApi {
       Q005,
       Q013,
       Q014,
-      q022,
       Q024,
       Q025,
       Q032,
@@ -40,6 +39,8 @@ trait LarQualityEngine extends LarCommonEngine with ValidationApi {
       Q064,
       Q066
     ).map(check(_, lar))
+
+    checks :+ q022(lar)
 
     validateAll(checks, lar)
   }
--- a/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
@@ -22,6 +22,7 @@ trait LarQualityEngine extends LarCommonEngine with ValidationApi {
       Q014,
       Q024,
       Q025,
+      Q027,
       Q032,
       Q035,
       Q036,
@@ -33,11 +34,14 @@ trait LarQualityEngine extends LarCommonEngine with ValidationApi {
       Q045,
       Q046,
       Q049,
+      Q051,
       Q052,
       Q049,
       Q059,
       Q064,
-      Q066
+      Q066,
+      Q067,
+      Q068
     ).map(check(_, lar))
 
     checks :+ q022(lar)
--- a/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
@@ -7,6 +7,10 @@ import hmda.validation.rules.lar.quality._
 
 trait LarQualityEngine extends LarCommonEngine with ValidationApi {
 
+  private def q022(lar:LoanApplicationRegister): LarValidation = {
+    Q022(lar, 2017)
+  }
+
   def checkQuality(lar: LoanApplicationRegister): LarValidation = {
     val checks = List(
       Q001,
@@ -16,6 +20,7 @@ trait LarQualityEngine extends LarCommonEngine with ValidationApi {
       Q005,
       Q013,
       Q014,
+      q022,
       Q024,
       Q025,
       Q032,
--- a/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
+++ b/validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
@@ -8,6 +8,7 @@ import hmda.validation.rules.lar.quality._
 trait LarQualityEngine extends LarCommonEngine with ValidationApi {
 
   private def q022(lar: LoanApplicationRegister): LarValidation = {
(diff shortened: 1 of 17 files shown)


validation/src/main/scala/hmda/validation/rules/lar/quality/Q022.scala
```diff
@@ -0,0 +1,28 @@
...
+
+    (year - applicationYear) is between(0, 2)
+  }
+
+  def apply(lar: LoanApplicationRegister, fYear: Future[Int])(implicit ec: ExecutionContext): Future[Result] = {
+    val applicationYear = Try(lar.loan.applicationDate.substring(0, 4).toInt)
+
+    fYear.map(year => {
+      (year - applicationYear.getOrElse(year)) is between(0, 2)
+    })
```
jmarin: This reads a little weird to me, it looks like `javascript`. Something like this is cleaner:

```scala
fYear.map { year =>
  (year - applicationYear.getOrElse(year)) is between(0, 2)
}
```

nickgrippin: Old habits, sorry 😄

awolfe76: Hope @wpears isn't reading this. 😠

validation/src/main/scala/hmda/validation/engine/lar/quality/LarQualityEngine.scala
```diff
@@ -7,6 +7,11 @@ import hmda.validation.rules.lar.quality._
 
 trait LarQualityEngine extends LarCommonEngine with ValidationApi {
 
+  private def q022(lar: LoanApplicationRegister): LarValidation = {
+    // TODO: 2017 is hardcoded for now.  It will be passed in through the API at a later date
```
lgdean: So, what is the definition of "done" that we're using here? At a first glance, I don't see how we can call it done if it's not hooked up. Is there something blocking us on completing the feature now?

nickgrippin: The activity year is passed in from the front end.  From my understanding (and discussion with @jmarin), we currently have no actual data passing between the front- and back-end.  In order to have this quality edit pass, I have the year hard-coded.  Once we have a working connection, it should be simple to replace this hard-coded value with the value passed in from the front-end.

nickgrippin: Also, check out the discussion on the issue ticket.  It includes most of what I just wrote, but it might have more helpful information as well.

lgdean: Yes, I remember discussing that it was going to come from the front end (though hadn't carefully read the issue itself, so that was a helpful pointer). And I know they're not currently hooked up yet. So perhaps a better way to phrase the question is, what (if anything) blocks us from doing all of the back-end part now?

(I've already asked separately about what's blocking integration of front and back end, which is outside the scope of this issue.  I don't see that question as blocking this one, though.)

validation/src/main/scala/hmda/validation/rules/lar/quality/Q022.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def apply(lar: LoanApplicationRegister, fYear: Future[Int])(implicit ec: ExecutionContext): Future[Result] = {
+    val applicationYear = Try(lar.loan.applicationDate.substring(0, 4).toInt)
+
+    fYear.map { year =>
+      (year - applicationYear.getOrElse(year)) is between(0, 2)
+    }
+  }
+
+  def name = "Q025"
```
schbetsy: `"Q022"`

validation/src/main/scala/hmda/validation/rules/lar/quality/Q022.scala
```diff
@@ -0,0 +1,28 @@
...
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax._
+
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.Try
+
+object Q022 {
+  def apply(lar: LoanApplicationRegister, year: Int): Result = {
+    // This edit should not fail if applicationDate is equal to "NA" or is otherwise non-convertible
+    val applicationYear = Try(lar.loan.applicationDate.substring(0, 4).toInt).getOrElse(year)
```
schbetsy: It looks like the code to extract `applicationYear` is the same for both of the `apply` methods. You could extract that into a method to make this cleaner.

validation/src/main/scala/hmda/validation/rules/lar/quality/Q022.scala
```diff
@@ -11,18 +11,22 @@ import scala.util.Try
 object Q022 {
   def apply(lar: LoanApplicationRegister, year: Int): Result = {
     // This edit should not fail if applicationDate is equal to "NA" or is otherwise non-convertible
-    val applicationYear = Try(lar.loan.applicationDate.substring(0, 4).toInt).getOrElse(year)
+    val applicationYear = parseYear(lar).getOrElse(year)
```
schbetsy: Why is the `getOrElse` in a different place on each method? Not sure whether I'm missing some nuance in their placement, or if they're effectively the same.

nickgrippin: The second method has the year passed in as a `Future`, so we won't know what year to pass to the `getOrElse()` method until the `Future` resolves.

I could call the `parseYear` function in the scope of the map function, but it seemed harder to read that way.

schbetsy:
So, Q022 isn't completed until we're actually using the filing year passed in via the API, correct? Is there a GH issue for completing the second half of this story, so it doesn't get overlooked?

nickgrippin:
Created #410, to create the API endpoint

validation/src/test/scala/hmda/validation/rules/lar/quality/Q022Spec.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.prop.PropertyChecks
+
+class Q022Spec extends PropSpec with PropertyChecks with MustMatchers with LarGenerators with BadValueUtils {
+  // Cases meeting preconditions
+  property("Passes if activity year is within two years after application date") {
+    forAll(larGen, Gen.choose(0, 2)) { (lar, x) =>
+      whenever(lar.loan.applicationDate != "NA") {
+        val applicationYear = lar.loan.applicationDate.substring(0, 4).toInt
+        Q022(lar, applicationYear + x) mustBe a[Success]
```
schbetsy: I'd also like to see tests for the `apply` method that returns a `Future[Result]`. `S013Spec` seems to show a simple strategy for doing that.

nickgrippin: Added

schbetsy:
LGTM

Issue #408, opened 2016-06-14
remove ad-hoc back-end failure messages for edits


lgdean:
As I understand it, the API will provide a standard error message (or messages) for each edit. This message will not vary based on the submitted value or the details of the failure.  Indeed, it may come from a [separate source](https://github.com/cfpb/hmda-platform-ui/issues/96).

(The UI may choose to [display the submitted values](https://github.com/cfpb/hmda-platform-ui/issues/97), but that will be based on the LAR, not the error message.)

I'd already intended to create this issue, and then PR #406 popped up and convinced me to go ahead and push for not spending more time maintaining this code.

PR #406, opened 2016-06-13
fixed language on less than and added language for equal to

kgudel:

PR #409, opened 2016-06-14
remove validation failure message

lgdean:
Addresses #408 

Conflicts with #406, so let's not merge both. (We can cope if it happens, though.)

If this one gets merged, then I may have a bit more cleanup to do.


schbetsy:
LGTM. This change has been necessary for a while.

This has some conflicts with other outstanding branches (particularly #402 ). Whichever merges first will cause conflicts for the other one. Should this be merged first?

lgdean:
Oops, missed this question earlier today. I'd vote to merge this one first, as that one still has outstanding comments to address, but I'm also quite willing to deal with the conflicts if it happens in the other order.

nickgrippin:
LGTM

PR #411, opened 2016-06-15
User institutions - DO NOT MERGE

### LINKED ISSUE #331: API - List of institutions for a user
`/institutions/period/<period>` (period is being used in this case to represent a time period, year or year and quarter)

`GET` - Complete list of institutions a user can/could file for a given year.

[Example](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json)

jmarin:
Closes #331 
Note: The list of institutions returned needs to be filtered further, per user. Since this is still a work in progress (authentication, frontend) I will create another ticket to describe that additional feature, which is not yet implemented here


jmarin:
@awolfe76 @wpears I implemented this as per the ticket, but I noticed that the current codebase in hmda-platform-ui does not filter this by period. Is this intentional? 
Also, as much as I hate to have endless discussions about naming, I think this might require a revisit. This endpoint contains data about institutions and the status of their submissions. It is different from what will surface from [this](https://github.com/cfpb/hmda-platform/pull/381). I wonder if this should be named something like `filings` (I've seen that before, not sure if we intend to use it already), but open to other suggestions.

wpears:
Currently the front-end is pointing to `/api/institutions` because it will render for multiple years. `/api/institutions/periods/2017` is just a backend filter, but, imo, the homepage should wants institutions from all years and then sort/group them appropriately (which it already does). @awolfe76 disagrees here and I believe wants the app to only focus on current year filings.

Something I like about the current approach is that nowhere is the period injected into the app; it is just extracted from the institution data and links are wired up appropriately.

From the backend's perspective, for 2017, the two endpoints should return identical results (since the subset of institutions filed in 2017 will equal the total filed institutions).

awolfe76:
I see that both work for sure.

I do think that when the app initially loads the default should be to the current years filing, that there should be a user choice to go back and re-submit to a prior year.

If I'm understanding correctly, `/api/institutions` would return a list of institutions for a user over all years. The year being defined in the response as part of the institution object as `"period": "2017",`. Would the front-end then have to render only 2017? Or would the backend only return 2017? Or are you suggesting we should all broken down by year?

awolfe76:
Or is it that once we have the full list, we save another API call to the backend to for the case when a user wants to resubmit?

wpears:
Yikes, I just realized #331 was specified outside the bounds of https://github.com/cfpb/hmda-platform-ui/blob/master/api.md and my comment above is incorrect in referring to `/api/institutions/periods/2017`. That is an invalid path, the real path is  `/api/institutions/<someInstitution>/periods/2017`

As can be seen from the api spec, there is no way currently specified to request all institutions for a current year (due to the nature of the hierarchy).

There has been much talk about this before and various solutions (period-first api paths, along with institution-first). How it is handled by the front-end, is all institutions are returned, grouped based on the `period`, then rendered in divisions of Not Started/In progress/Completed.

How this applies here, I think is that the backend should just return all institutions for `/api/institutions` (which is both specified and used by the front-end). It is likely just a change to `ProcessingHttpApi.scala`

jmarin:
I am flagging this as **DO NOT MERGE** for now. I have a few issues with the API structure overall. I will open another ticket with a proposal for discussion. My issue is not necessarily just with the API as it's used by the frontend app, but the lack of a data and persistence model, which needs to be formalized first. The current structure is mixing different concerns, that have their own lifecycle (and thus, different persistence requirements).

hkeeler:
Per @jmarin:

>This endpoint contains data about institutions and the status of their submissions. It is different from what will surface from [this](https://github.com/cfpb/hmda-platform/pull/381). I wonder if this should be named something like `filings` (I've seen that before, not sure if we intend to use it already), but open to other suggestions.

Agreed.  I don't think the current model represents what's going to come out of `/institutions`.  I also like changing the term `periods` to `filings`.  It has more meaning and is overall less confusing.

_If_ we want embed `filings` data within an `institution`, I think it could look something like:

```json
{
    "institutions": [
        {
            "id": "1234567",
            "name": "Bank of Someplace",
            "status": "active",
            "filings": [
                {
                    "id": "2020",
                    "status": "not-started",
                    "required": true
                },
                {
                    "id": "2019",
                    "status": "completed",
                    "required": false
                }
            ]
        },
        {
            "id": "1234568",
            "name": "Superawesomebank",
            "status": "active",
            "filings": [
                {
                    "id": "2020-q4",
                    "status": "completed",
                    "required": true
                },
                {
                    "id": "2020-q3",
                    "status": "completed",
                    "required": true
                }
            ]
        }
    ]
}
```

If we don't want this data embedded, _and_ you wanted to list current/past filings, you'd have to make a separate call to `/institutions/<inst_id>/filings/`, which would return something like:

```json
{
    "filings": [
        {
            "id": "2020",
            "status": "not-started",
            "required": true
        },
        {
            "id": "2019",
            "status": "completed",
            "required": false
        }
    ]
}
```

jmarin:
Yes. Now we are getting somewhere. This structure makes more sense. One question: in the `id` field for `filings` it looks like you put a year, is this meant to be the current status for that filing period? Would we show more than one filing for a specific period of time, or just the latest one?

wpears:
The embedded filings version above is very similar to the current model ([see here](https://github.com/cfpb/hmda-platform-ui/blob/master/server/json/user1-institutions.json)), with the exception of the `currentSubmission` key, which is incidentally the most egregious field w.r.t mixing concerns. When I added this key, I posited in standup that convenience versus strictly HATEOAS is a continuum and we'll probably find ourselves somewhere in the middle after hashing things out.

currentSubmission was included in here to save an extraneous call to `/api/institutions/<id>/periods/<id>/submissions` to get the latest submission to include in the "view filing status" link for routing (submission is done at `cf.gov/hmda/bankid/2017/<submission number>`). I will gladly make the call separately, as I too thought it seemed a bit muddled.

Alternatively, and to Juan's latest point, we could remove the `<submission number>` from the front-end URL and *always* associate user actions with the latest submission. This would disallow users from seeing the results of a previous submission.. but that is probably going to be a rare use case and the data will still be available via the API.

hkeeler:
Per @jmarin:
>One question: in the id field for filings it looks like you put a year, is this meant to be the current status for that filing period? Would we show more than one filing for a specific period of time, or just the latest one?

Good question!  I was sort of thinking latest submission, but I hadn't really followed that through logically.  It _could_ be its own status.  As a user looking at a set of institutions, do you really care about all the many statuses we have for submission?  Probably not.  You really only care about _do I need to do something for this bank_.  It could be a simple as:
* `not-started` - initial state before any submissions have been filed.
* `in-progress` - state as soon as first submission is uploaded, but have yet to get 
* `complete` - state as soon as a given submission has been _signed_.  Also, no new submissions can be posted if `filing` is in this state.
* `revoked` - state when a `complete` filing needs to be refiled.  May want a nicer name...`cancelled`?

hkeeler:
To @wpears point about extra calls to list endpoints needed to get the most recent, we may want to add `/current` or `/latest` that would redirect to the most recent.  This could be useful for both filings and submissions.  Something like:

* `/institutions/<inst_id>/filings/latest   ->   /institutions/<inst_id>/filings/2017`
* `/institutions/<inst_id>/filings/2017/submissions/latest   ->   /institutions/<inst_id>/filings/2017/submissions/3`

hkeeler:
I should really read the other posts more completely.  @wpears is saying the exact same thing on the frontend regarding filing status:

>How it is handled by the front-end, is all institutions are returned, grouped based on the period, then rendered in divisions of Not Started/In progress/Completed.

wpears:
> `complete` - state as soon as a given submission has been signed. Also, no new submissions can be posted if filing is in this state.

I was thinking about this Hans, and I'm wondering what happens if someone realizes that they screwed something up after the fact? Is that an email/call support message? (sorry, this is slightly off-topic)

lesgou:
@wpears I'm thinking that if a resubmission is needed, filers should be able to do so without having to contact hmda help.

hkeeler:
I agree, we should allow them to cancel a `complete` filing.  That's the whole point of the `cancelled`/`revoked` status; a state that a `complete` filing transitions to when either a user finds an issue and wants to re-file _or_ we determine it is invalid or needs to be refiled for whatever reason.

hkeeler:
I think it'd look something like:

image

awolfe76:
Just want to quickly add that changes to the API, end points or response, are fine IMO, even expected. 

When we had these conversations before, one of the key goals of building this out on the front-end was to inform the back-end of the front-end requirements for the API. And that is was pretty likely the front-end would have to make updates. I think we're in a good position and ready for any changes.

awolfe76:
@hkeeler are these the only states required?

> * `not-started` - initial state before any submissions have been filed.
* `in-progress` - state as soon as first submission is uploaded, but have yet to get 
* `complete` - state as soon as a given submission has been _signed_.  Also, no new submissions can be posted if `filing` is in this state.
* `revoked` - state when a `complete` filing needs to be refiled.  May want a nicer name...`cancelled`?

We have several others on the front-end now. Like `parsing`, `parsed`, `validating`, `validation complete`, etc. Sorry if I'm missing something here.

jmarin:
@awolfe76 I think it's another level in the hierarchy.

Institution --> Filings --> Submissions

An institutions has several filings (usually one per year, or per quarter). The status of that filing is as described by @hkeeler above. Basically it's a summary of the filing process as a whole. Within the filing period, there can be one or more submissions, each in one of the 10-12 states that we discussed before. This is how I'm interpreting things, and it makes sense to me

awolfe76:
Thanks @jmarin. That makes sense to me too.

jmarin:
Closing, as the discussion and implementation was done in #422

PR #420, opened 2016-06-17
LAR toDat() method

### LINKED ISSUE #68: Implement toDAT method (Loan Application Register)
https://github.com/cfpb/hmda-platform/blob/master/model/shared/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala#L27

[File spec](http://www.ffiec.gov/hmda/pdf/spec2016.pdf) (pages 10 and 11)

nickgrippin:
Closes #68

### DIFF
--- a/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
+++ b/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
@@ -34,7 +34,7 @@ case class LoanApplicationRegister(
    */
   def toDAT: String = {
     s"$id" +
-      padLeftWithZero(respondentId, 10) + //Parser does not parse leading zeroes
+      padRight(respondentId, 10) + //Parser does not parse leading zeroes
       s"$agencyCode" +
       padRight(loan.id, 25) +
       padRight(loan.applicationDate, 8) +
@@ -64,7 +64,7 @@ case class LoanApplicationRegister(
       padRight(applicant.coRace5, 1) +
       s"${applicant.sex}" +
       s"${applicant.coSex}" +
-      padLeftWithZero(applicant.income, 4) + //Parser does not parse leading zeroes
+      padRight(applicant.income, 4) + //Parser does not parse leading zeroes
       s"$purchaserType" +
       padRight(denial.reason1, 1) +
       padRight(denial.reason2, 1) +
--- a/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
+++ b/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
@@ -34,7 +34,7 @@ case class LoanApplicationRegister(
    */
   def toDAT: String = {
     s"$id" +
-      padRight(respondentId, 10) + //Parser does not parse leading zeroes
+      padRight(respondentId, 10) + 
       s"$agencyCode" +
       padRight(loan.id, 25) +
       padRight(loan.applicationDate, 8) +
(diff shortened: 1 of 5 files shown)


parser/src/test/scala/hmda/parser/fi/lar/LarDatParserSpec.scala
```diff
@@ -1,23 +1,30 @@
...
+import hmda.model.fi.lar._
 import hmda.parser.util.FITestData
-import org.scalatest.{ FlatSpec, MustMatchers }
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{ MustMatchers, PropSpec }
 
 import scala.concurrent.{ ExecutionContext, Future }
 
-class LarDatParserSpec extends FlatSpec with MustMatchers {
+class LarDatParserSpec extends PropSpec with MustMatchers with PropertyChecks with LarGenerators {
```
kgudel: Nice change to match the rest of the way we've been doing things!

model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
```diff
@@ -29,9 +29,65 @@ case class LoanApplicationRegister(
...
+      s"${applicant.coSex}" +
+      padRight(applicant.income, 4) +
+      s"$purchaserType" +
+      padRight(denial.reason1, 1) +
+      padRight(denial.reason2, 1) +
+      padRight(denial.reason3, 1) +
+      padNumOrNa(rateSpread, 5) +
+      s"$hoepaStatus" +
+      s"$lienStatus"
+  }
```
kgudel: There should be a filler at the end of 270 characters.

model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
```diff
@@ -29,9 +29,65 @@ case class LoanApplicationRegister(
...
   }
 
+  /**
+   * NOTE:  The DAT file format is not supported by CFPB
+   */
   def toDAT: String = {
-    //TODO: implement DAT output
-    ""
+    s"$id" +
+      padRight(respondentId, 10) +
```
kgudel: This field should be right justified and zero filled.

model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
```diff
@@ -29,9 +29,65 @@ case class LoanApplicationRegister(
...
+      padRight(applicant.race4, 1) +
+      padRight(applicant.race5, 1) +
+      s"${applicant.coRace1}" +
+      padRight(applicant.coRace2, 1) +
+      padRight(applicant.coRace3, 1) +
+      padRight(applicant.coRace4, 1) +
+      padRight(applicant.coRace5, 1) +
+      s"${applicant.sex}" +
+      s"${applicant.coSex}" +
+      padRight(applicant.income, 4) +
```
kgudel: Should be `padNumOrNa`

model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
nickgrippin: So, the issue with both this field and the income field is that the generated LARs are never zero-filled.  So converting the generated LAR to a DAT output will never equal the original LAR (a generated income of 456 will be converted to 0456 in the DAT output.  I suppose modifying the tests would be the best way to address this problem, but it will unfortunately make the tests more complicated than a blanket "converted LARs must equal generated LARs".  Another possible solution would be to make sure that we always store both respondent ID and income with leading zeroes.

PR #421, opened 2016-06-20
Begin Testing LAR Validity Engine

schbetsy:
This is a proof of concept for a way to test the `LarValidityEngine` and similar engines. I'm looking for feedback on it. I'm at a conference, so I'll be slow to respond to comments, but I will try to do so within a day or two.

Addresses nine of the check boxes in #276 .

### DIFF
--- a/validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
+++ b/validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
@@ -54,6 +54,23 @@ class LarValidityEngineSpec extends WordSpec with MustMatchers with LarValidityE
       val lar = validLar1.copy(loan = validLoan.copy(amount = -2))
       errors(lar) must contain("V250")
     }
+
+    "fail for V255" in {
+      val lar = validLar1.copy(actionTakenType = 9)
+      errors(lar) must contain("V255")
+    }
+
+    "fail for V260" in {
+      val denial = Denial("1", "", "")
+      val lar = validLar1.copy(actionTakenType = 1, denial = denial)
+      errors(lar) must contain("V260")
+    }
+
+    "fail for V262" in {
+      val loan = validLoan.copy(applicationDate = "NA")
+      val lar = validLar1.copy(actionTakenType = 1, loan = loan)
+      errors(lar) must contain("V262")
+    }
   }
 
   def errors(lar: LoanApplicationRegister): List[String] = {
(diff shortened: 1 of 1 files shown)


nickgrippin:
I'm struggling to see the value from individually testing each validation for a failure.  From the point of testing the validation engine, we want to make sure that it can return both a `Success` and a List of `Failures`.  To that end, we should only need one additional (smallish) test for the `Failure` case.  Testing each individual element in the List for a failure doesn't seem to add any value to our overall test coverage, as we already have extensive tests for single validations.

kgudel:
+1 to the comment by @nickgrippin

schbetsy:
It's important to test that every edit gets run. With the current state of the tests, even if we removed edits from the [list of edits](https://github.com/cfpb/hmda-platform/blob/master/validation/src/main/scala/hmda/validation/engine/lar/validity/LarValidityEngine.scala#L11) in the LAR Validity Engine, all of the tests would still pass. The only way we know that we are actually executing a validity edit on the LARs is to see it in the list of failures.

nickgrippin:
@schbetsy:  I agree that it is important to see that every edit is run.  However, the current testing approach seems to suggest we don't believe that the `map` function will run for every element in a list.  At that point, we're no longer testing our own code, but built-in Scala functionality.  We should only need to test that any individual edit can fail (and be returned), at which point we know the entire list is being passed correctly to the `validateAll` method.

Again, code coverage is crucial to any large project.  But the tests should only cover new behavior introduced.  We can always safely assume that the language we're using (and their functions, like the `map` method) will work as intended.

schbetsy:
I disagree. Your strategy does not account for human error: if someone accidentally leaves an edit out of the list, or accidentally deletes one while fixing a merge conflict, our current test suite will not catch that.

A good test suite acts as documentation of the requirements of the code. If we don't test that every edit is run, we can't prove that our code is meeting the requirements of the regulation.

High code coverage _percentage_ is not a guarantee that your code base has correct behavior. Tools like CodeCov can tell whether we run every line of code, but they can't tell whether we're actually testing all of the functionality.

lgdean:
I definitely think we need to be testing this, one way or another.  Whether in integration-test style here, or as in something more like #427, I haven't decided. (And haven't yet read the code carefully enough to decide. I will say that, on first glance, I do not favor #427 in its current state, because it makes a private thing public only for the purposes of testing. If on the other hand we were to refactor so that there's a coherent split between code and configuration, and could test each part, then I might favor that approach. But I can go into that more on that PR, once I've read it more carefully.)

schbetsy:
Per our discussion today, I'm closing this PR. We will find a new strategy for ensuring that all edits are run, probably at the integration test level, once the API is settled and more endpoints are written.

PR #422, opened 2016-06-20
Data model and HTTP API

jmarin:
Based on discussion from #411, also related to #381  
Initial implementation, could probably use some refactoring. Includes test and dummy data to see what the API looks like. 

Includes persistence based on event sourcing for institution, filing and submissions. Actor hierarchy for these is flat at the moment, this will likely change once other aspects are added to the project (i.e. edit processing).

Documentation with sample responses can be seen [here](https://github.com/jmarin/hmda-platform/blob/data_model/Documents/API.md).


awolfe76:
_This is to attempt to walk through the steps to render the **home page** of the UI. It's helpful to me to think of it this way ... and hopefully to others. If my steps are wrong, let me know._

Based on [API.md](https://github.com/jmarin/hmda-platform/blob/ccbc159c49374cab97772aaf70604f83728fe3dd/Documents/API.md) to render the home page for the front-end:

- call `GET /institutions` to return the list of institutions for the user
- using the institutions list, filter out the list of institutions required by `period` (filtering is done by the front-end and period is selected in the UI, think a drop down list of available periods)
- based on this filtered list call `GET /institutions/<institution>/filings/<period>` for each institution, again using the correct period selected in the UI, to get the institutions filing status for that period
  - this will be used to "group" the institutions based on their filing status for the period
    - not started
    - in progress (created)
    - completed
- also based on the filtered list of institutions call `GET /institutions/<institution>/filings/<period>/submissions` for the list of submissions for each institution in the current filing period
  - this will be used to render:
    - links to download any data for all submissions other than a current, "in progress", submission for the given filing period; this includes previous submissions
      - current submission being the highest `id`
    - information about the state of the current submission to allow the filer to quickly see the status of the submission (one of the states outlined in the [submission-states.md file](https://github.com/cfpb/hmda-platform-ui/blob/master/submission-states.md))

Those steps _should_ get us the current home page in the UI.

Like the `GET /institutions/<institution>/summary` endpoint, where the filings are added to the institution, would it be helpful to the UI to have `GET /institutions/<institution>/filings/summary` and `GET /institutions/<institution>/filings/<period>/summary` endpoints and have the submissions added to the filing? In the case of `GET /institutions/<institution>/filings/summary`, returning something like:
```json
{
  "id": "12345",
  "name": "First Bank",
  "filings": [
    {
      "period": "2017",
      "fid": "12345",
      "status": "not-started",
      "submissions": [
        {  
          "id": 3,
          "submissionStatus": "created"
        },
        {
          "id": 2,
          "submissionStatus": "created"
        },
        {
          "id": 1,
          "submissionStatus": "created"
        }
      ]
    },
    {
      "period": "2016",
      "fid": "12345",
      "status": "completed",
      "submissions": [
        {
          "id": 2,
          "submissionStatus": "created"
        },
        {
          "id": 1,
          "submissionStatus": "cancelled"
        }
      ]
    }
  ]
}
```

This would save extra API calls from the front-end to get the submission data after getting the filing status.

hkeeler:
@awolfe76, thanks for laying all of that out from the frontend perspective.  That's really helpful.  I actually think we can come up with a happy medium that'll keep the backend sane, and reduce the number or overall API calls.  He's my take:

1. _call GET /institutions to return the list of institutions for the user_ 👍 

    ```json
    {
      "institutions": [
        {
          "id": "12345",
          "name": "First Bank",
          "status": "active",
          "filings": [
             {"period": "2016", "status": "completed"},
             {"period": "2017", "status": "not-started"}
          ]
        },
        {
          "id": "123456",
          "name": "Second Bank",
          "status": "inactive",
          "filings": [
             {"period": "2016", "status": "completed"}
          ]
        }
      ]
    }    
    ```

    A few things to note here:

    1. I've removed the `fid` from `filing`.  We may well need an identifier in whatever data store we use internally, but I don't think it provides much for us in the API.  `institution.id` + `filing.period` are a composite unique identifier for a `filing`.  We're already using these two attributes in the URI to uniquely identify a `filing` (`/institutions/<institution_id>/filings/<period>/`).
    2. I didn't embed `submissions` in each `filing`.  Do we _really_ need submission data on the home page?  It seems like the majority of the time once a user start actively filing, they're going to be spending all of their time filing page, and they'd interact with submissions there.

1. _using the institutions list, filter out the list of institutions required by period (filtering is done by the front-end and period is selected in the UI, think a drop down list of available periods)_ 👍 

1. Group filings by `status`.  Notice, this does not require a call per institution/period since the data is already present in the list of institutions.

At this point, you have everything you need to build a pretty useful homepage, and it was just a single API call.  _If_ we really need submission info on the homepage, I think we should embed that data in `institution.filings` as well.  There are performance considerations here (having to join `institutions -> filings -> submissions`), but in most cases its going to be scoped to a small set of institutions (usually just one) for a given user, so probably not too expensive.

hkeeler:
Oh yeah, I'm curious about everyone's thoughts on what I proposed at standup today for `filings`, adding data about what dates are associated a given `filings`.

```javascript
{
  "period": "2017",
  "status": "not-started",
  "from": "2017-01-01",
  "to": "2017-12-31",
  "due": "2018-03-30" 
}
```

The `from` and `to` attributes don't have a ton of value for the first few filing periods as we know they'll always be for the full year.  They become more useful in the future when we support quarterly filings.

However, the due date seems pretty useful now.  It provides both a reminder to the user how much time they have left, and something we can trigger own events off of.

And if you'd rather I move this off to it's own issue, that's cool too.

awolfe76:
👍  to getting everything at once. I think maybe that call could be something like `/institutions/details`. That way we don't break what @jmarin is recommending because I think that all makes sense too. We shouldn't overwrite `/institutions` with the needs of the front-end.

I use `/details` here, I think that others should be `/details` as well. We're actually providing _more_ information about each of those rather than a summary. The summary seems to be the default. So it would be:
- `GET /institutions/<institution>/detail` - to wrap up filings and submissions within an institution
- `GET GET /institutions/<institution>/filings/detail` - to wrap up submissions within a filing
- etc

In both cases you would have to the ability to add a `period`.

> I've removed the fid from filing.

Sounds ok to me, but it doesn't hurt to have it right?

> I didn't embed submissions in each filing. 

I think we do need them on the home page. You're right, most of the interaction would be on the _current_ submission. The goal for pulling in the other previous submissions is simply to allow them to download them (probably as a csv). I think adding that functionality to the submissions screens would just clutter that, already somewhat cluttered, process. Something like:

editreports

From the home page they can quickly download the current submission csv and any previous submission csv. These would allow them to do analysis to fix any edits.

awolfe76:
@hkeeler I think maybe the dates would be good in another issue.

hkeeler:
I don't really get the point of `GET /institutions/details`.  It sounds like it is the same as `GET /institutions`, but with more..._details_?  Why do we need that?  For performance reasons?  That seems like premature optimization.  I'd rather maintain a single resource that lists institutions, and include enough detail to make it generally useful, and only come up with specialized endpoints once we start to hit up against performance issues.

Also, I now get why we need submission data on the homepage.  There's nowhere else to currently put it.  :)  In my mind there was an intermediate _filing_ page, equivalent to the `/institutions/<institution_id>/filings/<period>` API resource.  That's not the case, though, right?  There's currently just homepage -> submission.  I agree, cramming past submissions onto the current submission screen is a bad idea.  However, I think a filing page may be useful.  If we had that, we could move the submission history stuff there since most users are going to care about that on the homepage.

And as for the `filing` date fields, @lesgou let me know that it's _not quite so simple_, and we'll need to talk through that one before we go further.

awolfe76:
I was thinking more along the lines of `GET /institutions/details` being an API call for front-end purposes, which I guess is an optimization (1 call instead of 2 or 3), while trying to maintain @jmarin's idea of `/institutions` being simply the institutions.

Right, there is no "filing" page to list the submission for a filing. Like you say, right now that's all on the home page. We could have a "View details of your filing" link to get to a filings submissions but I would imagine that page being very bare in most case because it's filtered down by institution and period. At the moment I don't think that justifies a separate page. There just wouldn't be a lot there. I kind of think the filers would want quick access to past submission data for the filing period. Making it available on the home page would let them quickly download the csv(s) to be able to analyze and being to fix edits.

But for the API, of course it would be great to get all of that at once, but if a few API calls are required we would be ok. In the end we can work with whatever makes sense for both back-end and front-end.

jmarin:
@awolfe76 I've updated the API slightly with more details on some endpoints, please see updated [documentation](https://github.com/jmarin/hmda-platform/blob/data_model/Documents/API.md) and let me know what you think. I agree that adding an endpoint that merges institutions/filings/submissions in one call is desirable, but at this stage I would like to wait to implement it. This will require aggregation, which really calls for a query database technology choice, and I don't think we're there yet. I will explore other options though.

Documents/API.md
```diff
@@ -0,0 +1,135 @@
...
+      ]
+    }
+    ```
+
+* `/institutions/<institution>`
+    * `GET` - Details for Financial Institution
+    
+    Example response, with HTTP code 200: 
+    
+    ```json
```
awolfe76: I _think_ you meant like this:
```json
{
  "institution": {
  "id": "12345",
  "name": "First Bank",
  "status": "active",
  "filings": [
    {
      "id": "2017",
      "fid": "12345",
      "status": "not-started"
    },
    {
      "id": "2016",
      "fid": "12345",
      "status": "completed"
    }
  ]
} 
```
To have the `filings:[]` in the outer object, to match the response in `/institutions/<institution>/summary`.

Documents/API.md
```diff
@@ -0,0 +1,135 @@
...
+    } 
+    ```
+    
+    
+* `/institutions/<institution>/filings/<period>`
+    * `GET` - Details for a filing
+    
+    Example response, with HTTP code 200:
+    
+    ```json
```
awolfe76: Same as above, I _think_ you meant to have the `submisstions: []` in the outer object:
```json
{
  "filing": {
  "id": "2017",
  "fid": "12345",
  "status": "not-started",
  "submissions": [
    {
      "id": 3,
      "submissionStatus": "created"
    },
    {
      "id": 2,
      "submissionStatus": "created"
    },
    {
      "id": 1,
      "submissionStatus": "created"
    }
  ]   
}
```

awolfe76:
Thanks @jmarin. A couple of comments on the format of the responses, but this would simplify things.

For now we can get what we need for the front-end using 2 API calls.
- `GET /institutions` for the list
  - filter this on the front-end for the period needed
- `GET /institutions/<institution>/filings/<period>`
  - lets us group institutions (not-started, created, completed), __as well as list their submissions__

To help get the integration going we can implement it this way and re-visit when we get to a point where we can aggregate. So no rush there, this should get us started.

Documents/API.md
wpears: I think this was written as intended, so the schema of an `institution` object doesn't change whether returned from `/institutions` or `/institutions/<institution>`

Documents/API.md
wpears: Same as above re: schema

wpears:
With the current status of `API.md` the front-end flow would look like this:

 - Make a request to `/institutions`, get the set of the institutions associated with that user

 - *For each* institution, make a request to `/institutions/<institution>` to get filing info

- Either filter to currently selected filing period, or reverse sort the institutions for display (newest first). Because the filing info includes not-started/in-progress/completed we can provide actionable info without over-stuffing the homepage with data. Note also that if there was a filing period selection, we could request to `/institutions/<institution>/filings/<period>`. This would prevent the amount of data that is needlessly requested and filtered away from growing linearly with the number of filings.

- The user clicks a "Begin filing" or "View filing status" link and gets directed to the submission page, which grabs the current submission (out of scope for this PR) and renders the view appropriately.

--

I'm going to bring up again that API design is a continuum, from many, completely atomic requests (with more overhead) to a single request that dumps all subresources into a single large response.

For our case, we definitely want to lean as close as possible to atomic requests as this allows the backend to inform the front-end of state changes without needing to resend a slew of redundant data. In general, this more HATEOAS-ish approach allows backend services to be somewhat more decoupled which makes sense with our architecture.

Of course, the penalty is the HTTP overhead associated with many, small requests and associated problems (network waterfall due to max simultaneous connections per domain and not being able to request likely needed data until requests higher in the hierarchy return). @hkeeler is right that worrying about this now is probably a premature optimization, especially since the front-end won't care if we added more data to higher level responses if it provided worthwhile speedups.

To close with an aside, HTTP/2 is quite helpful re: reducing the overhead/latency of HATEOAS-style APIs  (server push, header compression, multiplexing, etc.). Just sayin' 😜

awolfe76:
>  Note also that if there was a filing period selection, we could request to /institutions/<institution>/filings/<period>.

Would we ever _not_ have a filing period?

debseidner:
@awolfe76 I think you would always have a filing period. I can't think of a time you wouldn't.

Documents/API.md
jmarin: @awolfe76 @wpears Just to make sure we're on the same page, which format would you guys prefer? I haven't changed it yet to what Andrew was proposing but that is a trivial change. Wyatt's comment is spot on, I was reusing an internal object and thus the resulting schema is the same for institution (or filing). But including it in the outer object is also fine with me.

awolfe76: I think what you had originally is good ... ignoring my comment.

model/src/main/scala/hmda/model/fi/Filing.scala
```diff
@@ -0,0 +1,9 @@
+package hmda.model.fi
+
+sealed trait FilingStatus
+case object NotStarted extends FilingStatus
+case object InProgress extends FilingStatus
+case object Completed extends FilingStatus
+case object Cancelled extends FilingStatus
+
+case class Filing(id: String = "", fid: String = "", status: FilingStatus = NotStarted)
```
kgudel: Is the `id` the filing period (it seems to be from the institutions api)? If so it might be good to make that explicate in the variable name here and in the filing persistence actor.

lgdean: Agreed.  I'd even suggest taking it a step further, and instead of a String, have a specific type for the filing period, so it has to be explicit everywhere.

api/src/main/scala/hmda/api/protocol/processing/SubmissionProtocol.scala
```diff
@@ -0,0 +1,54 @@
...
+        case Parsing => JsString("parsing")
+        case Parsed => JsString("parsed")
+        case ValidatingSyntaxAndValidity => JsString("validating syntactical and validity")
+        case ValidatedSyntaxAndValidity => JsString("validated syntactical and validity")
+        case ValidatingQualityAndMacro => JsString("validating quality and macro")
+        case Unverified => JsString("unverified")
+        case Verified => JsString("verified")
+        case Signed => JsString("signed")
+        case Failed(msg) => JsString(s"failed: $msg")
+
```
lgdean: I was going to suggest replacing all these cases with:
```
        case Failed(msg) => JsString(s"failed: $msg")
        case otherStatus => JsString(otherStatus.message)
```
... but then I noticed the various `SubmissionStatus.message` values don't actually match these strings.  Which in turn got me wondering: what are those messages for, then?  They are currently unused.

api/src/main/scala/hmda/api/protocol/processing/SubmissionProtocol.scala
```diff
@@ -0,0 +1,54 @@
...
+          case "validating syntactical and validity" => ValidatingSyntaxAndValidity
+          case "validated syntactical and validity" => ValidatedSyntaxAndValidity
+          case "validating quality and macro" => ValidatingQualityAndMacro
+          case "unverified" => Unverified
+          case "verified" => Verified
+          case "signed" => Signed
+          case "failed" => Failed("")
+          case _ => throw new DeserializationException("Submission Status expected")
+        }
+        case _ => throw new DeserializationException("Submission Status expected")
```
lgdean: Future maintainers may have an easier time if these two exception messages are different.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -0,0 +1,198 @@
...
+            complete(HttpResponse(StatusCodes.InternalServerError))
+        }
+      }
+    }
+
+  val institutionByIdPath =
+    path("institutions" / Segment) { fid =>
+      extractExecutionContext { executor =>
+        val institutionsActor = system.actorSelection("/user/institutions")
+        val filingsActor = system.actorOf(FilingPersistence.props(fid))
```
kgudel: As a reminder from when we talked yesterday, remember to shut down this actor after the call finishes.

jmarin: Thanks @kgudel I've pushed a commit fixing this

Documents/API.md
```diff
@@ -0,0 +1,139 @@
...
+          "id": "123456",
+          "name": "Second Bank",
+          "status": "inactive"
+        }
+      ]
+    }
+    ```
+
+* `/institutions/<institution>`
+    * `GET` - Details for Financial Institution
```
schbetsy: What should this endpoint return for an invalid institution id? Currently it returns this, which seems misleading:
```bash
# GET /institutions/bogus
{
  "institution": {
    "id": "",
    "name": "",
    "status": "inactive"
  },
  "filings": []
}
```

The same question goes for any other endpoint that takes parameters, I guess.

wpears: Seems like it would be worth 404ing on bogus institutions; though imo this should be ticketed and dealt with in a different PR

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+
+  val submissionPath =
+    path("institutions" / Segment / "filings" / Segment / "submissions") { (fid, period) =>
+      val submissionsActor = system.actorOf(SubmissionPersistence.props(fid, period))
+      post {
+        val submissionsActor = system.actorOf(SubmissionPersistence.props(fid, period))
+        submissionsActor ! CreateSubmission
+        val fLatest = (submissionsActor ? GetLatestSubmission).mapTo[Submission]
```
kgudel: If there are multiple users trying to create submissions are we guaranteed that the submission that was just created will be the latest? Say if there are two users trying to create a submission for the institution, is it possible that 1 will send `CreateSubmission` then 2 will send `CreateSubmission` then 1 will send `GetLatestSubmission`. If this happened then 1's `GetLatestSubmission` would return the submission created by 2 not by 1. I will admit that this is a small corner case but it might still be good to address.

Documents/API.md
```diff
@@ -0,0 +1,139 @@
...
+    
+    ```json
+    {
+        "id": 4,
+        "submissionStatus": "created"
+    }
+    ```
+    
+* `/institutions/<institution>/filings/<period>/submissions/<submissionId>`
+    * `POST` - Upload HMDA data to submission
```
schbetsy: This endpoint allows me to create a submission for a nonexistent institution. Not a task for this PR, but we need to a) decide what its behavior should be, and b) implement it.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -0,0 +1,200 @@
...
+        onComplete(fLatest) {
+          case Success(submission) =>
```
kgudel: I believe that the purpose of checking the success of `GetLatestSubmission` is to confirm that the submission was created. However, it looks like this will be true if there was ever a submission. Not only if a new one was created. For example, if Kenneth's Bank created a submission and completed it but then later had to refile and so attempted to create a new submission. Even if the creation of this filing failed I believe that `GetLatestSubmission` would still return a success as it contains the prior submission already.

kgudel: It looks as well like even if there was no original submission `GetLatestSubmission` would still return an empty submission causing it to hit the success case.

Documents/API.md
hkeeler: Agreed.  Should be a `404`.

Also, do we have a standard error response format for the API yet?  It's not really necessary for something simple like a `404`, but we'll need it for more complex fails.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
kgudel: We might need to do something like check the id of the submission returned by `GetLatestSubmission` to make sure that it is as expected which might require something like assigning the id of the submission from `CreateSubmission` as the date/time or if we want to keep using the incrementing method we might want to query want the current is and checking that the new is 1 above it. If we knew the id in the `submissionPath` it might be cleaner to query for the submission of that id rather than looking at the latest one.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
wpears: As far as creating new submissions goes, the user *always* wants the latest submission returned as the result of the POST (since only the latest submission is valid for uploading/verifying quality/justifying macro), which I think `GetLatestSubmission` solves here. Shouldn't matter to the user that their internally created submission was immediately invalidated as long as they get the newest one back as a result of the POST

Documents/API.md
wpears: Nope, no error response yet discussed. Definitely makes sense to do this soon as the front-end can start working on displaying errors where appropriate. Created #433

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
kgudel: Ok, that is interesting thanks Wyatt.

Documents/API.md
```diff
@@ -0,0 +1,139 @@
...
+     },
+     {
+       "id": 1,
+       "submissionStatus": "created"
+     }
+   ]
+   } 
+    ```
+    
+* `/institutions/<institution>/filings/<period>/submissions`
```
hkeeler: Do we want this to be a two-step process:

1. Create submission
2. Upload file

It seems like we could just as well have the file upload POST go directly to `.../submissions`, and have that create the new submission record as well.

How is the UI going to work?  Is the end user going to first create the submission, then upload, or just straight to upload?

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
kgudel: Does that imply that if a new submission is created for some fi and filing period then any current not completed submission for that same fi and filing period should be canceled?

Documents/API.md
hkeeler: Agreed.  If any parent attribute in the URL hierarchy (`institution_id` and `period` here) is not present, it should be a fail.  `404` still seems appropriate.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
hkeeler: @kgudel, good point.  I actually think the attempted _new_ submission should fail if there's an existing open submission.  I'd prefer an explicit `cancel` for any open submission that's not going to be completed.

wpears: @kgudel That was my assumption; that a request to create a new submission will `cancel` the current submission. In a two-simultaneous-filers scenario, this would result in the person with the `cancel`'d submission getting an error message (400 or 409 if we want to be fancy) when they try to POST something to the API that they are attempting to update an canceled submission with a link to the latest submission.

@hkeeler Are there scenarios where the explicit cancel wins out? eg, accidental double-clicks on "Refile" that aren't throttled/debounced wouldn't cause any trouble for the user, but could create some garbage/empty submissions in the backend. From a UX perspective I think just being able to click a single button is the experience we'd want but that could wrap two requests if an explicit cancel is needed before the new submission is created.

kgudel: I tend to agree with @hkeeler that at least in the backend any cancel should be explicit.

Documents/API.md
wpears: Currently, the homepage has a "Begin Filing" link which is what triggers the initial submission creation. The user is then presented with the upload form. 

I think our management of the  submission states should get in the user's way as little as possible. This is also why I was assuming auto-canceling of former submissions when the user wants to resubmit ([discussed above](https://github.com/cfpb/hmda-platform/pull/422#discussion_r69814490).

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
wpears: It's interesting, how this is implemented in the front-end mock server (which is somewhat mirrored w/ `GetLatestSubmission`) is that only the "current" submission is considered valid for API activity that includes a submission. Older submissions aren't explicitly marked as "canceled", they just aren't the latest and so are assumed de facto canceled. However, the front-end mock server modeled data a bit differently and didn't have a `submissionStatus` flag, here, we'd certainly need to adjust the `submissionStatus`, but I don't think a separate client request is necessary.

Then again, a separate client request would, in a simultaneous filers scenario, protect the in-progress work of the first user from the second-user's POSTing of a new submission... So I suppose I'm on the fence.

Documents/API.md
hkeeler: It is an interesting UX question.  Which is more annoying: 

1. Having to go cancel a submission before you can upload a new one.
1. Accidentally overwriting a submission you weren't actually done with yet.

In either case, I don't think we're required to have the two-step process.  The "Begin Filing" page doesn't have to write to the API before displaying the page, right?  That page could wait to do the POST until the upload.  This has the benefit of not having written anything if the user decides to cancel out of that page.

persistence/src/main/scala/hmda/persistence/FilingPersistence.scala
```diff
@@ -0,0 +1,89 @@
...
+    case CreateFiling(f) =>
+      if (!state.filings.contains(f)) {
+        persist(FilingCreated(f)) { e =>
+          log.debug(s"Persisted: $f")
+          updateState(e)
+        }
+      }
+
+    case UpdateFilingStatus(modified) =>
+      if (state.filings.map(x => x.id).contains(modified.id)) {
```
schbetsy: This pattern is common to all of the persistence actors. Should this `if` (and others like it) have an `else`? 

If I understand correctly, this line is preventing the `UpdateFilingStatus` method from trying to update a nonexistent record. If this case does receive an incorrect `id`, it seems like this will fail silently. Would it be better to log that error, for debugging purposes?

Documents/API.md
wpears: Yeah, the front-end would need some adjustment as it routes based on current submission number to allow the review of past submissions, but it's not too much work if we'd prefer submissions to start precisely at the time of upload.

awolfe76:
Lots of good conversation going here. When we had talked before there was a thought of just being able to get the home page working with a list of institutions, as a quick way to show the integration between front and back ends. I know that PR has more than that but if possible it would be good to be able to get the `/institutions` up and running.

Maybe we can focus on that for now and revisit the others with new issues and prs?

_If the conversation around the other endpoints isn't slowing this down or if its required please ignore this. Just wanted to float the idea again._

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -0,0 +1,200 @@
...
+            complete(ToResponseMarshallable(Institutions(institutions)))
+          case Failure(error) =>
+            log.error(error.getLocalizedMessage)
+            complete(HttpResponse(StatusCodes.InternalServerError))
+        }
+      }
+    }
+
+  val institutionByIdPath =
+    path("institutions" / Segment) { fid =>
```
hkeeler: Can we change `fid` to `institutionId` throughout the codebase?  I keep thinking it is a `filingId`...but it's not, right?

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
jmarin: This discussion veered off from the PR code review a bit. I think it's worthwhile to discuss this, but maybe as part of another ticket. This PR already has a lot of opinions and back and forth, and it needs some closure. @kgudel I'm not sure I follow your logic completely, but if you can create a test case that showcases this implementation failing according to the corner case you describe, that would be extremely useful. One thing to keep in mind is that we are assuming eventual consistency and "last write wins". Perhaps this answers your question.

Documents/API.md
jmarin: The way this is implemented, as a two step process, is based on @wpears description of how the UI works. It can be changed of course, but it is implemented here this way on purpose.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
jmarin: Before `GetLatestSubmission` gets invoked, a submission is created as part of this `POST`, in this line:

```scala
  submissionsActor ! CreateSubmission
```

kgudel: My thought though was that the purpose of calling `GetLatestSubmission` was to confirm that `CreateSubmission` was successful. Is that correct?

jmarin: Yes and no. It's there to return the created submission, as per the UI requirements. In that sense, yes, it's a "confirmation"

Documents/API.md
jmarin: I've added 404 responses to all endpoints. Quickly, so I might have missed something in there. In that case, I would create a ticket and separate that from this PR

Documents/API.md
jmarin: I've added this (returns 404)

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
kgudel: Ok, my comment was based on the assumption that we wanted to check whether, in the backend, the `CreateSubmission` message was properly received and processed, if this is not the case then you can ignore this thread.

jmarin: You _can_ check that a message was received by an actor, for a particular run of a test. That is a very poor guarantee though, in real life messages can get lost, given that these are "fire and forget", asynchronous messages. This initial implementation doesn't address guaranteed delivery. Akka Persistence has "at least once delivery" semantics, but it is not a perfect solution either

kgudel: Should guaranteeing delivery be moved into an issue for later then?

api/src/main/scala/hmda/api/protocol/processing/SubmissionProtocol.scala
```diff
@@ -0,0 +1,43 @@
...
+          case "validating syntactical and validity" => ValidatingSyntaxAndValidity
+          case "validated syntactical and validity" => ValidatedSyntaxAndValidity
+          case "validating quality and macro" => ValidatingQualityAndMacro
+          case "unverified" => Unverified
+          case "verified" => Verified
+          case "signed" => Signed
+          case "failed" => Failed("")
+          case _ => throw new DeserializationException("Submission Status expected")
+        }
+        case _ => throw new DeserializationException("Unable to deserialize")
```
lgdean: Nice resolutions in this file, thanks @jmarin!

Issue #423, opened 2016-06-20
HmdaFileUploadSpec fails without a `clean` before `test`


hkeeler:
The following error occurs when running `sbt test` without first doing an `sbt clean`.  @nickgrippin mentioned that the Akka Persistence data isn't being cleared out post-test.  This is likely the culprit.

```
[info] HmdaFileUploadSpec:
[info] A HMDA File
[info] - must be persisted *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: expected HmdaFileUploadState(Map(1466450321914 -> 4)), found HmdaFileUploadState(Map(1466450112373 -> 4, 1466450321914 -> 4))
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:350)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:326)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:734)
[info]   at hmda.api.processing.HmdaFileUploadSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(HmdaFileUploadSpec.scala:26)
[info]   at hmda.api.processing.HmdaFileUploadSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(HmdaFileUploadSpec.scala:21)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   ...
[info] - must recover with event *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: expected HmdaFileUploadState(Map(1466450321914 -> 4)), found HmdaFileUploadState(Map(1466450112373 -> 4))
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:350)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:326)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:734)
[...]

PR #425, opened 2016-06-21
Fix issue where clean has to be run between tests

kgudel:
This is done by deleting the snapshot store after tests are run and is based on the work of @nickgrippin and info found [here](http://doc.akka.io/docs/akka/current/scala/persistence.html#Plugin_TCK).


nickgrippin:
Closes #423

nickgrippin:
Tested locally and was able to run successive tests without any failures.  @lgdean, @hkeeler, can you check as well?

Other than that, LGTM

lgdean:
Works on my machine. :)

PR #427, opened 2016-06-27
LarQualityEngine Test

kgudel:
This is another idea for how we might test the various rule engines.

This test is meant solely to test whether all of the edits, and no extras, are being run by QualityEngine.

Instead of checking for edits failing it checks that the list of edits in the engine appears to be correct. It first checks that there are no duplicates in the list. It then checks that the list is the correct size. To check what the size of the list should be the edit counts the number of files in `validation/src/main/scala/hmda/validation/rules/lar/quality` and subtracts 1 because Q022 isn't in the list due to its slightly different behavior.

This pr also removes a duplicate Q049 in the list of edits.

### DIFF
--- a/validation/src/test/scala/hmda/validation/engine/lar/quality/LarQualityEngineSpec.scala
+++ b/validation/src/test/scala/hmda/validation/engine/lar/quality/LarQualityEngineSpec.scala
@@ -27,7 +27,7 @@ class LarQualityEngineSpec
   }
 
   property("There should be no duplicates in qualityEditsList") {
-    qualityEditsList.toSet.toList.length mustBe qualityEditsList.length
+    qualityEditsList.distinct.length mustBe qualityEditsList.length
   }
 
   property("There should be edits equal to the number of edit files - 1 (for Q022)") {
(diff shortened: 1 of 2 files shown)


validation/src/test/scala/hmda/validation/engine/lar/quality/LarQualityEngineSpec.scala
```diff
@@ -25,4 +25,17 @@ class LarQualityEngineSpec
       }
     }
   }
+
+  property("There should be no duplicates in qualityEditsList") {
+    qualityEditsList.toSet.toList.length mustBe qualityEditsList.length
```
nickgrippin: `List` has a method called `distinct` that might be helpful here (instead of converting to a set, then back to a list)

validation/src/test/scala/hmda/validation/engine/lar/quality/LarQualityEngineSpec.scala
```diff
@@ -25,4 +25,17 @@ class LarQualityEngineSpec
       }
     }
   }
+
+  property("There should be no duplicates in qualityEditsList") {
+    qualityEditsList.distinct.length mustBe qualityEditsList.length
+  }
+
+  property("There should be edits equal to the number of edit files - 1 (for Q022)") {
```
hkeeler: This is a pretty brittle way of determining the number of edits that _should_ be present.  What if some utility class file is added to that directory?  What if we want to turn off an edit by commenting it out of the list?  We'd have to delete the corresponding file as well.  It also doesn't test that its the expected set of edits, just the correct number.

kgudel:
Closing for now.

PR #430, opened 2016-07-05
Fix FileNotFoundException in LAR API

### LINKED ISSUE #426: Bug in census checks when using the HTTP API
When running the single lar validation API, the census checks fail due to a `FileNotFoundException`. There seems to be a problem with the relative path of the `tract_to_cbsa_2013.csv` file

```shell
[ERROR] Caused by: java.io.FileNotFoundException: model/src/main/resources/tract_to_cbsa_2013.csv (No such file or directory)
api[ERROR] 	at java.io.FileInputStream.open0(Native Method)
api[ERROR] 	at java.io.FileInputStream.open(FileInputStream.java:195)
api[ERROR] 	at java.io.FileInputStream.<init>(FileInputStream.java:138)
api[ERROR] 	at scala.io.Source$.fromFile(Source.scala:91)
api[ERROR] 	at scala.io.Source$.fromFile(Source.scala:76)
api[ERROR] 	at hmda.model.census.CBSATractLookup$.<init>(CBSATractLookup.scala:9)
api[ERROR] 	at hmda.model.census.CBSATractLookup$.<clinit>(CBSATractLookup.scala)
```

schbetsy:
Fixes #426 (I believe)

I tested this manually by running the project locally (`sbt`, then `reStart`), then using Postman to hit the validation endpoint (`localhost:8080/lar/validate`). The endpoint now returns the result of the validations, instead of throwing a `FileNotFoundException` and failing to return.

Should I test the fix in more places before we consider it done?

### DIFF
--- a/model/src/main/scala/hmda/model/census/CBSATractLookup.scala
+++ b/model/src/main/scala/hmda/model/census/CBSATractLookup.scala
@@ -6,7 +6,9 @@ import scala.io.Source
 
 object CBSATractLookup {
   val values: Seq[CBSATract] = {
-    val lines = Source.fromFile(new File("model/src/main/resources/tract_to_cbsa_2013.csv")).getLines
+    val cbsaResource = getClass.getResourceAsStream("/tract_to_cbsa_2013.csv")
+    val lines = Source.fromInputStream(cbsaResource).getLines()
+
     lines.map { line =>
       val values = line.split('|').map(_.trim)
       val name = values(0)
--- a/model/src/main/scala/hmda/model/census/CBSATractLookup.scala
+++ b/model/src/main/scala/hmda/model/census/CBSATractLookup.scala
@@ -1,13 +1,10 @@
 package hmda.model.census
 
-import java.io.File
+import hmda.model.ResourceUtils
 
-import scala.io.Source
-
-object CBSATractLookup {
+object CBSATractLookup extends ResourceUtils {
   val values: Seq[CBSATract] = {
-    val cbsaResource = getClass.getResourceAsStream("/tract_to_cbsa_2013.csv")
-    val lines = Source.fromInputStream(cbsaResource).getLines()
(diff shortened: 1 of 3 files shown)


lgdean:
You could do `sbt assembly` and then run with the resulting jar, but I wouldn't necessarily say you should, particularly if your current method was previously triggering the failure.

Speaking of which: when we had an internal error, it was failing to return, rather than returning some sort of error?  If so, that seems like an issue worth addressing (not as part of this fix, but eventually).

model/src/main/scala/hmda/model/census/CBSATractLookup.scala
```diff
@@ -6,7 +6,9 @@ import scala.io.Source
 
 object CBSATractLookup {
   val values: Seq[CBSATract] = {
-    val lines = Source.fromFile(new File("model/src/main/resources/tract_to_cbsa_2013.csv")).getLines
+    val cbsaResource = getClass.getResourceAsStream("/tract_to_cbsa_2013.csv")
+    val lines = Source.fromInputStream(cbsaResource).getLines()
```
lgdean: I'm torn: I notice that these lines are exactly like lines 9-10 in the other file (apart from the file name).  And so I want to suggest extracting a method and putting it into some convenient utility place, called `ResourceUtils` or some such bundle-of-things name.
But then there might have to be a `ResourceUtils` for each of `model`, `api`, etc. Which feels weird.

I still lean toward extracting it somewhere.  For one thing, maybe only `model` will even have such resources!  If the imagined situation becomes a problem, we can deal with it then.

schbetsy: Good idea. Done in ce854af.

schbetsy:
I also tested from the .jar, with the same results.

jmarin:
LGTM

PR #435, opened 2016-07-07
Fixes compiler warning for deprecated methods

### LINKED ISSUE #434: Fix code that causes compiler warnings
Compiling gives this warning:
```
[info] Compiling 1 Scala source to /Users/lortone/dev/hmda-platform/platform-test/target/scala-2.11/classes...
[warn] /Users/lortone/dev/hmda-platform/platform-test/src/main/scala/util/Dat2Csv.scala:34: method fromFile in object FileIO is deprecated: Use `fromPath` instead
[warn]       val source = FileIO.fromFile(datFile)
[warn]                           ^
[warn] /Users/lortone/dev/hmda-platform/platform-test/src/main/scala/util/Dat2Csv.scala:35: method toFile in object FileIO is deprecated: Use `toPath` instead
[warn]       val sink = FileIO.toFile(txtFile)
[warn]                         ^
[warn] two warnings found
```

Update the code so it doesn't give this warning.

jmarin:
addresses #434

### DIFF
--- a/platform-test/src/main/scala/util/Dat2Csv.scala
+++ b/platform-test/src/main/scala/util/Dat2Csv.scala
@@ -31,8 +31,8 @@ object Dat2Csv {
       val datFile = new File(datFilePath)
       val txtFile = new File(txtFilePath)
 
-      val source = FileIO.fromFile(datFile)
-      val sink = FileIO.toFile(txtFile)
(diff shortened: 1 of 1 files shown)


lgdean:
This fix looks good to me, and I plan to merge it.  While reviewing, I found a little more cleanup that I'll do in another PR on the same issue.

PR #440, opened 2016-07-08
use Paths.get for file paths; exit after processing file

### LINKED ISSUE #434: Fix code that causes compiler warnings

lgdean:
The `Paths.get` commit does additional cleanup and closes #434 

The other commit allows the program to exit once its work is done, which makes it a bit easier to use.

I used `sbt "project platformTest" "runMain util.Dat2Csv parser/src/test/resources/dat/ThirdTestBankData_clean_19039_2017.dat whatever.csv"` to run it; there may be other ways.  If you know those ways, feel free to let them inform your review.

I combined these commits into one PR because it seemed faster (and not harmful), but am happy to split them up if needed.

### DIFF
--- a/platform-test/src/main/scala/util/Dat2Csv.scala
+++ b/platform-test/src/main/scala/util/Dat2Csv.scala
@@ -17,14 +17,14 @@ Takes a .DAT file and converts to pipe delimited CSV (2017)
  */
 object Dat2Csv {
 
-  implicit val system: ActorSystem = ActorSystem()
-  implicit val materializer: ActorMaterializer = ActorMaterializer()
-  implicit val ec: ExecutionContext = system.dispatcher
-
   def main(args: Array[String]): Unit = {
     if (args.length != 2) {
       throw new Exception("Please provide input .DAT and output .txt files")
     } else {
+      implicit val system: ActorSystem = ActorSystem()
+      implicit val materializer: ActorMaterializer = ActorMaterializer()
+      implicit val ec: ExecutionContext = system.dispatcher
+
       val datFilePath = args(0)
       val txtFilePath = args(1)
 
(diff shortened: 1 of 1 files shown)


schbetsy:
is Dat2Csv used in the code anywhere? A quick search didn't find any calls to it.

schbetsy:
Also, it still doesn't terminate if I send it the wrong number of arguments (i.e. if it throws the exception on [line 26](https://github.com/cfpb/hmda-platform/pull/440/files#diff-3944c21ec796de4c943e07f526ed8defR26)).

lgdean:
I believe it is just for one-off uses. (Which is why I think it's ok to kill the actor system at the end.)

As for the wrong number of arguments: excellent point.  If I can fix it quickly, then I'll do that.  If not, then we can open another PR.  (Or we could decide not to: when the exception happens, it is annoying to have to control-c the program, but at least you're not sitting there wondering what happened.  Still, exiting would definitely be better behavior, and should be doable.)

schbetsy:
LGTM. Should I go ahead and merge?

lgdean:
I don't know of a reason to wait.

Issue #433, opened 2016-07-06
Create standard error response format for the API


wpears:
We'll need to create a standardized error response format so API consumers can react appropriately when things go wrong (eg, the front-end displays an error message). Will aid in addressing @schbetsy's observations [here](https://github.com/cfpb/hmda-platform/pull/422#discussion_r69803115) and [here](https://github.com/cfpb/hmda-platform/pull/422#discussion_r69805519)

Minimal example/starting point:

```
{
  path: <uri being requested>,
  code: <403 | 404 | 500 | etc>,
  message: <descriptive error message>
}
```

(`path` and `code` are redundant but convenient).

Issue #257, opened 2016-03-18
Parser: return strings for fields with numeric validations


schbetsy:
See this [comment chain](https://github.com/cfpb/hmda-platform/pull/89/files#r56238255) for context.

Open for discussion: There are some syntactical validations that test for numericality of a field. However, the parser already assumes these fields are numeric, so those syntactical validations never get a chance to fail. (And currently, if the fields are not numeric, the parser fails, rather than returning a helpful error). Should we change the parser so it always returns strings for these fields?

schbetsy:
This idea has been rejected. See #42 #98 and #62 for context and preferred solutions.

Issue #98, opened 2016-03-08
Validation (2017): S028 (Timestamp)


hkeeler:
**Edit #:** S028
**Type:** Syntactical
**Category:** Transmittal Sheet (only)
**Column:** Timestamp

Timestamp must be numeric and in ccyymmddhhmm format.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

lgdean:
In order for this edit to work as specified, we'd need to update the parser to return a string for this field. (Per #257). @jmarin (and anyone else), do you see a problem with that plan?

jmarin:
I see where you are coming from, but I think this change has negative consequences. In a type safe domain model like the one we are building, the [`TransmittalSheet`](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala) has the `timestamp` field which is coded as `Long` (as it should be, per the "Data Type" field in the File Spec). If the parser generates a String, we have a type mismatch between the File Spec and our model implementation, which I think is a step backwards. This might also be problematic from a backwards compatibility point of view (legacy data stores that field as numeric as well). 

The first part of this edit (numeric) is already checked for by the act of parsing the file due to type constraints. The format is checked in S028. So yes, a bit of this edit is redundant with out data model since a parseable file can't fail the numeric condition.

schbetsy:
So if the timestamp field is non-numeric in the submitted file, what happens from the user's perspective?

jmarin:
I think they should get an error stating that the file is malformed because of that reason. The parser doesn't do that yet

lgdean:
Business requirement question:
1. Does that error need to mention S028?

Process questions:
2. Is the relevant work, then, already captured as part of #62 (and/or #64)?

3. If the answer to question 1 is yes, then do we have (or do we need) a place to keep track of specific Syntactic edits whose failure messages need to come from the parser?

debseidner:
[...]

PR #444, opened 2016-07-11
LAR parser error handling

### LINKED ISSUE #62: Parser error handling

### LINKED ISSUE #431: Add parsing failure tests
Once #62 is done, add tests that check for parsing failures

nickgrippin:
Closes #62 
Closes #431


api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -31,10 +31,9 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol {
       path("parse") {
         post {
           entity(as[String]) { s =>
-            val lar = LarCsvParser(s)
-            //TODO: return human readable errors when parser fails. See issue #62
-            complete {
-              ToResponseMarshallable(lar)
+            LarCsvParser(s) match {
+              case Right(lar) => complete(ToResponseMarshallable(lar))
+              case Left(error) => complete(HttpResponse(StatusCodes.UnprocessableEntity))
```
nickgrippin: Still working on getting the errors added to the response.  If anyone knows an easy way, let me know!

kgudel:
I have placed a pull request on this branch that should be reviewed and merged or cancelled before merging this pr. It can be found [here](https://github.com/nickgrippin/hmda-platform/pull/1)

api/src/test/scala/hmda/api/processing/lar/SingleLarValidationSpec.scala
```diff
@@ -19,7 +18,7 @@ class SingleLarValidationSpec extends ActorSpec {
   val larValidation = createSingleLarValidator(system)
 
   val lines = Source.fromFile(new File("parser/src/test/resources/txt/FirstTestBankData_clean_407_2017.txt")).getLines()
-  val lars = lines.drop(1).map(l => LarCsvParser(l))
+  val lars = lines.drop(1).map(l => LarCsvParser(l)).filter(_.isRight).map(_.right.get)
```
hkeeler: I _think_ you can do something like:

```scala
val lars = lines.drop(1).map(line => LarCsvParser(line)).collect(Right(lar) => lar)
```

Seems cleaner.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -66,6 +69,11 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol {
       }
     }
 
+  def errorsAsResponse(list: List[String]): HttpResponse = {
+    val errorEntity = HttpEntity(ContentTypes.`application/json`, list.toJson.toString)
+    HttpResponse(StatusCodes.UnprocessableEntity, entity = errorEntity)
```
hkeeler: Do we really want a `422`?  By [definition](https://tools.ietf.org/html/rfc4918#page-78) it seems _OK_, but it also seems WebDAV-centric, and I don't think it is a very common error code.  I think a good ol' `400` with a standard error message format (to be defined https://github.com/cfpb/hmda-platform/issues/433) is probably a more consistent option.

Maybe something like:

```json
{
  "status_code": 400,
  "errors": [
    {"code": "field-not-parsable", "'Owner Occupancy' field is not an integer"},
    {"code": "field-not-parsable", "'Date of Action' field is not an integer"}
  ]
}
```

...or...

```json
{
  "status_code": 400,
  "errors": [
    {"code": "invalid-number-of-fields", "Incorrect number of fields: 37"}
  ]
}
```

This doesn't necessarily have to be fixed in this PR.  If we wait, let's link back to this as a refactor when we do #433.

hkeeler:
I'd like to question the whole point of parsing some fields into `int`s in the first place.  It seems like a weird middle ground of adding some small amount of type safety to `LoanApplicationRegister`.  What's the real value in this?  It doesn't mean those `int`s are _valid_.  Seems like this just saves a `toInt` in the edits, but at the cost of having to add a whole extra "pre-validation" layer.

The validity edits are already built for returning errors when bad data is passed in.  Do we really want a whole other logical flow when a user passes in non-`int` (say `ABC`) compared to an invalid int value (`-1`, `1000`, etc.).  This would likely cascade into the REST API as well, with two different failure states that'd need to be handled by the frontend and other clients.

I think we should pick a strategy, either:

1. Have all the fields be `String`, using the edit checks do **all** of the validation, and the only hard fail the parser on an invalid number of fields.
1. Make `LoanApplicationRegister` truly well-typed.  This would include:
    1. Using enum or sealed traits for all enum-like int fields (`agencyCode`, `loanType`).
    1. Using Java 8's `LocalDate` to represent data fields (`actionDate`, `loanDate`).
    1. Use `List`s for values when more than one option can be set (`appRace1-5`, `denial1-3`).
    1. Use `Option` when a given field supports `NA`.
    1. ...and probably lots of other type goodness.

Option 2 seems like a lot of work, and would likely change how validity errors in general would be handled by the system.  Option 1 seems like not-so-much work, but I'm not close to all of this work.

Perhaps we can discuss further at standup tomorrow.

schbetsy:
That's a great point, Hans. I agree we should discuss in standup. 

For reference, here are some links to previous conversations we've had around this topic:
- An issue I opened in March, suggesting we parse certain numeric fields as strings: https://github.com/cfpb/hmda-platform/issues/257
- A comment from Juan rejecting that idea, with some explanation: https://github.com/cfpb/hmda-platform/issues/98#issuecomment-203595518
- Another tangentially-related comment thread: https://github.com/cfpb/hmda-platform/pull/89/files#r56238255

api/src/main/scala/hmda/api/http/LarHttpApi.scala
nickgrippin: I'll make the switch over to `400` (that does make more sense), but I think we should wait until we have a standard error format before building out the response, since we'll probably have to refactor it anyway once #433 is complete.

parser/src/main/scala/hmda/parser/fi/FIDataCsvParser.scala
```diff
@@ -13,8 +13,9 @@ class FIDataCsvParser extends FIDataParser[String] {
   def parseLines(lines: Iterable[String]): FIData = {
     val tsLine = lines.head
     val ts = TsCsvParser(tsLine)
-    val lars = lines.tail.map(l => LarCsvParser(l))
-    FIData(ts, lars)
+    // Only parse LARs that do not contain errors
+    val larsWithoutErrors = lines.tail.map(l => LarCsvParser(l)).filter(_.isRight).map(_.right.get)
+    FIData(ts, larsWithoutErrors)
```
kgudel: It seems to me that we shouldn't return a collection of passing lars here. Instead we should return the group of failures with line counts.

nickgrippin: That idea might have potential, but regardless I think it's outside the scope of this ticket (the change in this file was just to address the different return type of `LarCsvParser`).  Maybe make this an issue where it can be discussed more?

Also, I can't find where this class is being used, so I can't say for certain whether the collection of failures or error-free LARs would be more appropriate.

schbetsy: @kgudel is right. This should return the list of errors when there are errors. As I understand it, we only process a file if it parses 100% correctly. If any lines don't parse, the whole file is invalid and the user must upload a whole new one, so it isn't useful to only return valid lines.

This class isn't being used yet because we haven't implemented the code that parses and validates a whole uploaded file. (Which I'm writing an issue about, and will bring up at standup).

nickgrippin: Ok, that does make a lot more sense.  So should changing the FIData behavior go into that issue that you're writing?  If not, I could create a new issue that addresses these changes, and implement the changes there.

schbetsy: I like the idea of creating a separate issue for this FIData behavior, so we can implement this first. I suspect the issue of parsing+validating whole files will be large, so it's good to reduce its scope where we can.

PR #445, opened 2016-07-11
remove Predicate boilerplate

lgdean:
addresses #408 

There are two changes here; both of them involve implicits.

The first is that we can implicitly create a `Predicate` from a Function, now that `Predicate` has just the one method. (This is the "bit more cleanup" mentioned in PR #409.)

The second is removing `implicit` from most of the defs in `PredicateCommon`, because they weren't being called implicitly, and generally couldn't be: if you import all of `PredicateCommon` (as it is before this change), then you have several implicit methods in scope, all with the same type signature.  So you can't implicitly treat a String (or generic T) as a `Predicate`, because the compiler wouldn't know which of these methods to choose.

Now only `equalTo` is implicit, which means we can use it in an implicit conversion when/if desired.  I changed it in two places where I think it reads nicely, to illustrate.

If the team likes these changes, we could do similar things in other `Predicate*` objects.

### DIFF
--- a/validation/src/main/scala/hmda/validation/dsl/Predicate.scala
+++ b/validation/src/main/scala/hmda/validation/dsl/Predicate.scala
@@ -4,3 +4,9 @@ trait Predicate[T] {
   def validate: T => Boolean
 }
 
+object Predicate {
+  import scala.language.implicitConversions
+  implicit def func2predicate[T](f: T => Boolean): Predicate[T] = new Predicate[T] {
+    override def validate = f
+  }
+}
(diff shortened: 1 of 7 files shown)


lgdean:
codecov is complaining that the coverage percentage is decreasing, which is true, but that's because there are fewer lines of code.  I might look into removing the unused lines, but it seems separate from this PR.

kgudel:
I like that this change makes the predicates terser through removing the boilerplate and I think that being able to use equalTo implicitly makes the code a bit cleaner as well. All in all LGTM.

schbetsy:
Looks good to me. Should we consider doing the same for the other Predicate files? Not necessarily in this PR; I'm just curious.

lgdean:
It was quick enough to do, that I went ahead and did it for the other files too.

schbetsy:
LGTM

PR #450, opened 2016-07-18
Add error reporting to Transmittal Sheet parser

kgudel:
The transmittal sheet parser is a little more complicated than the lar parser because the ts uses both ints and longs. In this pr I have attempted to use a single checkTs function that checks all values that have to be transformed and returns a single validationNEL that contain values of types int and long. Another approach might be to have two check functions that return separate validationNELS one which contains ints, the other with longs. The path I chose seemed originally to me to be a cleaner approach, however, I'm not sure anymore and welcome comments both on the details of the approach and whether it is the right one at all.


parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
```diff
@@ -1,47 +1,96 @@
...
+      )
+
+      val validationListInt = numericFields.map { case (key, value) => toIntorFail(value, key) }
+      val validationListLong = toLongorFail(fields(3), "Timestamp")
+      val validationListIntReduce = validationListInt.reduce(_ +++ _)
+      validationListIntReduce +++ validationListLong
+    }
+  }
+
+  def toIntorFail(value: String, fieldName: String): ValidationNel[String, List[AnyVal]] = {
```
nickgrippin: This should be `toIntOrFail` (capital O in "or").  Same with the method below.

parser/src/test/scala/hmda/parser/fi/ts/TsCsvParserSpec.scala
```diff
@@ -8,9 +8,32 @@ class TsCsvParserSpec extends PropSpec with PropertyChecks with MustMatchers wit
   property("Transmittal Sheet must be parsed from CSV") {
     forAll(tsGen) { (ts) =>
       whenever(ts.id == 1) {
-        TsCsvParser(ts.toCSV) mustBe ts
+        TsCsvParser(ts.toCSV).right.get mustBe ts
```
lgdean: I prefer `TsCsvParser(ts.toCSV) mustBe Right(ts)` here.
If you accept that one, consider analogous changes on other lines, too.

parser/src/test/scala/hmda/parser/fi/ts/TsCsvParserSpec.scala
```diff
@@ -8,9 +8,32 @@ class TsCsvParserSpec extends PropSpec with PropertyChecks with MustMatchers wit
   property("Transmittal Sheet must be parsed from CSV") {
     forAll(tsGen) { (ts) =>
       whenever(ts.id == 1) {
```
lgdean: I don't see the purpose of this `whenever` line.

Issue #437, opened 2016-07-07
Create initial Institution data model for use with edits


hkeeler:
This issue kicks off the work for building a financial institution data model required by various edit checks.  

We're going to start simple, initially including only:

* Institution ID
* Institution Name
* Regulating Agency
* 3rd Party Identifiers
    * FFIEC's RSSD ID
    * Federal Tax ID (EIN)
    * FDIC Certification Number
    * OCC Charter ID
    * NCUA Charter ID

With this data, we should be able to unblock [S025](https://github.com/cfpb/hmda-platform/issues/94) ~~and [Q012](https://github.com/cfpb/hmda-platform/issues/180)~~.

The initial implementation will simply read from a static file (likely CSV) at startup, and load the set of institutions into memory.  We will attempt to abstract out the underlying storage details from the query interface so it can be swapped out with a real storage backend at a future date without a bunch of refactoring.

hkeeler:
As discussed on https://github.com/cfpb/hmda-platform/issues/180#issuecomment-231439539, this does not actually unblock Q012.

Issue #94, opened 2016-03-08
Validation (2017): S025 (Control Number)


hkeeler:
**Edit #:** S025
**Type:** Syntactical
**Category:** Transmittal Sheet (TS) & Loan/Application Register (LAR)
**Column:** Control Number

Control number must = a valid respondent identifier/agency code combination for date processed.

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

kgudel:
This seems like it requires panel.

schbetsy:
What is the control number? I don't see a "Control Number" column in the filespec. (@lesgou ?)

lesgou:
@schbetsy It looks like control number is the combination of respondent id + agency code.
so, respondent id and agency code in row 1 of the data = respondent id and agency code in panel (which is currently not a concatenated field)

schbetsy:
Marking as "blocked". We will need panel to be implemented, so we can check against it that each Transmittal Sheet has an appropriate combination of Respondent ID and Agency Code.

Thanks @lesgou!

PR #452, opened 2016-07-26
Institution, InstitutionRepository, and S025

hkeeler:
This is a first stab at a simple Institution model, and an implementation of [S025](https://github.com/cfpb/hmda-platform/issues/94).  This approach assumes an `Institution` will be passed down into the edit.  Unfortunately, this requires **many** function signatures to be modified.  It's not _awful_, though it seems a little crude considering only 5 total edits need an `Institution` to do their job.

It also includes `InstitutionRepository` for looking up Institutions by ID.  The intent is that this would be used to lookup prior to executing the full suite of edits.  Unfortunately, there isn't _yet_ a place where all edits are executed from, so there's not good place to use this yet, but there will be soon.

Again, this is a WIP.  I will be adding more details as to why things are the way there are and moving the various `FIXME`s over to issues and todos on this issue shortly.

Closes https://github.com/cfpb/hmda-platform/issues/94
Closes https://github.com/cfpb/hmda-platform/issues/437


model/src/test/scala/hmda/model/institution/InstitutionSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+      failType mustBe RequiredExternalIdNotPresent(1, RssdId)
+      failType.message mustBe s"Institution 1 does not have an externalId of type $RssdId"
+    }
+
+    "resolve respondentId to RSSD ID when agency is CFPB and depository type is depository" in {
+      val inst = Institution(1, "Test Bank", externalIds, CFPB, Bank)
+      val expectedId = inst.respondentId
+
+      expectedId.isRight mustBe true
+      expectedId.right.get mustBe ExternalId("555555", RssdId)
```
lgdean: I prefer something like `expectedId mustBe Right(ExternalId("555555", RssdId))`

model/src/main/scala/hmda/model/institution/InstitutionRepository.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def findByExternalId(externalId: ExternalId): Option[Institution]
+}
+
+/**
+ * In-memory implementation of [[hmda.model.institution.InstitutionRepository]]
+ */
+class InMemoryInstitutionRepository(institutions: Set[Institution]) extends InstitutionRepository {
+
+  val instById: Map[Int, Institution] = institutions.map(i => (i.id, i)).toMap
```
hkeeler: Make this private.

model/src/main/scala/hmda/model/institution/InstitutionRepository.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  val instByExtId: Map[ExternalId, Institution] = institutions.flatMap(i =>
```
hkeeler: This is not used.  Toss it.

model/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -0,0 +1,53 @@
...
+        agency.externalIdTypes.get(dt) match {
+          case None => Left(UnsupportedDepositoryTypeByAgency(id, agency, dt))
+          case Some(extIdType) =>
+
+            extIdsByType.get(extIdType) match {
+              case None => Left(RequiredExternalIdNotPresent(id, extIdType))
+              case Some(extId) => Right(ExternalId(extId.id, extIdType))
+            }
+        }
+    }
```
lgdean: If we don't need the detailed error cases here (TBD), then this method can become much simpler, something like:
```
    val maybeRespId = for {
      depType <- institutionType.depositoryType
      extIdType <- agency.externalIdTypes.get(depType)
      extId <- extIdsByType.get(extIdType)
    } yield extId
    maybeRespId.toRight(RequiredExternalIdNotPresent(id, null)) // replace with whatever Left makes sense, or use Option
```
There may be slightly cleaner ways to write the last bit, but that's the basic idea.

model/src/main/scala/hmda/model/institution/InstitutionStatus.scala
```diff
@@ -0,0 +1,17 @@
+package hmda.model.institution
+
+import enumeratum.{ Enum, EnumEntry }
+
+/**
+ * The status of a financial institution
+ */
+sealed abstract class InstitutionStatus(override val entryName: String) extends EnumEntry
```
lgdean: currently unused, as we discussed but may not have noted.

hkeeler:
I think we're in pretty decent shape here.

1. [enumeratum](https://github.com/lloydmeta/enumeratum) - I'm using this library as Scala provides no equivalent to Java's `enum`.  I _generally_ like it, though it has a few quirks that I found odd.

    1. The good:
        1. Uses Scala's `sealed trait/abstract class` pattern, so you get all the algebraic datatype goodness.
        1. Easily lookup a specific case class by name or numeric value.  This is similar to Java enum's `valueOf`, which you cannot achieve with sealed trait/case classes without using Scala macros.
        1. Compile-time checking of name or value to make sure you haven't accidentally left in dupes.
    1. The less good:
        1. You can't mix [`Enum`](https://github.com/lloydmeta/enumeratum#enum) with [`ValueEnum`](https://github.com/lloydmeta/enumeratum#valueenum), which I think would be really useful for objects like `Agency` where we'll likely want to lookup the appropriate case class (say `CFPB`) by both the agency code (`9` in the TS and LAR) and by name (say `cfpb` in the REST API).  In these cases, I'm using `ValueEnum` as I think the lookup from the TS and LAR is more useful, and there's already JSON translation facilities built into Akka HTTP that can take care of the name-base resolution.

        Overall, I think enumeratum is OK, and probably worth its worts.  However, we _could_ implement our own simple `Enum`, and have it work _just_ how we'd like to use it.  I'll leave that to another PR, though.

1. `ValidationContext` - We've added `ValidationContext`, used for containing all things about a given validation request that cannot be derived from the file itself.  It currently only contains an `Option[Institution]`, but could also hold attributes such as:

    * `filingYear` - Once we have this, we can remove the hard-coded `2017` we have sprinkled in a few places.
    * `submissionId` - Not needed for any edits, but could be useful for logging.
    * `user` - Also, not needed for any edits, but could be useful for logging.

1. `Institution` - Contains all FI attributes required to perform the edits.  We decided not to update the existing `Institution` object as part of this PR as its tied up with Akka Persistence, and integrating _this_ with _that_ would have made this PR even larger.  However, I _do_ think they should be merged soon.

    There are a few attributes here that I'd like feedback on:

    1. `id: Int` - Unique identifier for an `Institution`.  We've made it an `Int` for now.  If we know at this point if its going to be something else, we _could_ fix it now.
    1. `externalIds: Set[ExternalId]` - Contains all _other_ identifiers (such as tax ID, RSSD ID, FDIC charter, etc.).  I built it this way because:
        1. It made the `respondentId` logic much cleaner.  
        1. Most of these IDs are not present for any given institution, so you'd end up with a lot of `Option`al properties.
        1. It'll be easy to add more ID types in the future...which we know we'll have to do.
    1. `respondentId` - This is where all the logic lives for deriving a `respondentId` based on agency and depository/non-depository.  I've added several cases and error types for dealing with different ways the data _could_ be bad, breaking `respondentId` resolution.  It's quite possible that some of the cases will not happen in our system based on which types of institutions we allow to file and whether we allow institutions with bad data into the system in the first place.  I will be creating a separate issue for this discussion shortly.

1. `InstitutionRepo` - This is simply used to retrieve a given `Institution` by its ID.  Unfortunately, there isn't really a place to use it yet.  Once the REST API can accept uploads for a given FI, we'll have a place for it. :)

    This is also likely to overlap with Institution persistence, so if that's ready first, this may be replaced with that instead.

1. `S025` - This edit is for both TS and LAR.  I don't think we have any other edits like this?  This lead to a few questions:
    1. Do we really want to run it for each LAR?  Is it OK to just test the TS?
    1. I put it in the `ts` package.  Are we OK with that since it really does both?  I didn't want to create a whole new package structure just for this one oddball.
    1. `S025` (as well as future Institution-based edits) is skipped whenever `ValidationContext.institution` is `None`.  This is currently a requirement for the LAR-parsing REST API, where you have no context of which institution your parsing it for.  From past discussions, I believe we want to continue this so FIs can parse file anonymously.

        I'm curious how we feel about this.  We don't have any other logic like this.  The other option would be to have that logic _in_ the edit itself, always `Success`ing if the `Institution` is not present.

More to come after standup...

validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
```diff
@@ -1,19 +1,29 @@
...
+
+  def checkSyntactical(lar: LoanApplicationRegister, ctx: ValidationContext): LarValidation = {
     val checks = List(
       S010,
       S020,
       S205
     ).map(check(_, lar))
 
+    // Exclude edits requiring institution data
+    if (ctx.institution.isDefined) checks :+ s025(lar, ctx)
```
lgdean: This line does not appear to have any effect. Same for the analogous one in `TsSyntacticalEngine`.  I now lean even more strongly toward trying another way (which I've started, but don't have a PR yet).

We'll also need tests for the intended functionality here, and there.

Side note: TIL a one-line `if` means codecov won't check whether it happens.  Too bad, but I suppose it has to be that way (because that's what line coverage means), and it's good to know!

hkeeler: Good find.  The last open item I have on my checklist for this PR is to write the tests for this logic.

hkeeler: This bug has been fixed.  In addition, I've made a very similar fix (https://github.com/cfpb/hmda-platform/pull/452/commits/f4ecec7818c1c6412069c37f422a3ce873b62f0c) to `LarQualityEngine`, which I borrowing to offending logic from, and was itself not working as expected.  Doh!  Classic example of copy/paste programming.  Let that be a lesson to me.

That said, this got past me primarily because I couldn't figure out a good way to test it.  There isn't currently (as far as I can tell) a clean way to get a list of which edits actually ran in a given "engine", so it's hard to write that class of tests.  The current option is to stage a given edit to fail, and if engine then fails, you can assume the edit was run.

We're not going to try to solve this on this PR.  I'll create separate issue for further discussion.

Again, good catch @lgdean!

hkeeler:
Removing the "WIP" flag.  Review away.

hkeeler:
Just to followup on the [above](https://github.com/cfpb/hmda-platform/pull/452#issuecomment-236234651) _5.ii_:

> `S025` (as well as future Institution-based edits) is currently skipped whenever `ValidationContext.institution` is `None`.  This is currently a requirement for the LAR-parsing REST API, where you have no context of which institution your parsing it for.  From past discussions, I believe we want to continue this so FIs can parse file anonymously.

This is no longer true.  As mentioned on #471, we are now doing this check within the `S025` edit.

model/src/main/scala/hmda/model/institution/Institution.scala
jmarin: I agree, pattern matching on `Option` is usually not want you want to do. A more idiomatic version of this function: 

```scala
  def respondentId: Either[InvalidRespondentId, ExternalId] = {
    institutionType.depositoryType.map { dt =>
      agency.externalIdTypes.get(dt)
        .map { extIdType =>
          extIdsByType.get(extIdType)
            .map(extId => Right(ExternalId(extId.id, extIdType)))
            .getOrElse(Left(RequiredExternalIdNotPresent(id, extIdType)))
        }
        .getOrElse(Left(UnsupportedDepositoryTypeByAgency(id, agency, dt)))
    }.getOrElse(Left(NoDepositoryTypeForInstitutionType(id, institutionType)))
  }
```

As @lgdean has pointed out, if you don't need to get error values this translates nicely to a `for-comprehension`. I'm also a bit suspicious of the `Either` usage here. Is it really necessary? I wonder if an `Option[RespondentId]` might work better

model/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -0,0 +1,54 @@
+package hmda.model.institution
+
+/**
+ * A financial institution, geared towards requirements for filing HMDA data.
+ */
+case class Institution(
+    id: Int,
```
jmarin: So this seems to be very similar to the previous implementation [here](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/fi/Institution.scala). Of course, this one is more complete and I favor it. But the other one needs to be removed. 
Also, the previous version is used in the HTTP API, i.e. [here](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala#L44). 

This needs to be consolidated. I'm ok with creating new issues that address this separately from this PR

model/src/main/scala/hmda/model/institution/InstitutionStatus.scala
```diff
@@ -0,0 +1,17 @@
...
+
+object InstitutionStatus extends Enum[InstitutionStatus] {
+
+  val values = findValues
+
+  case object Active extends InstitutionStatus("active")
+  case object Inactive extends InstitutionStatus("inactive")
```
jmarin: Similar to comment above, there was an initial attempt at this [here](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/fi/Institution.scala#L3), which results in duplication of code

validation/src/main/scala/hmda/validation/engine/lar/LarEngine.scala
```diff
@@ -1,23 +1,27 @@
...
-      checkQuality(lar)
+      checkSyntactical(lar, ctx) |@|
+      checkValidity(lar, ctx) |@|
+      checkQuality(lar, ctx)
     )((_, _, _) => lar)
 
   }
 
+  // FIXME: This is not actually used anywhere!  These should really be "macro" edits!
   def validateLars(lars: Iterable[LoanApplicationRegister]): LarsValidation = {
```
jmarin: Remove?

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S025.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.model.fi.ts.TransmittalSheet
+import hmda.model.institution.Institution
+import hmda.validation.context.ValidationContext
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax.PredicateOps
+import hmda.validation.dsl.{ Failure, Result, Success }
+
+object S025 extends { //} EditCheck[TransmittalSheet]{
```
jmarin: You probably wanted to remove `extends` as well

hkeeler: Huh.  Surprised that compiles.  Will fix.

validation/src/test/scala/hmda/validation/engine/lar/validity/LarValidityEngineSpec.scala
```diff
@@ -16,7 +18,7 @@ class LarValidityEngineSpec extends WordSpec with MustMatchers with LarValidityE
       }
 
       lars.foreach { lar =>
-        checkValidity(lar) mustBe a[scalaz.Success[_]]
+        checkValidity(lar, ValidationContext(None)) mustBe a[scalaz.Success[_]]
```
jmarin: The other tests are returning `.isSuccess mustBe true`. We should probably choose one (I don't think this was introduced as part of this PR).

model/src/main/scala/hmda/model/institution/Institution.scala
hkeeler: Agreed.  We definitely want to merge this with the existing `Institution`.  We were hesitant to do that as part of this PR while you were out since we didn't have a good understanding of what the implications would be with the Akka Persistence stuff.

I've created https://github.com/cfpb/hmda-platform/issues/472 to get the two implementations merged.

jmarin:
@hkeeler I've done a quick pass at this and I don't have any big concerns with it. Only my comment about duplication of `Institution`, as it stands right now there is a duplication of code and your proposed model here is decoupled from persistence and api, while the previous one is hooked up. That can be solved on different tickets though.

model/src/main/scala/hmda/model/institution/Institution.scala
hkeeler: I agree, I don't love the three-deep `match` logic.  That said, I personally find it more readable than the idiomatic alternative.  However, if others prefer the above style, I'm not opposed.

I do think this should have real error handling around it, not just wrapped in an `Option`.  There are multiple scenarios where data could be misaligned, and having a good error about _how_ the data is wrong could be key to fixing it.

model/src/main/scala/hmda/model/institution/InstitutionStatus.scala
hkeeler: 👍  -> https://github.com/cfpb/hmda-platform/issues/472

validation/src/main/scala/hmda/validation/engine/lar/LarEngine.scala
hkeeler: I was hesitant to remove this as it's not related to this PR.  Shall I create a new issue for this?  It's probably worth a bit of discussion on what it would be replaced with.

jmarin: Yeah, probably best in another issue

hkeeler: Cool.  Created https://github.com/cfpb/hmda-platform/issues/474 for this one.

hkeeler:
Great!  I've done a few simple fixes, and created a couple issues (#472, #474) for the other outstanding items.

Are we clear for mergification?

validation/src/main/scala/hmda/validation/engine/lar/LarEngine.scala
```diff
@@ -1,19 +1,22 @@
 package hmda.validation.engine.lar
 
 import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.model.institution.Institution
```
lgdean: This import is unused, here and in `LarSyntacticalEngine`, `S025`, and `TsSyntacticalEngine`. And a couple of other Engines too.

PR #456, opened 2016-07-27
shorter version of TS validation

lgdean:
This PR came from discussion when reviewing #450.  The goal is to illustrate ways to use scalaz validation, both for education as we all learn about scalaz, and potentially for use in the code if preferred.

The first two commits show two different, but similar, ways of constructing the `TransmittalSheet`.  (@kgudel, I showed you a version of this code last week, before leaving for travel; if you've already considered it, no need to re-read.)

The third is a simple refactoring of the two methods for converting strings to numbers, and is conceptually independent.  Happy to re-make it separately if needed.

### DIFF
--- a/parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
+++ b/parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
@@ -39,25 +39,17 @@ object TsCsvParser {
     val parent = Parent(parentName, parentAddress, parentCity, parentState, parentZip)
     val contact = Contact(contactPerson, contactPhone, contactFax, contactEmail)
 
-    val maybeTS = (
+    val maybeTS: ValidationNel[String, TransmittalSheet] = (
       id
       |@| code
       |@| timestamp
       |@| activityYear
+      |@| taxId.success
       |@| totalLines
-    ) {
-        TransmittalSheet(
-          _,
-          _,
-          _,
-          _,
-          taxId,
-          _,
-          respondent,
-          parent,
-          contact
-        )
-      }
+      |@| respondent.success
+      |@| parent.success
+      |@| contact.success
+    ) { TransmittalSheet }
 
     maybeTS.disjunction.toEither.bimap(_.toList, identity(_))
   }
--- a/parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
+++ b/parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
@@ -51,7 +51,7 @@ object TsCsvParser {
       |@| contact.success
     ) { TransmittalSheet }
 
-    maybeTS.disjunction.toEither.bimap(_.toList, identity(_))
+    maybeTS.leftMap(_.toList).toEither
   }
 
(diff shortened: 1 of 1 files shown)


parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
```diff
@@ -1,95 +1,71 @@
...
-    }
+  def toLongOrFail(value: String, fieldName: String): ValidationNel[String, Long] = {
+    convert(value.toLong, s"$fieldName is not a Long")
   }
 
-  def toLongOrFail(value: String, fieldName: String): ValidationNel[String, List[Long]] = {
-    Try(value.toLong) match {
-      case Failure(result) => s"$fieldName is not a Long".failure.toValidationNel
-      case Success(result) => List(result).success
+  private def convert[T](x: => T, message: String): ValidationNel[String, T] = {
```
kgudel: I like how this condenses the logic into one place!

parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
```diff
@@ -1,95 +1,71 @@
...
-    if (fields.length != 21) {
-      ("Incorrect number of fields. found: " + fields.length + ", expected: 21").failure.toValidationNel
-    } else {
-      val numericFields = ListMap(
-        "Record Identifier" -> fields(0),
-        "Agency Code" -> fields(2),
-        "Activity Year" -> fields(4),
-        "Total Lines Entries" -> fields(6)
-      )
+    maybeTS.disjunction.toEither.bimap(_.toList, identity(_))
```
kgudel: I'm wondering if we could use the same case logic here as currently exists to avoid some of the type conversions.

kgudel:
Overall, I am a big fan of this change. I think it makes the code a lot cleaner and more intelligible. +1 from me.

parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
lgdean: Something like this?
```
    maybeTS match {
      case scalaz.Success(transmittalSheet) => Right(transmittalSheet)
      case scalaz.Failure(errors) => Left(errors.toList)
    }
```

lgdean: Another option: `maybeTS.bimap(_.toList, identity(_)).toEither`

schbetsy:
LGTM

parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
lgdean: One more possibility: `maybeTS.leftMap(_.toList).toEither`

I prefer this one, but am open to match/case as well.  (I no longer like any `bimap` version.)

schbetsy: Are you planning to change this line in this PR, or should I go ahead and merge?

PR #459, opened 2016-07-27
WIP: Provide Better Error Messages Through the API

### LINKED ISSUE #433: Create standard error response format for the API

kgudel:
Closes #433 
Always feel free to comment on the approach. As it currently stands the error response looks something like this:

{
"httpStatus": 404,
"message": "Institution: 123456 not found"
}


hkeeler:
I would prefer we made the attribute names a bit more explicit.  We will likely add more attributes in the future, and these generic names could cause confusion.  How about:

```json
{
    "httpStatus": 404,
    "message": "Institution: 123456 not found"
}
```

Also, @wpears's original suggestion included a `path` attribute.  Was that left out for a particular reason?  I'm guessing that could be pulled out of the Routing DSL...somehow.

jmarin:
@kgudel Aren't you taking this too far? I have only done a quick look, but things like `PossibleInstitution ` seem to be doing more than this ticket is asking. 

If I look at the following line:

https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala#L71

I would want that response to come back with some additional detail, wrapping the HTTP code. I believe you already have the detail you need in the body of that function (i.e. what institution was searched, it wasn't found, error code 404).

Also, you seem to be modifying the lines that have `ToResponseMarshallable`, which are the "happy path" and provide results, which means that they have nothing to do with returning better error messages as described in this ticket. I think the proper scope for this is to just replace `NotFound`, `InternalServerError`, etc. with the appropriate responses in `JSON` format as described in the ticket.

kgudel:
In talking to @wpears he suggested that it would be useful to have as much detail in the error messages as possible, and so if the call was asking for a filing for an institution, returning different errors for if that institution didn't exist and if there was no such filing for an existing institution. To get that level of detail requires adding more logic to the routes.

To get that increased level of detail I thought that it would be useful to make it more clear when persistent actors were returning that there was no such (fill in the blank) rather than returning an empty object, which is where adding things like PossibleInstitution came in. In retrospect that was likely outside the scope of the ticket. However, now that it is done I do think that it makes the logic cleaner but I can easily remove it if we don't like it.

jmarin:
I think it is outside of the scope of this ticket. 

I think we should discuss that additional detail elsewhere, and assess if it's appropriate. 

In looking at the code I found it much harder to follow, so I would like this to focus just on the messages not being HTTP codes, but something a bit friendlier in `json` format (which you already have interspersed in this PR for the most part).

Issue #455, opened 2016-07-26
Dockerize hmda-platform and ui


ericspry:
Create `docker compose` configuration for `hmda-platform` and `hmda-platform-ui` so that the application and supporting backend services can be run together in development (i.e. in a laptop)

PR #460, opened 2016-07-27
Add docker-compose to run UI and API together

schbetsy:
Closes https://github.com/cfpb/hmda-platform/issues/455

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,11 @@
-# Docker compose configuration file for the HMDA Platform
-# Version: 1.0.0
-
-hmda:
-  build: .
-  ports:
-    - 8080:8080
-    - 2551:2551
+# Assumes UI repo is cloned in parent directory
+version: '2'
+services:
+  api:
+    build: .
+  ui:
+    build: ../hmda-platform-ui
+    ports:
+      - "80:80"
+    depends_on:
+      - api
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ drwxr-xr-x  25 lortone  staff   850B Jul 25 17:13 hmda-platform-ui/
 From `hmda-platform`'s root directory, run the following:
 
 ```shell
-docker-compose up -d
+docker-compose up -d --build
 ```
 
 This will bring up all the HMDA Platform services. The first run may take several minutes.
--- a/README.md
+++ b/README.md
@@ -101,13 +101,11 @@ java -jar target/scala-2.11/hmda.jar
 First, make sure that you have the [Docker Toolbox](https://www.docker.com/docker-toolbox) installed.
 
 If you don't have a Docker machine created, you can create one by issuing the following:
-
 ```shell
 docker-machine create --driver virtualbox dev
 ```
 
 After the machine is created, make sure that you connect your shell with the newly created machine
-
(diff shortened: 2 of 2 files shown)


schbetsy:
This is now ready for review. 

If you want to test on your machine, please follow the instructions I added to the README (and provide feedback to help me improve those instructions :) ). However, instead of using the `master` branch of `hmda-platform-ui`, use the `update-docker-nginx` branch, which is in [UI PR #175](https://github.com/cfpb/hmda-platform-ui/pull/175l)

awolfe76:
I'll test this afternoon but wanted to ask if creating a `docker-compose-dev.yml` file to do a few extra things for dev purposes should be part of this PR or should we create another issue to handle that? Either way works for me. 

Things like adding a `volumes:` on the UI side to allow us to still watch for changes during development ([an example in grasshopper](https://github.com/cfpb/grasshopper/blob/master/docker-compose.yml#L38)).

schbetsy:
I'd be happy to add a second `docker-compose` with more dev-friendly options. I'd vote we make that a separate issue for a future PR, just because I don't know much about what we want/how to do it, and I don't want to delay this PR.

hkeeler:
@awolfe76, I agree that it'd be nice to get the `volumes` setup like grasshopper.  I don't think there's a need for a separate Compose file, let's just be added right into the setup that @schbetsy created.

The grasshopper setup has:

```yaml
  volumes:
    - ../grasshopper-ui/dist:/usr/src/app/dist/
```

For hmda-platform-ui, would it just be:

```yaml
  volumes:
    - ../hmda-platform-ui/dist:/usr/src/app/dist/
```

...or is that app structured differently?

I'm also fine with this being a followup PR.  It'd probably be easier for @wpears or @awolfe76 to make that change, and be sure it's working the way they want.

awolfe76:
I just created #464 to update `docker-compose` for development.

docker-compose.yml
```diff
@@ -1,8 +1,11 @@
...
+# Assumes UI repo is cloned in parent directory
+version: '2'
+services:
+  api:
+    build: .
+  ui:
+    build: ../hmda-platform-ui
+    ports:
+      - "80:80"
+    depends_on:
```
hkeeler: Does `api` actually resolve to something in the container?  I thought that only works with [`links`](https://docs.docker.com/compose/compose-file/#/links), and not [`depends_on`](https://docs.docker.com/compose/compose-file/#/depends-on)?

awolfe76:
After much docker, -compose, -machine debugging I got this working.

👍  from me. Nice work @schbetsy!

docker-compose.yml
schbetsy: When I run `docker exec -ti hmdaplatform_ui_1 /bin/bash`, I can `curl http://api:8080/` and see this:
```
root@c540789ba433:/usr/src/app# curl http://api:8080/
{
  "status": "OK",
  "service": "hmda-api",
  "time": "2016-07-29T15:33:24.876Z",
  "host": "44e2e13fb69c"
}root@c540789ba433:/usr/src/app#
```

Also, I can see the correct api responses in the browser by visiting `/api`, `/api/institutions`, etc.

PR #465, opened 2016-07-28
Log API response time

### LINKED ISSUE #453: Log response time for HTTP APIs
Log response time for HTTP API internals, at `debug` level

nickgrippin:
Still needs tests

Closes #453


jmarin:
Hey @nickgrippin I think you have the right idea, but I wanted to propose an alternative implementation that is simpler and more concise. You can use a custom `directive` for this, I have pushed an example in this [branch](https://github.com/jmarin/hmda-platform/tree/response-time). 

The directive is as follows: 

```scala
def time: Directive0 = {
    val startTime = System.currentTimeMillis()
    mapResponse { response =>
      val endTime = System.currentTimeMillis()
      println(endTime - startTime)
      response
    }
  }
```
Code [here](https://github.com/jmarin/hmda-platform/blob/response-time/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala#L8)

And here is how you would use it: 

```scala
 val rootPath =
    pathSingleSlash {
      get {
        time {
          complete {
            val now = Instant.now.toString
            val host = InetAddress.getLocalHost.getHostName
            val status = Status("OK", "hmda-api", now, host)
            log.debug(status.toJson.toString)
            ToResponseMarshallable(status)
          }
        }
      }
    }
```
Code for this endpoint [here](https://github.com/jmarin/hmda-platform/blob/response-time/api/src/main/scala/hmda/api/http/HttpApi.scala#L24)

As you can tell, this is a simplistic example, it's printing to the console as opposed to logging. You can modify this approach to log to the appropriate level, based on the logging level configured. An exercise for the reader :-)

nickgrippin:
Thanks @jmarin!  I wasn't familiar with Scala directives, so I'll start work on this right away

api/src/main/scala/hmda/api/http/HttpApi.scala
```diff
@@ -21,12 +21,16 @@ trait HttpApi extends HmdaApiProtocol {
...
-          log.debug(status.toJson.toString)
-          ToResponseMarshallable(status)
+        time {
+          val requestTime = System.currentTimeMillis()
+          complete {
+            val now = Instant.now.toString
+            val host = InetAddress.getLocalHost.getHostName
+            val status = Status("OK", "hmda-api", now, host)
+            log.debug(status.toJson.toString)
+            log.debug("Elapsed time: " + (System.currentTimeMillis() - requestTime) + "ms")
```
jmarin: It looks like this line is a leftover from previous implementation? If that is the case, `requestTime` above should also be removed

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -34,10 +34,13 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol {
     pathPrefix("lar") {
       path("parse") {
         post {
-          entity(as[String]) { s =>
-            LarCsvParser(s) match {
-              case Right(lar) => complete(ToResponseMarshallable(lar))
-              case Left(errors) => complete(errorsAsResponse(errors))
+          time {
+            val requestTime = System.currentTimeMillis()
```
jmarin: Is this a leftover from previous implementation?

kgudel:
LG(and cool)TM, assuming the branch conflicts are resolved.

schbetsy:
Is this PR still open? When I try this locally, should I be seeing the API response times in SBT? I have my log level set to `info`, but I'm not seeing any output about individual API responses.

schbetsy:
Is this PR still open? 

When I run this locally, should I be seeing the "Request took X ms" logging in the SBT server log? I have my log level set to `debug`, but I'm not seeing any log output when I make requests to the API.

nickgrippin:
It's still open.  I'm not sure exactly where the logs are supposed to go, but I'm reading through Juan's event processor PR to try to get a better understanding.

nickgrippin:
So, in order to see the logs, follow these steps:

- In the `application.conf` file under the API project, set `loglevel` to `"DEBUG"`
- Run SBT, set the project to `api`, and run
- Hit the `0.0.0.0:8080` endpoint, and watch the console.  You should see something similar to: "Request took 51 ms" in the logs.

jmarin:
@nickgrippin @schbetsy Just for the record, the event processor has nothing to do with these logs. The steps laid out above by @nickgrippin are the correct ones to see the response times.

PR #466, opened 2016-07-29
Stripped down standard error response

### LINKED ISSUE #433: Create standard error response format for the API

kgudel:
Create an error response object and protocol to convert it into json and use that in place of only return status codes.

Closes #433

### DIFF
--- a/api/src/main/scala/hmda/api/model/ErrorResponse.scala
+++ b/api/src/main/scala/hmda/api/model/ErrorResponse.scala
@@ -3,5 +3,5 @@ package hmda.api.model
 case class ErrorResponse(
   httpStatus: Int,
   message: String,
-  parth: String
+  path: String
 )
--- a/api/src/main/scala/hmda/api/model/ErrorResponse.scala
+++ b/api/src/main/scala/hmda/api/model/ErrorResponse.scala
@@ -1,0 +1,7 @@
+package hmda.api.model
+
+case class ErrorResponse(
+  httpStatus: Int,
+  message: String,
+  parth: String
+)
--- a/api/src/test/scala/hmda/api/http/InstitutionsHttpApiSpec.scala
+++ b/api/src/test/scala/hmda/api/http/InstitutionsHttpApiSpec.scala
@@ -53,6 +53,7 @@ class InstitutionsHttpApiSpec extends WordSpec with MustMatchers with ScalatestR
       }
       Get("/institutions/xxxx") ~> institutionsRoutes ~> check {
         status mustBe StatusCodes.NotFound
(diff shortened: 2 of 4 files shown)


api/src/main/scala/hmda/api/model/ErrorResponse.scala
```diff
@@ -0,0 +1,7 @@
+package hmda.api.model
+
+case class ErrorResponse(
+  httpStatus: Int,
+  message: String,
+  parth: String
```
jmarin: I think you meant "path" here?

jmarin:
LGTM

api/src/test/scala/hmda/api/http/InstitutionsHttpApiSpec.scala
```diff
@@ -73,9 +74,11 @@ class InstitutionsHttpApiSpec extends WordSpec with MustMatchers with ScalatestR
       }
       Get("/institutions/12345/filings/xxxx") ~> institutionsRoutes ~> check {
         status mustBe StatusCodes.NotFound
+        responseAs[ErrorResponse] mustBe ErrorResponse(404, "xxxx filing not found for institution 12345", "institutions/12345/filings/xxxx")
       }
       Get("/institutions/xxxxx/filings/2017") ~> institutionsRoutes ~> check {
         status mustBe StatusCodes.NotFound
+        responseAs[ErrorResponse] mustBe ErrorResponse(404, "2017 filing not found for institution xxxxx", "institutions/xxxxx/filings/2017")
```
schbetsy: Is this the response we want? If a consumer of the api makes a request for a nonexistent institution, this response is misleading. I would think we want "Institution xxxxx not found".

I don't want to re-open this issue if it was already decided elsewhere, so let me know if there's discussion somewhere that I should refer to.

kgudel: This would require adding more logic to the api which is not in the scope of this pr. It probably should be ticketed though.

schbetsy:
LGTM. I'll draft a ticket for the error message comment I made.

PR #470, opened 2016-08-01
TS toDat method

### LINKED ISSUE #67: Implement toDAT method (TransmittalSheet)
Implement function that outputs the Transmittal Sheet in DAT format. The specification can be accessed [here](http://www.ffiec.gov/hmda/pdf/spec2016.pdf) (pages 8 and 9)

https://github.com/cfpb/hmda-platform/blob/master/model/shared/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala#L27

nickgrippin:
Closes #67

### DIFF
--- a/model/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala
+++ b/model/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala
@@ -1,5 +1,7 @@
 package hmda.model.fi.ts
 
+import hmda.model.fi.StringPaddingUtils
+
 case class TransmittalSheet(
     id: Int,
     agencyCode: Int,
@@ -10,7 +12,7 @@ case class TransmittalSheet(
     respondent: Respondent,
     parent: Parent,
     contact: Contact
-) {
+) extends StringPaddingUtils{
 
   def toCSV: String = {
     s"$id|${respondent.id}|$agencyCode|$timestamp|$activityYear" +
@@ -47,14 +49,6 @@ case class TransmittalSheet(
       contact.phone +
       contact.fax +
       padRight(contact.email, 66)
-  }
-
-  private def padRight(s: String, n: Int): String = {
-    String.format("%1$-" + n + "s", s)
-  }
-
-  private def padLeftWithZero(s: String, n: Int): String = {
-    String.format("%1$" + n + "s", s).replace(' ', '0')
   }
 }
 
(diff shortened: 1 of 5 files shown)


model/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala
```diff
@@ -21,10 +21,40 @@ case class TransmittalSheet(
...
+      contact.phone +
+      contact.fax +
+      padRight(contact.email, 66)
   }
 
+  private def padRight(s: String, n: Int): String = {
+    String.format("%1$-" + n + "s", s)
+  }
+
+  private def padLeftWithZero(s: String, n: Int): String = {
```
schbetsy: Looks like these methods want to be extracted into a string-manipulation utility.

parser/src/test/scala/hmda/parser/fi/ts/TsDatParserSpec.scala
```diff
@@ -19,34 +21,43 @@ class TsDatParserSpec extends FlatSpec with MustMatchers {
     respondent.zipCode mustBe "22102"
   }
 
-  it should "parse other info about the respondent" in {
+  property("generate proper dat format") {
+    forAll(tsGen) { (ts: TransmittalSheet) =>
+      val parsedTs = TsDatParser(ts.toDAT)
+      val updatedRespondent = parsedTs.respondent.copy(id = parsedTs.respondent.id.replaceFirst("^0+(?!$)", ""))
```
schbetsy: I'm having trouble understanding these two lines. It looks like extra symbols were inserted somewhere in the process, and this line is removing them. Why is that?

nickgrippin: The spec states that the respondent ID should be right-justified and zero-filled, which is what the `toDAT` method outputs.  However, the parser only trims leading and trailing spaces, so we have to manually trim the extra zeroes so that the two transmittal sheets match.

We had to do the same thing on a couple of the fields in the `LarDatParserSpec`.

model/src/main/scala/hmda/model/fi/ts/TransmittalSheet.scala
nickgrippin: Good point, I'll move these over.

parser/src/test/scala/hmda/parser/fi/ts/TsDatParserSpec.scala
schbetsy: So, currently when the parser persists the Respondent ID, it keeps all the unnecessary 0s at the front? That's a bug in the parser, right? 

Sounds like we should make a ticket to make the DAT parser work as the specification states.

schbetsy:
LGTM. I'll create an issue for the parser bug.

PR #476, opened 2016-08-02
Friendly docker dev

awolfe76:
- adds the local volume for the UI
- updates to README with instructions to use docker and `npm run watch` and to point to /hmda for more information about HMDA

### For testing 

In the browser, the expected result is to see the "Home" link with the FFIEC logo just below. Also, if you open up dev tools, you should see the API request and the correct response returned (even though its not rendered).

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,14 @@ To find your docker machine endpoint:
 docker-machine ip dev
 ```
 
+Also, for convenience when doing development on the UI, the `docker-compose` file uses a `volumes` which mount the local directory into the `hmda-platform-ui` container. This means you can make UI changes and refresh the browser to view them.
+
+``` shell
+# while still in the hmda-platform directory
+cd ../hmda-platform-ui
+npm run watch
+```
+
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


schbetsy:
Cool. I can indeed make changes in the ui and see them when I refresh. That looks great.

But it also looks like a few of the front-end issues that y'all solved last week are back. I no longer see the FFIEC logo, and I don't see any institution info. Here's the console output when I run the app: 
screen shot 2016-08-02 at 3 54 07 pm
Could that be an issue with the docker setup? (It could also be a problem with my instance of either repo, or known issues in the ui, I'm not sure)

schbetsy:
I don't want to hold up this PR while I'm on vacation. If the console errors are expected and/or unrelated, feel free to merge. I just wanted to make sure you were aware of them.

awolfe76:
Thanks @schbetsy! I think this is probably a branch thing. The only error I see is the `Unexpected institution status` error, which https://github.com/cfpb/hmda-platform-ui/issues/176 was created to fix. That fix will likely be part of https://github.com/cfpb/hmda-platform-ui/issues/181. Sound right to you @wpears?

To be safe I'll see if someone else has some time to run through it too.

kgudel:
Setting this up on my machine I also only found the `Unexpected institution status` lending credence to @awolfe76 theory that the error that @schbetsy found was a branch one. As such, I am merging.

Issue #471, opened 2016-08-01
How should we handle optional edits?


hkeeler:
As mentioned on #452, certain edits cannot be applied under certain circumstances.  For example, we want to allow anonymous LAR validations, and thus, all edits that compare LAR data to the institution data of the filer, cannot be applied.

Our first attempt on #452 was to filter out these edits in validation engines when the optional data (`Institution`) was not present.  However, we moved that logic to within the edit itself, always returning `Success()` when the optional data is not present.  This made testing much easier, but does give the false sense that a given edit _passed_ when it was really _ignored_.  It also means we'd be duplicating this logic within each edit that follows this pattern.

I'm curious to get other's thoughts on this.  There are currently only 5 edits with this optional institution .  I'm not sure if there are other types edits that need to be made optionalized.  That may drive whether or not this should be institution-specific, or something more general.

hkeeler:
@lgdean put together the following as a possible solution:

* https://github.com/hkeeler/hmda-platform/compare/inst-model...lgdean:no-institution-from-hkeeler

lgdean:
I pushed that idea farther [here](https://github.com/hkeeler/hmda-platform/compare/inst-model...lgdean:further-s025-changes).  Not totally polished: for example, it's weird that in `S025` the companion object comes before the class.  (And that `EmptyEditCheck` is there at all: it and the bulk of `inContext` would likely move somewhere more general, to serve other edits.)  Still, it shows the important parts of how the code could look, and how `S025` (and the like) would fit in with other `EditCheck`s in the Engines.

lgdean:
The ideas from the above branches (and the commits from #477, rebased) ended up in PR #494.  I tried to keep the code open to various possibilities for answering this question.  In particular, `EmptyEditCheck` is a class, not an object, and we could easily make its `name` a constructor parameter, if needed.

Meanwhile, each edit's companion object, such as `S025`, has an `inContext` method that calls out to shared code that decides whether to instantiate the actual edit (or use `EmptyEditCheck`).  This indirection means that the knowledge of the fact of institution-dependence lives in each edit, but the logic isn't duplicated.  It also worked well for addressing @schbetsy's [question about constructor visibility](https://github.com/cfpb/hmda-platform/pull/477#issuecomment-238927684).

[...]

PR #477, opened 2016-08-02
NOT TO MERGE: proposed S025 structure

lgdean:
One person noted that the changes linked from #471 were a hard to compare to the current code on master. Here's a version based on master, for discussion. It may have a somewhat messy git history, so it's probably not for merging at this time.

Also note: these changes are mentioned in #471, but they do not address the core question there, of whether the system needs to behave in a particular way when ignoring/skipping an edit. (They are related mainly in that, if we do need to support a particular behavior, then it may affect how we structure the code.)

### DIFF
--- a/model/src/main/scala/hmda/model/fi/HasControlNumber.scala
+++ b/model/src/main/scala/hmda/model/fi/HasControlNumber.scala
@@ -1,0 +1,6 @@
+package hmda.model.fi
+
+trait HasControlNumber {
+  def respondentId: String
+  def agencyCode: Int
+}
--- a/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
+++ b/model/src/main/scala/hmda/model/fi/lar/LoanApplicationRegister.scala
@@ -1,5 +1,6 @@
 package hmda.model.fi.lar
 
+import hmda.model.fi.HasControlNumber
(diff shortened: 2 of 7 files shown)


schbetsy:
Question: is there anything that prevents a developer from calling `s025(lar)` directly (without `.inContext`)? Should there be?

lgdean:
@schbetsy Yes, there is.  Well, mostly.  To create the `s025` in your example, you have to either call `S025.inContext`, or, if you want, directly construct it with `S025(institution)` or the like.  The latter is perhaps not ideal, though (IMO) probably not harmful.

We could make the constructor private or protected.  (Though that might then prevent sharing the `inContext` logic with code that creates `Q033`, which is in another package.)

Does that answer (or the proposed change) address the concern?

schbetsy:
Okay, I see what you mean. That makes me wonder: I can't (off the top of my head) think of any time we would _want_ to directly construct it (like `S025(institution)`; do you know any?  In any case, I think I agree that it isn't harmful that it's possible.


In any case, I like this strategy for handling optional edits. I think it would work well, and we could extend it easily to the other optional edits.

lgdean:
I also don't know of a case where we'd want to construct it directly.  So I'll see if I can prevent it, but not worry too much if I can't.

lgdean:
These changes now appear in #494. Thanks!

PR #480, opened 2016-08-03
Event processor

### LINKED ISSUE #12: Implement event controller
The event controller should be the main mechanism by which events get fired and are communicated to the relevant consumers. It decouples producers from consumers of events.

jmarin:
Closes #12 

At this stage, the event processor is very simple, it just reacts to the upload of a file by logging those events. This is the base for starting the processing pipeline (upload -> parse -> validate). 
The parsing stage, including persisting parsing errors for reporting, will be hooked up to the `UploadCompleted` event (this will be added in a separate PR, after this one).


persistence/src/main/scala/hmda/persistence/processing/HmdaRawFile.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    case StartUpload =>
+      log.debug("Start upload")
+      publishEvent(UploadStarted(submissionId))
+
+    case cmd: AddLine =>
+      persist(LineAdded(cmd.timestamp, cmd.data)) { e =>
+        log.debug(s"Persisted: ${e.data}")
+        updateState(e)
+        context.system.eventStream.publish(e)
```
kgudel: Is there a difference between this and the publishEvent method?

jmarin: Huh, oversight. Actually, I don't think we want to publish every line persisted as an event. Will remove this line

kgudel:
LGTM! I really like the idea of testing the logs as an integration test! Seems like it might be a useful way of testing #465 though it shouldn't stop it from being merged.

Issue #438, opened 2016-07-07
Add Institution Type and Parent data to Institution data model


hkeeler:
Once our initial Institution model in good shape, let's add institution type and parent data.  This should unblock edits [V110](https://github.com/cfpb/hmda-platform/issues/105) and [Q033](https://github.com/cfpb/hmda-platform/issues/178), which both require these data element.

This is dependent on #437.

hkeeler:
We actually did a bit of this work on PR https://github.com/cfpb/hmda-platform/pull/452.  With that PR we do have an `InstitutionType`, each of which designated as _depository/non-depository_.  However, not all types are represented, and what we do have are all _depository_:

* Bank
* Credit Union
* Savings & Loan

We also have some [temporary types](https://github.com/cfpb/hmda-platform/pull/452/files#diff-eae408d1158c3baa5448f38605801612R24) for testing, that should be replace with real values once we know what they are.  Hopefully, a semi-final list can be compiled as part of this issue.

The following are mentioned within edits:

* Bank,
* Saving Association (same thing as S&L I believe)
* Independent Mortgage Company
* Mortgage Banking Subsidiary

Hopefully, each can be represented as an `InstitutionType`. If not, we may need another level of categorization, and/or some combination of `InstitutionType` with parent data.

As for parent institution data adding to `Institution`, it _could_ be as simple as adding a `parents: Set[Institution]` attribute.

lgdean:
Created #534 and #563 for some miscellaneous loose ends; can now close this ticket.

PR #481, opened 2016-08-03
Update Institutions model

### LINKED ISSUE #472: Merge new and old Institution classes
Once #452 is merged, we should replace the previous [`hmda.model.fi.Institution`](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/fi/Institution.scala) with the new `hmda.model.institution.Institution` version.

nickgrippin:
Still working on the JSON spray stuff

Closes #472


api/src/test/scala/hmda/api/model/ModelGenerators.scala
```diff
@@ -70,4 +78,41 @@ trait ModelGenerators {
...
+  }
+
+  implicit def institutionTypeGen: Gen[InstitutionType] = {
+    Gen.oneOf(
+      Bank,
+      CreditUnion,
+      SavingsAndLoan,
+      NonDepositInstType,
+      NoDepositTypeInstType
+    )
```
lgdean: I prefer `Gen.oneOf(InstitutionType.values)` (assuming it works) so I won't have to remember to modify it in #438.
Same deal for `Agency`, `ExternalIdType`, and any others I missed.

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -27,8 +27,13 @@ object InstitutionPersistence {
         case InstitutionCreated(i) =>
           InstitutionsState(institutions + i)
         case InstitutionModified(i) =>
-          val x = institutions.find(x => x.id == i.id).getOrElse(Institution())
-          InstitutionsState((institutions - x) + i)
+          if (institutions.exists(x => x.id == i.id)) {
+            val x = institutions.find(x => x.id == i.id).get
+            InstitutionsState((institutions - x) + i)
+          } else {
+            InstitutionsState(institutions + i)
```
lgdean: I suspect this line is untested. (Check codecov results when available.)

I'm also not too happy about iterating over the list twice and using `get`.  Are you aiming to improve readability?  If so, I suggest:
```
          val others: Set[Institution] = institutions.filterNot(_.id == i.id)
          InstitutionsState(others + i)
```
We may eventually want to use something that supports fast+easy lookup by id, too.

No strong preference about the `x => x.id` vs `_.id` here, by the way.

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -73,7 +78,7 @@ class InstitutionPersistence extends PersistentActor with ActorLogging {
       }
 
     case GetInstitutionById(institutionId) =>
-      val institution = state.institutions.find(x => x.id == institutionId).getOrElse(Institution())
+      val institution = state.institutions.find(x => x.id.toString == institutionId)
```
lgdean: Note: our choice of `Int` for the ID is not set in stone (see point 3.i [here](https://github.com/cfpb/hmda-platform/pull/452#issuecomment-236234651)), so feel free to keep using `String` if it makes your changes easier, or seems better for some other reason.  I realize you'll have to change some code either way, though.

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
nickgrippin: This is definitely more readable, thanks!

jmarin:
Overall, LGTM
I'm surprised that `Institution` id is now an `Int`. That will not work with [`LEI`](https://www.gleif.org/en/lei-focus/about-the-lei/iso-17442-the-lei-code-structure) in the future. Since it's a topic that is not related to this PR, it shouldn't hold it back

PR #483, opened 2016-08-08
Strip leading zeros from .dat

### LINKED ISSUE #475: DAT Parser should strip leading zeroes from parsed fields
DAT Parser (Transmittal Sheet and LAR) should strip leading zeroes from any fields that have them, so it can accept files according to the file spec. See [this comment thread](https://github.com/cfpb/hmda-platform/pull/470#discussion_r73147076) for context. 

Here's a list of all fields that are marked as having leading zeroes in the FFIEC (DAT) [File Spec](http://www.ffiec.gov/hmda/pdf/spec2016.pdf). Ensure the parser handles all of these correctly. Investigate whether this is necessary for numeric fields (Loan Amount, Applicant Income), or whether `toInt` handles this automatically.

Transmittal Sheet:
- [ ] Respondent ID

LAR:
- [ ] Respondent ID
- [ ] Loan Amount
- [x] ~~State Code~~
- [x] ~~County Code~~
- [ ] Applicant Income

Make sure to update any specs that are affected by the change, for instance `LarDatParserSpec` and `TsDatParserSpec`.

kgudel:
Strips leading 0s from Applicant Income, it does not change Loan Amount because it is already converted to an int or Respondent ID because it is valid for it to have leading 0s.

Closes #475

### DIFF
--- a/parser/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
+++ b/parser/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
@@ -5,7 +5,7 @@ import hmda.model.fi.lar._
 object LarDatParser {
   def apply(s: String): LoanApplicationRegister = {
     val id = s.substring(0, 1).trim.toInt
-    val respId = s.substring(1, 11).trim.replaceFirst("^0+(?!$)", "")
+    val respId = s.substring(1, 11).trim
     val code = s.substring(11, 12).trim.toInt
     val loanId = s.substring(12, 37).trim
     val loanDate = s.substring(37, 45).trim
--- a/parser/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
+++ b/parser/src/main/scala/hmda/parser/fi/lar/LarDatParser.scala
@@ -5,7 +5,7 @@ import hmda.model.fi.lar._
 object LarDatParser {
   def apply(s: String): LoanApplicationRegister = {
     val id = s.substring(0, 1).trim.toInt
-    val respId = s.substring(1, 11).trim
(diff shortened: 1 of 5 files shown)


nickgrippin:
So, my one concern with this approach (and it may not even matter, since we're not supporting DAT anyways), but what if an institution has leading zeroes for their Respondent ID (like, `0045`)?  Or if they have `0000000000` as an entry?

kgudel:
In my mind an all `0` entry shouldn't be a concern. Theoretically if all `0`s was valid that field should have an `NA`. @debseidner might be able to speak more to that though.

debseidner:
@nickgrippin @kgudel I think there may be cases where the respondent ID can have a valid leading 0 value and we would not want to strip the 0s. I checked previous filings and there are tax ids with leading 0s. Those are probably considered real 0's and not leading as a tax ID has to be XX-XXXXXXX so I don't think an institution would drop the leading 0 for a tax ID.  

I think that is that is the only case because I checked FDIC cert number on NIC and charter number on NCUA's and OCC"s sites and they don't appear to have leading zeros outside of HMDA.

kgudel:
Just to make sure I'm on the same page, if we are taking a .dat file we would not want to strip leading 0s from respondent id. Did I get that right?

debseidner:
I think the same concerns would apply with a .DAT - you still wouldn't be able to strip the leading zero of a tax id. You can strip everything else if you needed to.

debseidner:
Also, I don't think anyone can send an all zero respondent id - if you stripped the all 0s, you would just have a blank field, which should produce a parsing error.

nickgrippin:
LGTM

PR #485, opened 2016-08-08
Introduce Akka request-timeout config option

### LINKED ISSUE #443: on unexpected errors, return 500
In #430 we had a failure that was causing an exception to be thrown, and the relevant endpoint not to return. That specific problem was fixed, but in general, it would be better to return a 500 rather than simply not giving the client any information.

May depend on #433.
May be separate pieces of work for different parts of the API.

kgudel:
Akka has a global http request timeout config option that I am making explicate here. It's default is 20 seconds, I have set it to the more reasonable 5 seconds. By default if a timeout occurs the route returns:

`The server was not able to produce a timely response to your request.
Please try again in a short while!`

Further I have made sure that this feature is working properly, the code to test it is in the first commit in this pr.

More info on akka http timeouts can be found [here](http://doc.akka.io/docs/akka/2.4/scala/http/common/timeouts.html)

Closes #443

### DIFF
--- a/api/src/main/resources/application.conf
+++ b/api/src/main/resources/application.conf
@@ -4,6 +4,7 @@ akka {
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
   http.parsing.max-content-length = 1G
   http.server.default-host-header = "cfpb.gov"
+  http.server.request-timeout = 5 s
 }
 
 hmda {
--- a/api/src/main/scala/hmda/api/http/HttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/HttpApi.scala
@@ -36,20 +36,7 @@ trait HttpApi extends HmdaApiProtocol {
       }
     }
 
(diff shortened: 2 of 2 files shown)


jmarin:
Why 5 seconds? I don't think we should bake this into the configuration unless it's an environment variable. This sort of value is great to override at runtime, with a value that is specific to the conditions of the supporting environment (i.e. latency, percentiles for response time, etc.).

kgudel:
Do you think that that is something reasonable to start working on now, or should I just close this out along with the ticket that it comes from, as akka will return a timeout after 20 seconds anyway, and make a new ticket for the more precise version?

jmarin:
I guess I don't see the relation between shortening the timeout and ticket #443 
As for the timeout, I would leave it as default until we know more about how our platform runs in a production like environment. That will involve measuring (as you can guess, I created #453 with this in mind)

kgudel:
Sounds good!

PR #487, opened 2016-08-09
Parser persistence

### LINKED ISSUE #458: Add HMDA Parser persistence

jmarin:
Closes #458


persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -0,0 +1,110 @@
...
+  override def postStop(): Unit = {
+    log.debug(s"Parsing ended for $submissionId")
+  }
+
+  override def persistenceId: String = s"$name-$submissionId"
+
+  override def receiveCommand: Receive = {
+
+    case ReadHmdaRawFile(persistenceId) =>
+      val parsed = events(persistenceId)
```
kgudel: Just to make sure I understanding what is happening here this is creating a flow, starting from the sink which is `events(persistenceId)` that is being executed below on line 77. Is that correct?

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+      parsed
+        .runForeach(parsed => self ! parsed)
+        .andThen {
+          case _ =>
+            self ! Shutdown
+        }
+
+    case lp @ LarParsed(lar) =>
```
kgudel: To further my own understanding I was wondering why the parsed rows and errors are persisted here instead of in the stream above.

kgudel:
Since it seems like this PR is only to persist Lar data is it worthwhile to make another ticket to address the transmittal sheet?

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+  val hmdaFileParser = createHmdaFileParser(system, submissionId)
+
+  val probe = TestProbe()
+
+  val timestamp = Instant.now.toEpochMilli
+  val lines = fiCSV.split("\n")
+  val badLines = fiCSVParseError.split("\n")
+
+  override def beforeAll() {
+    persistRawData()
```
kgudel: With the current tests I don't see what is using this data?

kgudel:
Is it worthwhile to test the ReadHmdaRawFile case? Though I suppose that that kind of integration testing might want to wait until we have things like validity persistance as well to do a more full test?

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
jmarin: Close, but no. If you got to the `events` function in `HmdaQuery`, you can see its signature, it's a `Source[Event, NotUsed]`. So basically it reads a stream of events from the journal, for a specific `persistenceId`.
The `runForEach` method is a shortcut to run a `Source` with a foreach procedure

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
jmarin: Separation of reading from writing. The responsibility for doing both is currently encapsulated within the same actor, but that might not be the case in the future.

kgudel: Oh interesting. So this is preparing for that separation. I hadn't thought about that. Thanks!

kgudel:
This LGTM

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+  val hmdaFileParser = createHmdaFileParser(system, submissionId)
+
+  val probe = TestProbe()
+
+  val timestamp = Instant.now.toEpochMilli
+  val lines = fiCSV.split("\n")
+  val badLines = fiCSVParseError.split("\n")
+
+  override def beforeAll() {
+    super.beforeAll()
```
schbetsy: What is this `override def beforeAll()` doing? I might be missing something, because it looks to me like it ends up being a no-op / empty method.

jmarin: It's doing nothing. Leftover from previous commit. Removing

schbetsy:
LGTM

PR #489, opened 2016-08-10
Prevent upload to existing submission

### LINKED ISSUE #484: Prevent upload to existing submission
Avoid uploading data to a submission that already exists and has data

kgudel:
This PR creates a directive to prevent the uploading of a file to a submission with a status other than created.

Closes #484

### DIFF
--- a/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
+++ b/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
@@ -1,11 +1,30 @@
 package hmda.api.http
 
+import akka.actor.{ Actor, ActorSystem }
 import akka.http.scaladsl.server.Directive0
 import akka.http.scaladsl.server.Directives._
 import akka.event.LoggingAdapter
+import akka.http.scaladsl.marshalling.ToResponseMarshallable
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse, StatusCodes }
+import akka.stream.ActorMaterializer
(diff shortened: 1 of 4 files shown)


jmarin:
I feel like creating a custom directive for this purpose is overkill, since this functionality is very limited in scope. I would change the implementation to a function within the body of `InstitutionsHttpApi`. 

Also, I think you are checking for submission = `Created`, which is not correct. The submission can be created and empty (i.e. without data) which means that the upload can happen. Uploads should only be prevented when there is a submission that already contains data (`SubmissionState.submissions` within `SubmissionPersistence` is not `Nil`)

kgudel:
While it is true that the new directive has a very specific function that is only intended to be used in one route I prefer it here to a function. I believe that the directive more effectively segregates the logic used for this check from the logic of the route itself. In so doing I think that you end up with cleaner and more understandable code. From my experience with #459 it seems that trying to add new functionality to the route itself ends up in code where it is unclear what code is serving each function and so very hard to read.

I also think that using a directive for this kind of code is more sustainable as the routes get more complicated. Using directives grows the code very slowly with each new feature whereas new functions would grow them considerably faster.

jmarin:
I stand by my first comment, I don't think creating a directive here is appropriate, given how we are treating them in the project (generic abstractions, such as the time directive I suggested for measuring response times, or authorization mechanisms).

kgudel:
This PR is replaced by the introduction of #495 and #496

PR #493, opened 2016-08-11
Add user authorization to API

### LINKED ISSUE #441: Add user authorization to REST API
We need to protect our RESTful resources from unauthorized access.  Our goal is to keep authentication and authorization as simple as possible within our APIs, using an upstream security service (coming soon...) for handling all the gory auth details.  All our API code has to care about are the existence of the following HTTP headers.

#### `CFPB-HMDA-Username`

This header will contain the username of the user accessing the API.  We will check for it's existence by all but the root (`/`) resource.  Lack of this header will result in a `403` HTTP error, and an appropriately detailed error message in the response (related to #433).  This username will be available for use in logging or any other features that need to know _who_ took a given action.

#### `CFPB-HMDA-Institutions`

This header will contain a comma-separated list of institution IDs (probably integers, see #437).  We will check for its existence on all resources rooted from `/institutions`, which at this point is pretty much everything.  Think of it as a whitelist of institutions authorized to a given user.

Similar to the `CFPB-HMDA-Username` header, a `403` error will be returned if:
* The header is not present.
* ~~An attempt is made to access an institution-specific resource (`/institutions/<institution_id>/*`) not in the user's whitelist.~~ (Now a separate issue: #504)

This header will also be used in the future for filtering lists of institutions (`/institutions`), but that functionality is likely tied to #437, so I don't want to block this one.

#### Implementation

Akka HTTP's provides an [`authorize()`](http://doc.akka.io/docs/akka/2.4.8/scala/http/routing-dsl/directives/security-directives/authorize.html#authorize) directive for just this sort of thing.

Eventually the upstream security service will guarantee that these headers are not being spoofed.  In the meantime, this lack of anti-spoofing keeps this from being a block for testing and frontend integration.

schbetsy:
Closes #441 

Also refactors API. Since the diff is hard to read, some details:
- create `timedGet` and `timedPost` directives, to reduce nesting levels and remove repetition of `get { time {`
- create `completeWithInternalError` method to abstract repeated lines dealing with throwable errors

Known issue: When someone makes an unauthorized request to the API, the response does not include the path they requested (currently, the `path` value in the error response is empty). I'm hoping to split that into a separate issue, since it would take some reworking of the directives, and it seems lower priority than hooking up the rest of the authorization.


api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -71,9 +69,7 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
               }
             case Failure(error) =>
               filingsActor ! Shutdown
-              log.error(error.getLocalizedMessage)
-              val errorResponse = ErrorResponse(500, "Internal server error", path)
-              complete(ToResponseMarshallable(StatusCodes.InternalServerError -> errorResponse))
+              fiveHundred(path, error)
```
jmarin: When I first read this I found this confusing. Not the fact that it's a function, which I like, but the nomenclature. I would expect something like `completeWithInternalError`, `returnInternalError` or similar

schbetsy: Good suggestion. I'll update the name of the function.

api/src/test/scala/hmda/api/http/LarHttpApiSpec.scala
```diff
@@ -72,26 +75,42 @@ class LarHttpApiSpec extends WordSpec with MustMatchers with ScalatestRouteTest
...
       //should fail Q036
-      Post("/lar/validate?check=quality", badLar) ~> larRoutes ~> check {
+      postWithCfpbHeaders("/lar/validate?check=quality", badLar) ~> larRoutes ~> check {
         status mustEqual StatusCodes.OK
         responseAs[List[ValidationError]].length mustBe 1
       }
     }
   }
 
+  "reject requests without 'CFPB-HMDA-Username' header" in {
```
nickgrippin: Small thing here, but the Username test is adding the `institutionsHeader` (and vice versa).

nickgrippin:
LGTM, aside from merge conflicts.

I love the `timedGet` and `timedPost` directives, great idea!

kgudel:
This LGTM, but if there become a bunch of moving pieces in authorization that make the system hard to test it might be worthwhile in the future to have a config option to turn it off. That is a down the road question though. Merging for now.

Issue #264, opened 2016-03-28
Describe Persistence requirements (File Submission)


jmarin:
The HMDA Platform has a strong requirement for managing the state of the submission process. A HMDA Submission can be in one of the following states: 

1. Uploaded to the system: the user has **sent** the HMDA File to the system for validation
2. Parsed: the file has been **parsed** and found to be structurally correct as per the File Spec. If errors are present, they need to be corrected and the file resent. 
3. Valid: the contents have been **checked** for correctness and validity, and the submission is ready to be verified by the user.If there are syntactical or validity errors, the user must correct them and resend the file.
4. Checked: the information has been **validated** for consistency with macro and quality errors, and needs input from the user to explain the edit errors found. 
5. Verified: the user has **verified** the results of the macro and quality errors
6. The user has sent **signed** confirmation that the information is correct and has been verified. 

Every time a file is sent again during the submission process (or as a result of resubmissions) the process starts from step 1) above. 

From a system perspective, for each submission we want to store the following information: 

1) Raw data sent to HMDA Platform. The contents of the information sent by the FI are stored in their original text format. 
2) Parsing results. Store the results of the parsing operation (especially errors). 
3) Syntactical edits (errors)
4) Validity edits (errors)
5) HMDA content is checked (passes syntactical and validity)
6) Quality edits (errors)
7) Macro edits (errors)
8) Justification to Quality edits (persisted individually)
9) Justification to Macro edits (persisted individually)
10) HMDA content is verified (by user providing feedback about macro/quality errors).
11) Edit reports (IRS, Summary report, etc). Triggered by previous step
12) Submission verified (signed) by user

Steps 1-6 are triggered without user intervention, with one exception: if the parsing fails, the HMDA File needs to be corrected and resubmitted. The file will also need to be resubmitted in case of a system failure in step 1. Steps 8-10 and 12 require user input to proceed. 

The system must provide a **full audit trail** of the process. It needs to log every partial, incomplete or invalid submission and persist errors for reporting to the user and further analysis.

lesgou:
1. "Valid: the contents have been checked for correctness and validity, and the submission is ready to be verified by the user.If there are syntactical or validity errors, the user must correct them and resend the file."  
To clarify, does this mean that there are two "stages" of validation? The first is through the syntactical and validity edits, and will need the user to upload a new file. And the second is after the quality and macro edits, prior to the user's verification of each edit?


[...]

PR #495, opened 2016-08-11
Prevent uploading to submission with data 2

### LINKED ISSUE #484: Prevent upload to existing submission

kgudel:
This PR along with #496 replace #489

This PR has been chosen over #496.

It is an implementation of preventing an upload to a submission that already has data using a small exterior function but putting most of the logic within the route.

Closes #484


schbetsy:
I haven't had time to fully review, but I wanted to make a quick comment. My initial impression is that I prefer this version. I think it could be the more readable option, if we can change the order of the cases a bit and maybe find clearer names for a few variables. 

I'll continue to review in the morning, and I might contact you if it seems easier to talk through my ideas instead of typing them.

nickgrippin:
From a purely stylistic viewpoint, I think I prefer the other approach.  This file is already a bit bloated, and separating out the duplicate submission logic into it's own function seems cleaner and easier to follow when reading through the code.  But it's purely an aesthetic choice, I don't really see any major difference in performance between the two approaches.

awolfe76:
> Why are "create a submission" and "upload a file" separate steps/endpoints?

I _think_ it was as simple as it would give us the ability to see those steps and provide feedback if necessary. This issue should help https://github.com/cfpb/hmda-platform/issues/264 and this comment https://github.com/cfpb/hmda-platform/issues/264#issuecomment-203686220.

awolfe76:
Also [this file](https://github.com/cfpb/hmda-platform-ui/blob/master/submission-states.md) has the states as the front-end sees them today.

jmarin:
I find this version a bit clearer / more readable. There are merge conflicts at this point, those will need to be resolved as well.

kgudel:
Ok, then we will go with this one and I will close out the other. @schbetsy it would be great to talk about your suggestions for how I can make this one more readable.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -162,38 +164,56 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
-                val errorResponse = ErrorResponse(400, "Invalid File Format", path)
-                complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
-            }
-          }
+      extractExecutionContext { executor =>
+        val uploadTimestamp = Instant.now.toEpochMilli
+        val processingActor = createHmdaRawFile(system, submissionId)
+        val submissionsActor = system.actorOf(SubmissionPersistence.props(institutionId, period))
+        implicit val ec: ExecutionContext = executor
+        val isNotOverwrite = preventSubmissionOverwrite(submissionsActor, submissionId.toInt)
```
schbetsy: I'd like to see a variable name here that indicates that it's a `future`. It seems like lower-case `f` at the start of a variable indicates `future` in other places. So maybe `fCheckForOverwrite` or `fCheckSubmissionStatus`.

OR, you could just circumvent the naming problem with: `onComplete(preventSubmissionOverwrite(submissionsActor, submissionId.toInt))  {`

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -162,38 +164,56 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
+        onComplete(isNotOverwrite) {
+          case Success(isNotOverwrite) =>
```
schbetsy: I found a way to remove one layer of nesting, so the logic doesn't get so deep.

Instead of having an `if`/`else` statement on the outcome of `isNotOverwrite`, you can use `case` statements like this:

```scala
onComplete(preventSubmissionOverwrite(submissionsActor, submissionId.toInt)) {
          case Success(false) =>
            val errorResponse = ErrorResponse(400, "Submission already exists", path)
            complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
          case Success(true) =>
            //upload the file
          case Failure(_) =>
            submissionsActor ! Shutdown
            val errorResponse = ErrorResponse(500, "Internal server error", path)
            complete(ToResponseMarshallable(StatusCodes.InternalServerError -> errorResponse))
        }
```

... if you use this strategy, you might want to update the name of the `preventSubmissionOverwrite` function to indicate what `true

kgudel: I really like this one! Fantastic catch here!!!!

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -162,38 +164,56 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
+            submissionsActor ! Shutdown
+            if (isNotOverwrite) {
+              processingActor ! StartUpload
+              fileUpload("file") {
+                case (metadata, byteSource) if (metadata.fileName.endsWith(".txt")) =>
+                  time {
```
schbetsy: Another thought: Is it possible to put the `time` directive at a higher level within this route, so it doesn't have to be repeated in the various cases?

kgudel: Let me check on that.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -161,39 +160,54 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
-          }
+      time {
+        val path = s"institutions/$institutionId/filings/$period/submissions/$submissionId"
+        extractExecutionContext { executor =>
+          val uploadTimestamp = Instant.now.toEpochMilli
+          val processingActor = createHmdaRawFile(system, submissionId)
+          val submissionsActor = system.actorOf(SubmissionPersistence.props(institutionId, period))
+          implicit val ec: ExecutionContext = executor
+          val fIsSubmissionOverwrite = checkSubmissionOverwrite(submissionsActor, submissionId.toInt)
+          onComplete(fIsSubmissionOverwrite) {
```
schbetsy: Good choice on this variable name. I think it's appropriately descriptive. 👍

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -161,39 +160,54 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
+                      val errorResponse = ErrorResponse(400, "Invalid File Format", path)
+                      complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
+                  }
+                case _ =>
+                  processingActor ! Shutdown
+                  val errorResponse = ErrorResponse(400, "Invalid File Format", path)
+                  complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
+              }
+            case Success(true) =>
+              {
```
schbetsy: I believe this set of braces is unnecessary.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -161,39 +160,54 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
-                  "uploaded"
-                }
-              case Failure(error) =>
-                processingActor ! Shutdown
-                log.error(error.getLocalizedMessage)
-                val errorResponse = ErrorResponse(400, "Invalid File Format", path)
-                complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
-            }
-          }
+      time {
```
schbetsy: Nice, I think moving `time` to here improves concision and readability. 

Though I'm not sure how the `time` directive interacts with code that acts asynchronously, so I'm not 100% sure this will actually wait until the upload is finished to stop timing. Worrying about that is outside the scope of this PR, but we might want to make an issue so we don't forget to check.

kgudel: So the time directive will consider the time that it takes to send the messages to persist the raw data but it is true that it won't capture the entire process of persisting, parsing, and validating the file. But I also don't believe that it should. This is just trying to capture the time that it takes from receiving a response to giving one. Timing processing will probably be its own task though.

schbetsy:
Pending some conflict resolution, LGTM

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -144,39 +144,52 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
+                  processingActor ! Shutdown
+                  val errorResponse = ErrorResponse(400, "Invalid File Format", path)
+                  complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
+              }
+            case Success(true) =>
+              val errorResponse = ErrorResponse(400, "Submission already exists", path)
+              complete(ToResponseMarshallable(StatusCodes.BadRequest -> errorResponse))
+            case Failure(_) =>
+              submissionsActor ! Shutdown
+              val errorResponse = ErrorResponse(500, "Internal server error", path)
```
schbetsy: Now that #493 merged, you can use `completeWithInternalError` for this 500 error

kgudel: Done and Done! Thanks for that function by the way.

jmarin:
Some merge conflicts, otherwise LGTM

Issue #509, opened 2016-08-17
Authorization behaves incorrectly when project is run


schbetsy:
Authorization works correctly in the test suite, but not when hitting endpoints when the project is running locally. (Haven't tested in docker yet; probably wouldn't work there either)

When hitting `/institutions`, `/institutions/12345`, `/lar/parse`, or any other endpoint, I expect a request with proper headers to authorize properly and return data. Instead, I receive this response:
```json
{
  "httpStatus": 403,
  "message": "Unauthorized Access",
  "path": ""
}
```
... which would indicate that the request was not authorized as expected. This is true for every endpoint except  except `/`, 

To reproduce:

1. get current `master`
2. run `sbt`
3. within sbt, run `reStart`
4. From Curl or Postman, send a request to `localhost:8080/institutions`. Include headers `CFPB-HMDA-Username=betsy` and `CFPB-HMDA-Institutions=12345`. (Their values don't matter; just the presence of the headers)
5. See that you receive a `403` response with `Unauthorized Access`

Notes:
- You can demonstrate that the headers are recognized by the `hmdaAuthorize` if you extract the following code and print its results. It returns `true`:
``` scala
hasHeader("CFPB-HMDA-Username", ctx) &&
        hasHeader("CFPB-HMDA-Institutions", ctx)
```

wpears:
FWIW, I'd like to stress that "any other endpoint" really means *any other endpoint*. All paths return a 403, expect the health check at the root uri.

PR #510, opened 2016-08-17
Don't run authorization code while it's broken

schbetsy:
Until we figure out #509


lgdean:
LGTM, but you may also want input from someone who's been paying more attention to that part of the code.

PR #512, opened 2016-08-17
TS Parsing and Persistence

### LINKED ISSUE #502: Parse and persist Transmittal Sheet
Once #487 gets merged, add parsing of transmittal sheet

nickgrippin:
Closes #502


persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -26,17 +26,42 @@ class HmdaFileParserSpec extends ActorSpec {
...
     }
 
     "persist parsed LARs and parsing errors" in {
       parseLars(badLines)
       probe.send(hmdaFileParser, GetState)
-      probe.expectMsg(HmdaFileParseState(5, Seq(List("Agency Code is not an Integer"))))
+      probe.expectMsg(HmdaFileParseState(6, Seq(List("Timestamp is not a Long"), List("Agency Code is not an Integer"))))
+    }
+
+    "read entire raw file" in {
```
kgudel: I'm not sure that this test is doing doing anything. These tests have all be additive, in that the what is persisted in previous tests remains remains. In the get state in this test we are expecting the same state as in the previous test, thus it looks like we are expecting that the command `ReadHmdaRawFile` has not done anything, and the test is passing.

I also believe that `ReadHmdaRawFile` requires a `submissionId` which is the persistent id for the raw data.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -26,17 +26,42 @@ class HmdaFileParserSpec extends ActorSpec {
...
+      parseTs(badLines)
+      probe.send(hmdaFileParser, GetState)
+      probe.expectMsg(HmdaFileParseState(1, Seq(List("Timestamp is not a Long"))))
+    }
+
     "persist parsed LARs" in {
       parseLars(lines)
       probe.send(hmdaFileParser, GetState)
-      probe.expectMsg(HmdaFileParseState(3, Nil))
+      probe.expectMsg(HmdaFileParseState(4, Seq(List("Timestamp is not a Long"))))
```
jmarin: Doesn't `lines` produce a valid file? (i.e. result would be `HmdaFileParseState(4, Nil)`). If not, I think it should be modified to check for the case where we parse the file successfully, both TS and LARs

nickgrippin: It does produce a valid file, which is why the size goes up by 3.  The `HmdaFileParseState` persists across tests, so the results are cumulative.  The `Seq(List("Timestamp is not a Long"))` is the error from the previous test.

This behavior could probably be changed if I restarted the actor for every test.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
nickgrippin: I'll play with this test and see if I can get it to be more meaningful.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
jmarin: Ah, I see. That makes sense, I guess I was confused by reading this in isolation from the rest of the file

lgdean: I think that confusion is a sign that it is confusing. In general, when reading (or running) a test, I would expect it to make sense, and to work, both in isolation and as part of a larger group.

I don't know whether restarting is expensive in this case. If it is expensive, then maybe there is a way to check the state of persistent things at the beginning, and then do a comparison when it's time to make assertions.

parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
```diff
@@ -51,22 +58,31 @@ object TsCsvParser {
...
   }
 
   private def convert[T](x: => T, message: String): ValidationNel[String, T] = {
-    Try(x) match {
-      case Failure(result) => message.failure.toValidationNel
-      case Success(result) => result.success
+    val t = Try(x) match {
+      case Failure(result) =>
+        println("failure " + message); message.failure.toValidationNel
+      case Success(result) => println("this is where it stops: " + result + ", " + result.getClass); val x = result.success; println("success! " + x); x
```
nickgrippin: Ok, so this is the bug, as near as I can figure:

When running the `HmdaFileParserSpec` test, it stops here on this line (at `result.success`).  However, if you run the `TsCsvParserSpec`, it passes in similar values and continues on with the test.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
nickgrippin: I could modify this to check for changes in the size of the persisted data... That way it would work in isolation, but without needing to restart the actors.

parser/src/main/scala/hmda/parser/fi/ts/TsCsvParser.scala
hkeeler: Is this an issue of which `x` is being used?  You have both the `x: => T` param, and the local `val x` variable.

hkeeler: ...seems unlikely re-reading _how_ it's failing, but seemed worth noting.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
nickgrippin: This test has been updated, and should now read the persisted data from the `HmdaRawFile`

nickgrippin:
So, biggest changes that fixed the weird bug (huge shoutout to @lgdean for helping me troubleshoot):

Removing `self ! Shutdown`.  This was firing before the Transmittal Sheet had finished parsing, causing the whole test to stop before the TS results could be persisted.

Adding `Thread.sleep(2000)`.  I don't like having to do this, but `GetState` was being processed and returned before `ReadHmdaRawFile`.  I had thought that actors would process messages in the order they were received (barring any weird changes within the actors), but for whatever reason `GetState` was returning first.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -29,6 +33,10 @@ object HmdaFileParser {
 
   case class HmdaFileParseState(size: Int = 0, parsingErrors: Seq[List[String]] = Nil) {
     def updated(event: Event): HmdaFileParseState = event match {
+      case TsParsed(ts) =>
```
lgdean: This line could be `case TsParsed(_) | LarParsed(_)` (with the obvious later case removed).

(The above assumes that TS and LAR parse results want to be represented together, not separately. Do they? I haven't thought it through.)

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -3,40 +3,87 @@ package hmda.persistence.processing
...
   val probe = TestProbe()
 
   val timestamp = Instant.now.toEpochMilli
   val lines = fiCSV.split("\n")
   val badLines = fiCSVParseError.split("\n")
 
+  var initialSize: Int = _
+  var initalList: Seq[List[String]] = _
+
+  override def beforeEach(): Unit = {
```
kgudel: I really like the you are using the before each method to make the tests no longer dependent on each other. It seems like a really clean solution. I think it might clean the code up a bit more though to make the var be the hmdaFileParser and just create a new one of those before each test. It would make it so you don't have to carry around `initialSize` and `initialList`.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -65,7 +71,23 @@ class HmdaFileParser(submissionId: String) extends PersistentActor with ActorLog
...
+        .map(line => TsCsvParser(line))
+        .map {
+          case Left(errors) => TsParsedErrors(errors)
+          case Right(ts) => TsParsed(ts)
+        }
+
+      parsedTs
+        .runForeach(pTs => self ! pTs)
+        .andThen {
+          case _ => if (!finished) finished = true else self ! Shutdown
```
jmarin: Now that I'm looking at both parsing operations, I'm not sure I like this shutdown behavior here. It's probably easier to think about the message being responsible for reading the stream of data and sending parsing commands to itself, and leave it there. And then put the responsibility of shutting down on the caller. This would allow more than one read operation per actor instance. Probably requires some testing (i.e. manual testing through the API, to make sure behavior is consistent)

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -35,8 +67,30 @@ class HmdaFileParserSpec extends ActorSpec {
...
+      val hmdaFileParser2 = createHmdaFileParser(system, "12345-2017-2")
+      val hmdaRawFile = createHmdaRawFile(system, "12345-2017-2")
+      for (line <- lines) {
+        probe.send(hmdaRawFile, AddLine(timestamp, line.toString))
+      }
+      probe.send(hmdaRawFile, GetState)
+      probe.expectMsg(HmdaRawFileState(4))
+
+      val msg = "Parsing ended for 12345-2017-2"
+      EventFilter.debug(msg, source = hmdaFileParser2.path.toString, occurrences = 1) intercept {
```
kgudel: I believe it might be better to keep the original test as well as adding this one. Just to make sure that the data has been properly loaded.

kgudel:
This code looks really good and I'm glad that you've been able to refine it so much!

Maybe this isn't terribly relevant anymore but I put together a small PR to this branch as a way of showing the ideas I had been testing out about making sure the data was loaded before shutting the actor down. The relevant commit is [here](https://github.com/nickgrippin/hmda-platform/commit/5460c7d6684f51dc9df7966d11ee5d16afd611ea) it makes a small graph as a way of putting together the transmittal and lar parsing flows. It also uses the Sink.actorRef method which, as a sink, sends whatever comes to it to an actor ref and also provides the handy method of sending a message to that actor on the complete of that stream. Even if we go to not shutting down the actor in the `readRawHmdaFile` it might still be useful to be able to send a message, possibly back to the `EventProcessor` when the stream completes and I think that a graph might be preferable for doing that because it avoids using mutable state which I think could cause problems.

jmarin:
@nickgrippin This PR looks good to me, with one comment about actor shutdown behavior. 
@kgudel I don't think implementing a Graph here is the best way to go, it's unnecessary.  Also, `Sink.actorRef` is not back pressured, so that benefit is gone.

kgudel:
@jmarin that sounds fine with me :) ! Just so I understand better do you think its unnecessary from a principle of least power principle perspective or another?

Also I took another look at the [akka docs](http://doc.akka.io/docs/akka/current/scala/stream/stream-integrations.html) and found the warning that there is no back pressure signal from the destination actor so it could over flow the actors mailbox which is certainly problematic! However, in looking at the code I'm not sure that the current implementation has any back pressure signal from the destination actor either. Is this something that we should address?

jmarin:
@nickgrippin I've put some ideas on how to move the `Shutdown` out of this actor in this [gist](https://gist.github.com/jmarin/df6e72a2ef5a4b586f8a05f41f6e302c). Let me know if this helps

kgudel:
I think that this LGTM. I really like that the tests are no longer dependent on each other! This ended up being a huge body of work but I I think that it turned out really well!!! Thanks @nickgrippin!

nickgrippin:
The current Travis failure is from an unrelated issue (#532)

Issue #334, opened 2016-05-18
API - List of all edits for a submission


awolfe76:
`/institutions/<institution>/period/<period>/submissions/<submission>/edits` (period is being used in this case to represent a time period, year or year and quarter)

`GET` - List of all edits for a given submission, grouped by edit type.
Example: _(individual files, [syntactical](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/syntactical.json), [validity](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/validity.json), [quality](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/quality.json), [macro](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/macro.json))_

PR #514, opened 2016-08-18
Validation error type

jmarin:
This PR introduces `ValidationErrorType` to edits, so that they can be classified and grouped according to their type (syntactical, validity, quality, macro), as per the [requirements](https://github.com/cfpb/hmda-platform/issues/334) of the API. This PR also includes modifying the response for single LAR validation in line with the general API, grouping errors by type in the response.


api/src/main/scala/hmda/api/model/SingleValidationErrorResult.scala
```diff
@@ -0,0 +1,10 @@
+package hmda.api.model
+
+import hmda.validation.engine.ValidationErrorsSummary
+
+case class SingleValidationErrorResult(
+  syntactical: ValidationErrorsSummary,
+  validity: ValidationErrorsSummary,
+  quality: ValidationErrorsSummary,
+  `macro`: ValidationErrorsSummary
```
kgudel: Do we want to have macro errors in this object? To my understanding they can't be run on a single Lar.

validation/src/main/scala/hmda/validation/engine/ValidationError.scala
```diff
@@ -1,3 +1,12 @@
 package hmda.validation.engine
 
-case class ValidationError(id: String, msg: String)
+sealed trait ValidationErrorType
+case object Syntactical extends ValidationErrorType
+case object Validity extends ValidationErrorType
+case object Quality extends ValidationErrorType
+case object Macro extends ValidationErrorType
+
+case class ValidationError(id: String, msg: String, errorType: ValidationErrorType)
```
kgudel: It might be good for clarity to change the names of id and msg. I believe that msg is the edit number and id is the loan id, but it took me a bit of puzzling to figure that out.

kgudel:
This LGTM! I think it will really help to break out what edits go with what types. I just had a couple comments and there are some merge conflicts. :)

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -68,6 +69,22 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
...
+    val validity = errors.filter(_.errorType == Validity).map(e => e.msg)
+    val quality = errors.filter(_.errorType == Quality).map(e => e.msg)
+    val macroErrors = errors.filter(_.errorType == Macro).map(e => e.msg)
+
+    SingleValidationErrorResult(
+      ValidationErrorsSummary(syntactical),
+      ValidationErrorsSummary(validity),
+      ValidationErrorsSummary(quality),
+      ValidationErrorsSummary(macroErrors)
+    )
```
lgdean: Another possible approach to the four filters:
```
    val errors = validationErrors.errors.groupBy(_.errorType)
    def allOfType(errorType: ValidationErrorType) = errors.getOrElse(errorType, List()).map(e => e.msg)

    SingleValidationErrorResult(
      ValidationErrorsSummary(allOfType(Syntactical)),
      ValidationErrorsSummary(allOfType(Validity)),
      ValidationErrorsSummary(allOfType(Quality)),
      ValidationErrorsSummary(allOfType(Macro))
    )
```

validation/src/main/scala/hmda/validation/engine/ValidationError.scala
lgdean: It's the loan ID for a loan/LAR, and the control number (agency code + respondent ID) for a transmittal sheet.  I agree that renaming would help clarify.

lgdean:
Yes, I also like having structure around the types of edits. Good addition, thanks!

api/src/main/scala/hmda/api/model/SingleValidationErrorResult.scala
jmarin: Yeah, good point. Will remove

Issue #504, opened 2016-08-15
Check headers to ensure API calls refer to allowed institutions


schbetsy:
The `CFPB-HMDA-Institutions` header will contain a comma-separated list of institution IDs (probably integers, see #437). We will check for its existence on all resources rooted from /institutions. Think of it as a whitelist of institutions authorized to a given user.

Return a 403 error if an attempt is made to access an institution-specific resource (/institutions/<institution_id>/*) not in the user's whitelist.

(Splitting this issue off from #441, since it requires more groundwork.)

PR #515, opened 2016-08-18
Use RejectionHandler to return 403 and 404 responses

### LINKED ISSUE #509: Authorization behaves incorrectly when project is run

schbetsy:
Closes #509 
Related to https://github.com/cfpb/hmda-platform/issues/441

See commit messages for more detail about the changes I made here.

### DIFF
--- a/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
@@ -41,16 +41,14 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
   val institutionsPath =
     path("institutions") {
       val path = "institutions"
-      hmdaAuthorize {
-        val institutionsActor = system.actorSelection("/user/institutions")
-        timedGet {
-          val fInstitutions = (institutionsActor ? GetState).mapTo[Set[Institution]]
-          onComplete(fInstitutions) {
-            case Success(institutions) =>
-              val wrappedInstitutions = institutions.map(inst => InstitutionWrapper(inst.id, inst.name, inst.status))
-              complete(ToResponseMarshallable(Institutions(wrappedInstitutions)))
-            case Failure(error) => completeWithInternalError(path, error)
-          }
+      val institutionsActor = system.actorSelection("/user/institutions")
+      timedGet {
+        val fInstitutions = (institutionsActor ? GetState).mapTo[Set[Institution]]
+        onComplete(fInstitutions) {
+          case Success(institutions) =>
+            val wrappedInstitutions = institutions.map(inst => InstitutionWrapper(inst.id, inst.name, inst.status))
+            complete(ToResponseMarshallable(Institutions(wrappedInstitutions)))
+          case Failure(error) => completeWithInternalError(path, error)
         }
       }
     }
@@ -58,26 +56,24 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
   val institutionByIdPath =
     path("institutions" / Segment) { institutionId =>
       val path = s"institutions/$institutionId"
-      hmdaAuthorize {
-        extractExecutionContext { executor =>
-          val institutionsActor = system.actorSelection("/user/institutions")
-          val filingsActor = system.actorOf(FilingPersistence.props(institutionId))
-          timedGet {
-            implicit val ec: ExecutionContext = executor
-            val fInstitutionDetails = institutionDetails(institutionId, institutionsActor, filingsActor)
-            onComplete(fInstitutionDetails) {
-              case Success(institutionDetails) =>
-                filingsActor ! Shutdown
-                if (institutionDetails.institution.name != "")
-                  complete(ToResponseMarshallable(institutionDetails))
-                else {
-                  val errorResponse = ErrorResponse(404, s"Institution $institutionId not found", path)
-                  complete(ToResponseMarshallable(StatusCodes.NotFound -> errorResponse))
-                }
-              case Failure(error) =>
-                filingsActor ! Shutdown
-                completeWithInternalError(path, error)
-            }
+      extractExecutionContext { executor =>
+        val institutionsActor = system.actorSelection("/user/institutions")
+        val filingsActor = system.actorOf(FilingPersistence.props(institutionId))
+        timedGet {
+          implicit val ec: ExecutionContext = executor
+          val fInstitutionDetails = institutionDetails(institutionId, institutionsActor, filingsActor)
+          onComplete(fInstitutionDetails) {
+            case Success(institutionDetails) =>
+              filingsActor ! Shutdown
+              if (institutionDetails.institution.name != "")
+                complete(ToResponseMarshallable(institutionDetails))
+              else {
+                val errorResponse = ErrorResponse(404, s"Institution $institutionId not found", path)
+                complete(ToResponseMarshallable(StatusCodes.NotFound -> errorResponse))
+              }
+            case Failure(error) =>
+              filingsActor ! Shutdown
+              completeWithInternalError(path, error)
           }
         }
       }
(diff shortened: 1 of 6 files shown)


schbetsy:
The failure on Travis should be the same failure that's on master. This is ready to review.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -32,11 +32,13 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
   val parseLarRoute =
     pathPrefix("lar") {
       path("parse") {
-        timedPost {
-          entity(as[String]) { s =>
-            LarCsvParser(s) match {
-              case Right(lar) => complete(ToResponseMarshallable(lar))
-              case Left(errors) => complete(errorsAsResponse(errors))
+        hmdaAuthorize {
```
hkeeler: Do we want to require authorization for `/lar/parse`?  Seems like we want anonymous access to this resource.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -46,21 +48,23 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
...
-                case "validity" => CheckValidity(lar, ValidationContext(None))
-                case "quality" => CheckQuality(lar, ValidationContext(None))
-                case _ => CheckAll(lar, ValidationContext(None))
-              }
-              onComplete((larValidation ? checkMessage).mapTo[List[ValidationError]]) {
-                case Success(xs) =>
-                  complete(ToResponseMarshallable(xs))
-                case Failure(e) =>
-                  complete(HttpResponse(StatusCodes.InternalServerError))
+        hmdaAuthorize {
```
hkeeler: Do we want to _require_ authorization for `/lar/validate`?  It seems like it should be optional.  In addition, any institution data in the header is not being passed on to the underlying validation engines.  To do so, you'd need to add `ValidationContext(institution)` instead of `ValidationContext(None)`

Also, the current header allows for multiple institutions, while these require a specific Institution to be set.  Seems like we should leave this unauthorized until we figure out how best to deal with this.

hkeeler:
I think we _should_ 403 on all resources under the top-level secure resource...in this case `/institutions`.  Without this, unauthorized clients may be able to gain information about resources they shouldn't have access to be figuring out what 403s vs. 404s?  Currently our URIs don't expose much info (just IDs), so this is fairly-anonymous data, but that may not be the case in the future.

I also think we should follow the original pattern of having a single authorization check at the top-level secure resource.  This not only removes duplicate code, but also _guarantees_ that all future sub-resources will also be protected.  Was this what was causing the original failure that spawned this issue?

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -41,39 +41,43 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
-                val errorResponse = ErrorResponse(404, s"Institution $institutionId not found", path)
-                complete(ToResponseMarshallable(StatusCodes.NotFound -> errorResponse))
-              }
-            case Failure(error) =>
-              filingsActor ! Shutdown
-              completeWithInternalError(path, error)
+      hmdaAuthorize {
+        extractExecutionContext { executor =>
+          val institutionsActor = system.actorSelection("/user/institutions")
+          val filingsActor = system.actorOf(FilingPersistence.props(institutionId))
```
hkeeler: We should be checking that the given `institutionId` is one of the ones present in `CFPB-HMDA-Institutions`.  Without this, so long as you have any value in the header, you can access any resource.  Same goes for all sub-resources of `/institutions/<institution_id>`.

schbetsy:
Okay, if the desired behavior is that everything in the `/institutions` namespace returns a `403` for unauthorized requests, that shouldn't be hard to implement. That strategy makes sense to me. I assumed that we wanted nonexistent routes to always 404 (I didn't realize the requirements were ambiguous until you mentioned it).

As for authorizing requests to `/lar/parse` and `/lar/validate`, I was following the requirements I read in #441. If we don't want to protect those endpoints, it's easy to remove the check.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
schbetsy: Yep, that's part of a separate issue: https://github.com/cfpb/hmda-platform/issues/504. It was too much to implement all in one go.

hkeeler: OK, cool.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
schbetsy: Everything `ValidationContext`-related is not in scope of this PR. Please make a separate issue for that.

schbetsy: > the current header allows for multiple institutions, while these require a specific Institution to be set

Not sure what you mean by this. Could you clarify?

hkeeler:
>As for authorizing requests to `/lar/parse` and `/lar/validate`, I was following the requirements I read in #441. If we don't want to protect those endpoints, it's easy to remove the check.

Good point.  The original _We will check for it's existence by all but the root (`/`) resource_ statement isn't quite true.  I wasn't thinking of the pre-existing resources that don't start with `/institutions`.  Sorry about that.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
hkeeler: Yes, I'll create a separate issue for this.

schbetsy:
🚧  Under Construction, while I implement the strategy we discussed above 🚧

api/src/main/scala/hmda/api/http/LarHttpApi.scala
hkeeler: Created https://github.com/cfpb/hmda-platform/issues/523 for this.

api/src/test/scala/hmda/api/http/LarHttpApiSpec.scala
```diff
@@ -97,22 +97,31 @@ class LarHttpApiSpec extends WordSpec with MustMatchers with ScalatestRouteTest
...
-      responseAs[ErrorResponse] mustBe ErrorResponse(403, "Unauthorized Access", "")
+    "allow requests without 'CFPB-HMDA-Username' header" in {
+      Post("/lar/validate", lar).addHeader(institutionsHeader) ~> larRoutes ~> check {
+        status mustEqual StatusCodes.OK
+        responseAs[List[ValidationError]] mustBe Nil
+      }
+    }
+
+    "not handle routes that aren't defined in this API" in {
+      // Request the endpoint without username header (but with other headers)
```
hkeeler: This comment seems to conflict with the line below.  Doesn't `getWithCfpbHeaders()` add the `CFPB-HMDA-Username` header?  I may also be confused as I'm just basing what `getWithCfpbHeaders()` does other tests.

schbetsy: Good call. Removed in 0e26011

schbetsy:
Commemorating a conversation from Mattermost, in case anyone looks here for answers:


>hans.keeler
>6:15 PM
>Is handled mustBe false a 404...or a lower-level Akka HTTP construct saying it couldn't figure out what to do with it.  I'd look it up, but I know you're trying to get out of here ASAP.
>6:16 PM
>I'm also fine merging as-is and figuring it out on my own if you need to run. :)
>
>betsylorton
>6:43 PM
>Yep, isHandled mustBe false means that the institutions API shouldn't do anything with that request. Once the request gets through all our routes unhandled, the RejectionHandler takes charge and returns a 404
>
>hans.keeler
>6:51 PM
>OK.  Cool.
>
>betsylorton
>6:51 PM
>That plus `rejections mustBe List()` is how the internet recommends testing that something 404s (though it's kind of exposing implementation details of akka http). If the rejection list is empty (and no route has handled the request), that's how it indicates to itself that it should be a 404

hkeeler:
Looks good.  I'm going ahead with the merge.

The current Travis build failure (https://travis-ci.org/cfpb/hmda-platform/builds/153425661) is not related to this PR.  Issue https://github.com/cfpb/hmda-platform/issues/524 is for finding/fixing it.

PR #516, opened 2016-08-18
Parse and Validate endpoint

### LINKED ISSUE #501: Parse and validate single lar in one request
Combine parse and validate functionality into one HTTP call (keeping the separate endpoints, as an additional feature). Input should be a single LAR text line in CSV format, output is a list of errors (if any)

nickgrippin:
Not sure if there's a way to just combine the two routes (maybe with `redirect`?), without just copying and reorganizing the code.  

Closes #501


kgudel:
This LGTM I like the idea of converting the validation route into a function to avoid code duplication!

nickgrippin:
I don't know what's going on with codecov, but I think my tests have good coverage for this PR.

PR #520, opened 2016-08-18
Update API Documentation

### LINKED ISSUE #507: Update API doc
#493 was merged so we should update the [API documentation](https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md) to show the headers that are required.

_Came across this while trying to run docker-compose setup. It runs but I get a `Failed to load resource: the server responded with a status of 403 (Forbidden)` on the API call, which is good._

schbetsy:
Closes #507 
Also removes trailing whitespace.

I updated all instances of `fid` in API responses to to `institutionId`. Is that correct?

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -58,12 +58,12 @@ All endpoints below require two headers (see "Authorization" section below for m
       "filings": [
         {
           "id": "2017",
-          "fid": "12345",
+          "institutionId": "12345",
           "status": "not-started"
         },
         {
           "id": "2016",
-          "fid": "12345",
+          "institutionId": "12345",
           "status": "completed"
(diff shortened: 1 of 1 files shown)


awolfe76:
👍

schbetsy:
Don't merge this one until we've settled the design of the API authorization. We might decide to change which endpoints are protected by auth, in which case this documentation will need to change.

schbetsy:
@awolfe76 I'll need you to take another pass at review. In #515, we decided that `/lar/parse` and `/lar/validate` don't need auth, so I updated the docs to reflect that.

awolfe76:
LGTM

PR #521, opened 2016-08-18
Return Institution ID as a String in the API

### LINKED ISSUE #518: /institutions update - id is number and should be a string
__tl;dr:  `institution.id` should be a string.__

When running the docker-compose setup I couldn't get the institution status to render. When looking at the json response for `/institutions/[id]`:
```json
{
  "institution": {
    "id": 0,
    "name": "Bank 0",
    "status": "active"
  },
  "filings": [{
    "period": "2017",
    "institutionId": "0",
    "status": "not-started"
  }, {
    "period": "2016",
    "institutionId": "0",
    "status": "completed"
  }]
}
```
You can see that the `institution.id` is a number and the `filings.institutionID` is a string. __Both should be strings.__ On the front-end we were filling an array based on `filing.institutionId === institution.id` which doesn't work with that response. 

The response for just `/institutions` also has the `id` as a number:
```json
{
  "institutions": [{
    "id": 0,
    "name": "Bank 0",
    "status": "active"
  }]
}
```
which I'm sure make sense ... same endpoint.

kgudel:
Closes #518

### DIFF
--- a/api/src/main/scala/hmda/api/model/InstitutionWrapper.scala
+++ b/api/src/main/scala/hmda/api/model/InstitutionWrapper.scala
@@ -3,7 +3,7 @@ package hmda.api.model
 import hmda.model.institution.InstitutionStatus
 
 case class InstitutionWrapper(
-  id: Int,
+  id: String,
   name: String,
   status: InstitutionStatus
 )
--- a/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
@@ -46,7 +46,7 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
         val fInstitutions = (institutionsActor ? GetState).mapTo[Set[Institution]]
         onComplete(fInstitutions) {
           case Success(institutions) =>
-            val wrappedInstitutions = institutions.map(inst => InstitutionWrapper(inst.id, inst.name, inst.status))
+            val wrappedInstitutions = institutions.map(inst => InstitutionWrapper(inst.id.toString, inst.name, inst.status))
             complete(ToResponseMarshallable(Institutions(wrappedInstitutions)))
           case Failure(error) => completeWithInternalError(path, error)
         }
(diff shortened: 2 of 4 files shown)


awolfe76:
Cool, I'll try this out first thing in the morning.

awolfe76:
@kgudel - this works 💥, but it looks like there is a build failure.

Also, I noticed another small thing. From the API doc we should be seeing:
```json
"filings": [
  {
    "id": "2017",
    "institutionId": "12345",
    "status": "not-started"
  },
  {
    "id": "2016",
    "institutionId": "12345",
    "status": "completed"
  }
]
```
Notice it has an `id` for the filings. What is being returned now is a `period`:
```json
{
  "period": "2017",
  "institutionId": "0",
  "status": "not-started"
}
```

We can either add that fix to this PR, I imagine its quick, or create a new issue. Let me know.

kgudel:
So I believe that returning `period` is actually more accurate than returning `ID`. Would it be acceptable to change the API docs to reflect the naming change?

awolfe76:
Not against the idea at all but it would require not only the API doc update but also a change on the front-end, which isn't a huge deal just pointing it out.

@jmarin

api/src/main/scala/hmda/api/model/InstitutionWrapper.scala
```diff
@@ -3,7 +3,7 @@ package hmda.api.model
 import hmda.model.institution.InstitutionStatus
 
 case class InstitutionWrapper(
-  id: Int,
+  id: String,
```
jmarin: I see how this fixes the problem of integrating with the UI, but the bigger problem imho is why the institution ID is an Int. For 2018 we have [LEI](http://www.sifma.org/issues/operations-and-technology/legal-entity-identifier/overview/) which is certainly a String of alphanumeric characters

jmarin:
I would suggest not changing the naming on filing ids / periods as part of this PR, since it's a different concern than what this fix is trying to do.

awolfe76:
I created #525 to handle the `period` to `id`.

api/src/main/scala/hmda/api/model/InstitutionWrapper.scala
awolfe76: Isn't that was this is doing, updating it to be a String?

jmarin: Sort of. This object is a wrapper around what really gets persisted (so that we don't send out all details about an institution, that are not needed at the API level), see the the lines above where id.id.toString is called. My concern is that the internal [Institution](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/institution/Institution.scala#L7) object doesn't model its id as a String. I'm ok with merging this one and ticketing this as a different issue.

awolfe76:
The goal of this PR was to get the front-end and the /institutions API synced up ... and it does. But, as mentioned above #525 was created and another issue will need created to handle [this comment](https://github.com/cfpb/hmda-platform/pull/521/files#r75500204):

> Sort of. This object is a wrapper around what really gets persisted (so that we don't send out all details about an institution, that are not needed at the API level), see the the lines above where id.id.toString is called. My concern is that the internal Institution object doesn't model its id as a String. I'm ok with merging this one and ticketing this as a different issue.

Merging.

jmarin:
I've created #526 for changing institution id to string. I'm good with the changes in this PR

PR #529, opened 2016-08-19
Fixes case-sensitive headers bug

### LINKED ISSUE #527: Auth headers should be case insensitive
Both node, the browser, and nginx are auto lowercasing `CFPB-HMDA-Insitutions` and `CFPB-HMDA-Username`. Via RFC 7230, headers should be case-insensitive, so there should be a case insensitive comparison here.

Relevant RFC 7230 section:
```
3.2.  Header Fields

   Each header field consists of a case-insensitive field name followed
   by a colon (":"), optional leading whitespace, the field value, and
   optional trailing whitespace.
```

hkeeler:
Fixes #527


hkeeler:
This should be fixed now.  I haven't yet tested it in the full Docker Compose setup yet, where this issue was discovered.  Working on that now.

PR #530, opened 2016-08-22
S020 cleanup: remove duplicate object and agency hard-coding

lgdean:
This duplication-removal became possible due to the work in #94 and #438 (and associated edits), but the PR for the latter was large enough that I kept this part separate.

I am particularly interested in feedback on the type signatures.


lgdean:
Note: the codecov decrease comes from removing a file that was well-covered, not because any coverage was lost (or un-covered lines newly added) in code that remains.

validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
```diff
@@ -6,19 +6,19 @@ import hmda.validation.context.ValidationContext
...
   def checkSyntactical(lar: LoanApplicationRegister, ctx: ValidationContext): LarValidation = {
-    val checksToRun: List[EditCheck[LoanApplicationRegister]] = List(
+    val checksToRun: List[EditCheck[_ >: LoanApplicationRegister]] = List(
       S010,
       S020,
       S025.inContext(ctx),
       S205
     )
-    val checks = checksToRun.map(check(_, lar, lar.loan.id, Syntactical))
+    val checks = checksToRun.map(check[LoanApplicationRegister](_, lar, lar.loan.id, Syntactical))
```
jmarin: I think adding the type here is not necessary? In any case, the other `check` methods don't have it, it would probably make sense to be consistent.

jmarin:
Interesting use of types. I haven't followed the hierarchy in detail, but I played a bit with this. I _think_ you can achieve the same by just making `EditCheck` contravariant (from a compiler point of view, it works and tests pass; the logic might be flawed though, so correct me if I'm wrong). Just throwing it out there as food for thought, see [here](https://github.com/jmarin/hmda-platform/tree/s020-contravariant)

validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
lgdean: With the code as it is right now, removing that type leads to a compile error.
You're right that `TsSyntacticalEngine` seems not to need it, though, even though it has similar types involved. I wonder what the relevant difference is! Definitely worth exploring.

jmarin: Interesting, and a bit weird. The other LAR engines didn't require it? I guess that's because they run "normal" edit checks? (guessing, haven't checked)

lgdean:
Yes, contravariance makes sense here; the logic is good. An `EditCheck[HasControlNumber]` can be used in place of an `EditCheck[LoanApplicationRegister]` in any context. The former is definitely a subtype of the latter. Good catch!

validation/src/main/scala/hmda/validation/engine/lar/syntactical/LarSyntacticalEngine.scala
lgdean: Yeah, the other LAR engines have only `EditCheck[LoanApplicationRegister]`s.

As for the difference between `LarSyntacticalEngine` and its `Ts` counterpart, it turns out to be subtle, and have to do with the declared types.  In the LAR version, that line of code is in a method that is declared to return a specific type. If you change the `checkSyntactical` return type to `ValidationNel[ValidationError, _ >: LoanApplicationRegister]`, then it compiles fine either way: with or without the explicit type in the `check` call.

Meanwhile in the TS version, those lines of code are in the context of a larger method that doesn't constrain its types (and specifically, the return type of the `validateAll` call) as much, and so things work out fine.

Contravariance in `EditCheck` is a better answer, but this question was educational and interesting.  I have a somewhat better sense for how the type inference is working here, now.

jmarin:
LGTM

kgudel:
LGTM also thanks @lgdean and @jmarin for teaching me a great deal about scala's type system through this PR!

PR #533, opened 2016-08-24
Submissions/latest

### LINKED ISSUE #531: 500 error on /submissions/latest API call
When running the integration of back-end and front-end using docker-compose we noticed a `500 internal server error` on the `/institutions/0/filings/2017/submissions/latest` API call.

@wpears pointed out that this is likely due to the `/latest` not being accepted by the back-end.

~~For now, this issue is a question of whether the front-end needs to update to not use the `/latest` or if the back-end wants to update to accept it.~~

Back-end will implement. 

@cfpb/hmda-team-write

kgudel:
Creates an endpoint to return the latest submission for some institution and period.

This implementation checks the `submission.id` of the returned submission to make sure that it is not 0. This is meant to be a check to insure that the requested submission actually exists as even if the institution or filing do not exist, or that combo has no submissions, a get latest request will always return a `submission(0, Created)` .

Closes #531


awolfe76:
I'll run through this one in the morning. Thanks @kgudel!

jmarin:
LGTM

awolfe76:
Running this locally with docker-compose I see the response. So, this is good.

But I do see an error in coming from the UI:
`TypeError: Cannot read property 'code' of undefined(…)`

I _think_ this is because the front-end has [defined a submission](https://github.com/cfpb/hmda-platform-ui/blob/968f5635a85274a218c08b34ee5247979e3c07a2/server/json/submissions.json#L4) as:
```json
{
  "id": 1,
  "status": {
    "code": 1,
    "message": ''
  }
}
```

And here we're returning (which is what was defined in the [API doc](https://github.com/cfpb/hmda-platform/blob/05b73023b8cab476656d20f043d996ae18e52b66/Documents/API.md)):
```json
{
  "id": 3,
  "submissionStatus": "created"
}
```

Also on the `POST` I'm seeing a `500 Internal Server Error`. Could be related??? I'm uploading a dummy .csv, is there a better file that I can use?

@wpears may have some input there too.

jmarin:
@awolfe76 can you share the .csv that you are trying to upload?

awolfe76:
I can, but it's not at all formatted in the correct way. To be honest, until you asked I hadn't even opened it. It's an old file from the QM tool. Still want it?

I was just assuming that I would get something back in any case ... even a "file isn't formatted" properly.

jmarin:
Well, you've discovered a bug / missing feature in that endpoint. It's worth creating a test that makes it fail in the same way, and fixing it with an error message. I guess the file in question is irrelevant. Could you create a ticket with the error you are seeing and the `json` response you would expect from that request?. Thanks

awolfe76:
#535 created.

awolfe76:
I think we should update the format of the response and the documentation in this PR to match whats mentioned above:
```json
{
  "id": 1,
  "status": {
    "code": 1,
    "message": ""
  }
}
```

We can figure out in #535 if we should allow the `POST` to `.../latest`.

awolfe76:
@kgudel - I'm getting a 404 now. But I _think_ it makes sense. We just have mismatched data.

For example, calling `/institutions/0/filings/2016/submissions/latest` is a 404 but that's because bank 0 doesn't have a 2016 submission.

 Same for `/institutions/1/filings/2017/submissions/latest`.

So it seems to be working but it would be good to have some of this data matched up to be able to see a good response.

Can you double check my thinking?

kgudel:
@awolfe76 , So yes that is expected behavior. The only call that should not return a 404 is:

`/institutions/0/filings/2017/submissions/latest`

I am happy to update the data though!

awolfe76:
Right, that is good sign! But we should also be able to see some successes.

README.md
```diff
@@ -79,7 +79,7 @@ Confirm that the platform is up and running by browsing to http://localhost:8080
 3. To build JVM artifacts (the default, includes all projects), from the sbt prompt:
 
 ```shell
-> assembly
+> clean assembly
```
awolfe76: I think this should be `sbt clean assembly`

kgudel: This is fine as just `clean assembly` because an early step has you start up `sbt` but I fixed the one on 108. Good catch!

awolfe76:
👍  from me.

awolfe76:
@jmarin this is working and i think ready for merge. Didn't know if you wanted to give it a quick look again?

PR #540, opened 2016-08-29
Fix Institution Error Messages

### LINKED ISSUE #467: "Institution X not found" error messages from API
Currently, some API endpoints that take `institutionId`, when sent an invalid institution ID, return an inaccurate/misleading error message. 

Example: `/institutions/xxxxx/filings/2017`
returns error message: "2017 filing not found for institution xxxxx"
A more accurate error would be: "Institution xxxxx not found" or similar.

Another example: `/institutions/xxxxx/filings/2017/submissions`

nickgrippin:
Closes #467


jmarin:
LGTM

kgudel:
All of the logic LGTM. I think that it would be useful to good to add a couple of new tests that show off the old responses. So, currently there are tests for when the route fails because the institution doesn't exist but it might be nice to add some for when the filing or submission doesn't exist. Other than that I like it!

PR #550, opened 2016-09-01
Added 405 error route to /latest

### LINKED ISSUE #535: 500 error on POST to /institutions/0/filings/2017/submissions/latest
Based on the conversation from https://github.com/cfpb/hmda-platform/pull/533, when running the setup locally with docker-compose the `POST` to `/institutions/0/filings/2017/submissions/latest` returns a `500 Internal Server Error`.

Also, from the [API docs in the PR](https://github.com/cfpb/hmda-platform/pull/533/files#diff-853a8f52dcda495b60f31ee64a0fe01d) it doesn't look like there is a `POST` available to the `/institutions/0/filings/2017/submissions/latest` endpoint. This is probably ok. On the front-end we should be using `/institutions/<institution>/filings/<period>/submissions/<submissionId>` to upload the file.

---

#### Screenshot with request highlighted:
screen shot 2016-08-25 at 10 40 51 am

---

#### File upload (changed to .txt to allow for upload, you can change it back to .csv)
[test.txt](https://github.com/cfpb/hmda-platform/files/437300/test.txt)
_this file is not formatted correctly, but a response should be returned_

---

#### Possible response?
You can see in the UI that we rendered "Error uploading file". So the 500 error was taken care of. Should there be a better way to handle this?

nickgrippin:
Thanks to some amazing sleuthing from @schbetsy, I found out that `POST`ing to the /latest endpoint will redirect to the `uploadPath` route (as opposed to the `submissionLatestPath`, where we want the request to go).  In the `uploadPath` route, the "latest" String will fail to parse as an Integer, returning a 500 error.  

I decided to hard-code a `405` route instead of changing any path signatures to avoid forcing any changes on the front-end.

Closes #535

### DIFF
--- a/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
@@ -169,7 +169,11 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
             submissionsActor ! Shutdown
             completeWithInternalError(path, error)
         }
-      }
+      } ~
+        timedPost {
+          val errorResponse = ErrorResponse(405, "Method not allowed", path)
+          complete(ToResponseMarshallable(StatusCodes.MethodNotAllowed -> errorResponse))
+        }
     }
 
   val uploadPath =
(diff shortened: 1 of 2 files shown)


schbetsy:
LGTM

hkeeler:
Wow!  _Amazing sleuthing_ indeed.  I stared at it for a bit, but it all seemed OK.

Now that I get what's going on, I think there are some other solutions that wouldn't involve an explicit `405` having to be constructed.  The simplest one I can think of would be to switch the `uploadPath` route to using a different [`PathMatcher`](http://doc.akka.io/docs/akka/2.4.9/java/http/routing-dsl/directives/path-directives.html#Path_Matchers) than the generic `Segment`.  If you switched it to a `intValue` for instance, I think it would reject both `/latest` (since it has no `POST`) and `/<submission_id>` (since "latest" isn't an int).

nickgrippin:
@hkeeler: That seems like it should work, I'll see if I can switch that over right away.  Thanks for pointing out the PathMatchers!

PR #551, opened 2016-09-02
Simplify validation states, add doc

### LINKED ISSUE #546: Simplify edit states
The states that __really__ need to be captured in the application for validation are: 

* Validating
* Validated
* Validated with errors

jmarin:
Closes #546


Documents/submission-status.md
```diff
@@ -0,0 +1,22 @@
...
+* `code`: `Integer`
+* `message`: `String`
+
+In order to track the status of a filing for a financial institution, the following states are captured by the backend:
+
+* `1`: `Created`. The filing has been created and is ready to accept data
+* `2`: `Uploading`. Data is currently being uploaded to the system
+* `3`: `Uploaded`. Data has finished uploading, is stored in the `HMDA Platform` and is ready to be checked
+* `4`: `Parsing`. The submitted information is being checked for parsing errors according to the [HMDA File Specification]().
+* `5`: `Parsed`. Filed data is conformant and is ready to be validated by the rule engine
```
awolfe76: Do we need a `ParsedWithErrors`? Or is that being handled by the `-1: Failed` with a decent `message`?

awolfe76:
For the states, right now we have a way to tell when the IRS report has been generated and verified too, see https://github.com/cfpb/hmda-platform-ui/blob/master/submission-states.md.

Should we keep that? This happens when state `7`, from your states doc, is reached. (We can do it earlier but it really only matters when the file is valid, all errors are fixed. Right @gou?) 

This could happen in state `7`. For example, if the current state is `8`, with errors, the front-end would do a `POST` with the last update to the edits and if that's successful the back-end could automatically start to generate the IRS report and the new state  returned would be `7`. The front-end would then know that we could request the IRS report.

We would still need a state for the IRS being verified.

Also, like in the above example, each `POST` to fix an error would need to return the `edits: []` along with the status so we know what edits to render. Doing this would let us keep that state simple, we would continue to see a status of `8` along with the edits. When a `7` is reached we would know its good and request the IRS report.

One last thing, do you think it would be better to have state `7` be `ValidatedWithErrors` and `8` be `Validated`? Just to keep in the sequence in order. In that case it's possible that we would skip `7` in some cases.

Documents/submission-status.md
jmarin: Yes, good point. This needs to be implemented in the persistence model. Created #552 for this, will update this documentation and the submission states with it in this PR

schbetsy:
I can't comment on the business logic (i.e. which states should indicate what, etc), but the code LGTM.

schbetsy:
I can't comment on the business logic (i.e. which states should indicate what, etc), but the code LGTM.

PR #553, opened 2016-09-02
Change in-memory persistence test library

### LINKED ISSUE #543: Explore changing in-memory persistence implementation
The current in-memory Akka Persistence implementation is poorly documented and probably lacks some capabilities (i.e. `PersistentQuery`). Replace with a more capable implementation,  i.e. https://github.com/dnvriend/akka-persistence-inmemory

jmarin:
Closes #543

### DIFF
--- a/persistence/src/test/resources/application.conf
+++ b/persistence/src/test/resources/application.conf
@@ -3,7 +3,6 @@ akka {
   actor.warn-about-java-serializer-usage = off
 }
 
-akka.persistence.journal.plugin = "akka.persistence.journal.inmem"
-akka.persistence.snapshot-store.plugin = "akka.persistence.snapshot-store.local"
-akka.persistence.snapshot-store.local.dir = "target/snapshots"
+akka.persistence.journal.plugin = "inmemory-journal"
+akka.persistence.snapshot-store.plugin = "inmemory-snapshot-store"
 akka.loggers = [akka.testkit.TestEventListener]
--- a/persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
+++ b/persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
@@ -20,14 +20,7 @@ class HmdaFileParserSpec extends ActorSpec with BeforeAndAfterEach with HmdaFile
     ActorSystem(
       "test-system",
       ConfigFactory.parseString(
-        """
-          | akka.loggers = ["akka.testkit.TestEventListener"]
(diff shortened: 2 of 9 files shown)


persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -24,9 +24,9 @@ class HmdaFileParserSpec extends ActorSpec with BeforeAndAfterEach with HmdaFile
           | akka.loggers = ["akka.testkit.TestEventListener"]
           | akka.loglevel = DEBUG
           | akka.stdout-loglevel = "OFF"
-          | akka.persistence.journal.plugin = "akka.persistence.journal.inmem"
-          | akka.persistence.snapshot-store.plugin = "akka.persistence.snapshot-store.local"
-          | akka.persistence.snapshot-store.local.dir = "target/snapshots"
+          | akka.persistence.journal.plugin = "inmemory-journal"
```
schbetsy: It looks like this block of config code is repeated in `HmdaFileValidatorSpec`. Could we extract the test config settings into a file?

persistence/src/test/resources/application.conf
```diff
@@ -3,7 +3,6 @@ akka {
   actor.warn-about-java-serializer-usage = off
 }
 
-akka.persistence.journal.plugin = "akka.persistence.journal.inmem"
-akka.persistence.snapshot-store.plugin = "akka.persistence.snapshot-store.local"
-akka.persistence.snapshot-store.local.dir = "target/snapshots"
+akka.persistence.journal.plugin = "inmemory-journal"
```
schbetsy: This looks very similar to the test config, but the test config has `akka.persistence.query.journal.id = "inmemory-read-journal"` and this one doesn't. Why?

jmarin: These are common settings for all persistent actor tests. The `akka.persistence.query.journal.id = "inmemory-read-journal"` is only needed for those that have to read a stream of events from the journal

schbetsy:
Here's my current high-level understanding of what we're doing with persistence. Let me know if/what I'm misunderstanding:
- This PR switches the persistence library we're using for our automated tests (from akka built-in to dnvriend's implementation)
- We're currently still using leveldb when we run the project (locally or otherwise). This PR doesn't affect that.

jmarin:
@schbetsy Yes, both those statements are correct.

schbetsy:
LGTM

PR #555, opened 2016-09-02
Require Institution Header to match institutionId of requests

### LINKED ISSUE #504: Check headers to ensure API calls refer to allowed institutions

schbetsy:
Closes #504 
Also includes minor cleanup of `InstitutionsHttpApiSpec`

As usual, I'm looking for feedback on my approach, idiomatic use of Scala, etc.

### DIFF
--- a/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
+++ b/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
@@ -16,12 +16,16 @@ trait HmdaCustomDirectives extends ApiErrorProtocol {
     RejectionHandler.newBuilder()
       .handle {
         case AuthorizationFailedRejection =>
-          val errorResponse = ErrorResponse(403, "Unauthorized Access", "")
-          complete(ToResponseMarshallable(StatusCodes.Forbidden -> errorResponse))
+          extractUri { uri =>
+            val errorResponse = ErrorResponse(403, "Unauthorized Access", uri.toString)
+            complete(ToResponseMarshallable(StatusCodes.Forbidden -> errorResponse))
+          }
       }
       .handleNotFound {
-        val errorResponse = ErrorResponse(404, "Not Found", "")
-        complete(ToResponseMarshallable(StatusCodes.NotFound -> errorResponse))
+        extractUri { uri =>
+          val errorResponse = ErrorResponse(404, "Not Found", uri.toString)
+          complete(ToResponseMarshallable(StatusCodes.NotFound -> errorResponse))
+        }
       }
       .result()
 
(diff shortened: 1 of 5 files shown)


schbetsy:
This will require the front-end to update their headers [here](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/api.js#L9). For testing, that line could be

```
'CFPB-HMDA-Institutions': '0,1,2 ,3'
```

jmarin:
Overall, LGTM. One question though. Some of this code is using the Java API (i.e. [here](https://github.com/cfpb/hmda-platform/pull/555/files#diff-95ceb1992c3531f923df7c469b47af7aR62)). Have you considered implementing some of this logic by leveraging the built in header directives? (http://doc.akka.io/docs/akka/2.4.9/scala/http/routing-dsl/directives/header-directives/index.html)

schbetsy:
I chose the `authorize` directive over the `header`-related directives when I originally implemented the header checking, but I don't remember exactly why. Thanks for the tip, I'll revisit that decision and see whether the built-in directives would be better.

api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
```diff
@@ -17,25 +16,32 @@ trait HmdaCustomDirectives extends ApiErrorProtocol {
     RejectionHandler.newBuilder()
       .handle {
         case AuthorizationFailedRejection =>
-          val errorResponse = ErrorResponse(403, "Unauthorized Access", "")
-          complete(ToResponseMarshallable(StatusCodes.Forbidden -> errorResponse))
+          extractUri { uri =>
+            val errorResponse = ErrorResponse(403, "Unauthorized Access", uri.toString)
```
hkeeler: This is going to result in the full URI being put into the `path` attribute, while the rest of the routes look like they're using a manually constructed path string, and passing that into all errors.  I think we should make this consistent.

I personally prefer having the full URI, and I also think we could replace all of the manual path building with a single directive or error handler.  That seems out outside the scope of this PR, though.  Perhaps for now, we just do a `uri.path.toString` for now to make it consistent?

schbetsy: +1 for consistency. Updated in 13c1c5f84c4903390f71c3c356f52900a6192178.

hkeeler:
Looks good.  Merging...

PR #558, opened 2016-09-08
Filter /institutions by user auth

### LINKED ISSUE #331: API - List of institutions for a user

### LINKED ISSUE #412: Filter list of institutions by user
See #331 
The list of institutions returned needs to be filtered according to the user that is logged in. The approach that has been discussed at this point is that the list of institutions to be returned will be part of a request header

schbetsy:
Closes #412 (I believe. See comment in issue)
Closes #331

### DIFF
--- a/persistence/src/test/scala/hmda/persistence/institutions/InstitutionPersistenceSpec.scala
+++ b/persistence/src/test/scala/hmda/persistence/institutions/InstitutionPersistenceSpec.scala
@@ -2,8 +2,8 @@ package hmda.persistence.institutions
 
 import akka.testkit.{ EventFilter, TestProbe }
 import hmda.actor.test.ActorSpec
-import hmda.model.institution.Agency.CFPB
-import hmda.model.institution.ExternalIdType.{ FederalTaxId, RssdId }
+import hmda.model.institution.Agency.{ CFPB, FDIC }
+import hmda.model.institution.ExternalIdType.{ FdicCertNo, FederalTaxId, RssdId }
 import hmda.model.institution.{ ExternalId, Institution }
 import hmda.model.institution.InstitutionStatus.Active
 import hmda.model.institution.InstitutionType.Bank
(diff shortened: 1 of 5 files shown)


schbetsy:
CodeCov fails because the failure case of this endpoint (and all of our other endpoints), where it returns a 500 error, isn't covered with tests.

nickgrippin:
LGTM

persistence/src/test/scala/hmda/persistence/institutions/InstitutionPersistenceSpec.scala
```diff
@@ -34,6 +34,24 @@ class InstitutionPersistenceSpec extends ActorSpec {
...
       probe.expectMsg(modified)
     }
+
+    "return a set of institutions matching a list of ids" in {
+      // Setup: Persist demo institutions
+      val i1 = Institution(71, "CFPBank", Set(ExternalId("99-1234567", FederalTaxId)), CFPB, Bank, hasParent = true, Active)
+      val i2 = Institution(72, "FRBank", Set(ExternalId("654321", RssdId)), CFPB, Bank, hasParent = true, Active)
+      val i3 = Institution(73, "MLBank", Set(ExternalId("externalTest0", FdicCertNo)), FDIC, Bank, hasParent = true, Active)
+      for (institution <- List(i1, i2, i3)) {
+        probe.send(institutionsActor, CreateInstitution(institution))
```
kgudel: Is there a reason for using these specific institutions over the test institutions?

schbetsy: My reason (tbd whether it's a good reason :) ) was that `DemoData.testInstitutions` is a `Set`, and I don't know of a good way to get a reference to an individual member of a set. That is, I wanted to call `GetInstitutionsById` with just 2 of the 4 institutions, but I couldn't find a good way to set up my expectation (line 49) if I did that.

kgudel: You could always convert it to a list with `DemoData.testInstitutions.toList` or another type that is easier to work with :)!

schbetsy: OH YEAH. So I just tried your suggestion (good suggestion, by the way!), and I remembered the real reason why I had to create my own data with new ids. In the collection of persisted Institutions, there are already multiple records with id 0, 1, 2, and 3. So when I use `GetInstitutionsById`, I get multiple responses for each `id`. The fact that we allow duplicate IDs is a separate issue, so I'll create a GH issue for it.

PR #559, opened 2016-09-08
Make InstitutionId a String, for future compatibility

### LINKED ISSUE #526: Institution id modeled as Int
Currently [Institution](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/institution/Institution.scala#L7) has its id modeled as an `Integer`. In 2018 we are adopting [LEI](https://en.wikipedia.org/wiki/Legal_Entity_Identifier), which is certainly alphanumeric. If we use LEI as an institution identifier in the future our model would have to be changed to have id as a String

schbetsy:
Closes #526

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
+++ b/persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
@@ -70,7 +70,7 @@ class InstitutionPersistence extends HmdaPersistentActor {
       }
 
     case GetInstitutionById(institutionId) =>
-      val institution = state.institutions.find(x => x.id.toString == institutionId).getOrElse(Institution(-1, "", Set(), CFPB, Bank, hasParent = false, Inactive))
+      val institution = state.institutions.find(x => x.id.toString == institutionId).getOrElse(Institution("", "", Set(), CFPB, Bank, hasParent = false, Inactive))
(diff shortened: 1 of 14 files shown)


persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -69,7 +69,7 @@ class InstitutionPersistence extends HmdaPersistentActor {
       }
 
     case GetInstitutionById(institutionId) =>
-      val institution = state.institutions.find(x => x.id.toString == institutionId).getOrElse(Institution(-1, "", Set(), CFPB, Bank, hasParent = false, Inactive))
+      val institution = state.institutions.find(x => x.id.toString == institutionId).getOrElse(Institution("", "", Set(), CFPB, Bank, hasParent = false, Inactive))
```
schbetsy: When `GetInstitutionById` doesn't find anything (in the `getOrElse`), I chose to make the `id` of the empty institution an empty string (`""`). Is this what we want?

jmarin: In the absence of more information (i.e. requirements), I think this is a good default

nickgrippin:
LGTM!

Issue #97, opened 2016-03-08
Validation (2017): S013 (Timestamp)


hkeeler:
**Edit #:** S013
**Type:** Syntactical
**Category:** Transmittal Sheet (only)
**Column:** Timestamp

Timestamp must be later than timestamp on database.

Additional resources:
- [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
- [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
- [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

kgudel:
This is already [partially implemented](https://github.com/cfpb/hmda-platform/blob/master/validation/src/main/scala/hmda/validation/rules/ts/syntactical/S013.scala), it checks that the time stamp is greater than some other time which is passed in as a variable. However, it is not clear what that database variable should reference, and it does not seem to align with the transactional model of our back end.

schbetsy:
Marking as "blocked", since we haven't decided how to proceed. Need to decide what to use as the "database", or--if this edit is actually irrelevant with the new architecture--whether the edit can be removed.

lesgou:
This edit is no longer relevant due to changes in the architecture.

Issue #96, opened 2016-03-08
Validation (2017): S100 (Activity Year)


hkeeler:
**Edit #:** S100
**Type:** Syntactical
**Category:** Transmittal Sheet (only)
**Column:** Activity Year

Activity year must = year being processed (= 2017).

Additional resources:
* [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
* [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
* [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

kgudel:
How do we want to tell the edit what year the activity year should be? This links with #131.

jmarin:
I personally believe this edit doesn't make much sense given how the architecture for `hmda-platform` has been designed

schbetsy:
I was under the impression that all of the edits we are writing are specific to the 2017 filing year (since the file format will change drastically in 2018, and therefore the parser, model, and edits will change drastically with it). If that were true, it wouldn't be a problem to hard-code 2017 into the edit.

Is this not the case? How much of the system will we try to re-use in 2018?

jmarin:
We should try to reuse as much as possible. There will be some changes in data model / parser, etc. but many of the edits will most likely be the same (though this is still pending, and no decision has been made as of yet).

schbetsy:
Fair. I'm glad we'll try to re-use.

In any case, this edit still seems valuable. If a FI tries to submit 2016 data to the 2017 system, we'd want to stop them. Why do you say this edit doesn't make sense?

PR #560, opened 2016-09-08
Validate and persist TS

### LINKED ISSUE #542: Validate and persist transmittal sheet
Once #541 has been merged, add validation and persistence of Transmittal Sheet to `HmdaFileValidator`

nickgrippin:
Closes #542


schbetsy:
What's the deal with `S100` and `S013`? Will they be added back in at some point?

nickgrippin:
Yes, both tickets (#96 and #97) are still open.  The code was added in without knowing where we would get the additional data, but it was blocking progress on this ticket, so we're just removing the code stumps for now.

validation/src/test/scala/hmda/validation/engine/ts/syntactical/TsSyntacticalEngineSpec.scala
```diff
@@ -98,21 +82,15 @@ class TsSyntacticalEngineSpec extends PropSpec with PropertyChecks with MustMatc
   protected def failGenTs(badTs: Gen[TransmittalSheet]): Assertion = {
     badTs.sample match {
       case Some(x) =>
-        val fValidated = checkSyntactical(x, ValidationContext(None))
-        whenReady(fValidated) { validated =>
-          validated.isFailure mustBe true
-        }
+        checkSyntactical(x, ValidationContext(None)).isFailure mustBe true
```
schbetsy: I believe you can use `mustBe a[Failure]` here instead, which would give more sensible error messages.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -72,6 +79,16 @@ class HmdaFileValidator(submissionId: String) extends HmdaPersistentActor with L
       val validationStarted = ValidationStarted(submissionId)
       publishEvent(validationStarted)
       events(parserPersistenceId)
+        .filter(x => x.isInstanceOf[TsParsed])
```
schbetsy: Since there's only one TransmittalSheet per submission, should this be `find` instead of `filter`? (I haven't researched), but it seems like it would be less computationally intensive.

nickgrippin: There's unfortunately no `find` method for a `Flow`.  I think `take(1)` could work, but I'm not sure if we can guarantee the order of the events.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -72,6 +79,16 @@ class HmdaFileValidator(submissionId: String) extends HmdaPersistentActor with L
...
+        .map(e => e.asInstanceOf[TsParsed].ts)
+        .map(ts => validateTs(ts, ctx).toEither)
+        .map {
+          case Right(ts) => ts
+          case Left(errors) => ValidationErrors(errors.list.toList)
+        }
+        .runWith(Sink.actorRef(self, NotUsed))
```
schbetsy: I'm not as familiar with akka streams, so I'm testing my understanding here. I think this line is saying that once the stream is processed and all Transmittal Sheets have been validated, this actor isn't listening for the "completed" event. That's okay because line 99 of this file sends another "completed" message that we do listen for.

... is that right?

nickgrippin: So, I think this line points the outcome of this Flow at itself (it will send itself all the messages coming out, so `ts` will be sent right back here and be caught by `receiveCommand` case further down for the type `TransmittalSheet`).  The `NotUsed` isn't what the actor is listening for, it's the event the actor sends when the Sink has been completed.

lgdean:
If I'd known it was blocking anything, I would have created (and done) #563 sooner. Oh well. (To be clear, I knew it was slowing things down a bit, as I kept seeing the same changes happening in code that was almost the same but couldn't quite be refactored, because of those two weird cases. I just hadn't realized it was an actual blocker.) Sorry about that!

schbetsy:
LGTM

Issue #439, opened 2016-07-07
Add location, branch, and CRA data to Institution data model


hkeeler:
In order to unblock edit [Q595](https://github.com/cfpb/hmda-platform/issues/213), we must add Metropolitan Statistical Area / Metropolitan Division (MSA/MD) data to Institution.  Edit [Q030](https://github.com/cfpb/hmda-platform/issues/248) requires this data, but also requires branch and Community Reinvestment Act (CRA) data.

If branch or CRA data proves difficult to implement, we could split this issue in two, and deal with just MSA/MD data first.

This issue is dependent on https://github.com/cfpb/hmda-platform/issues/438.

Kibrael:
CRA data can be found here:
https://www.ffiec.gov/cra/reporter16.htm

jmarin:
Closing, since Q595 is not being implemented

Issue #248, opened 2016-03-08
Validation (2017): Q030 (MSA/MD,State,County,Census Tract)


hkeeler:
**Edit #:** Q030
**Type:** Micro
**Category:** Quality Edit Q030 is generated by the FFIEC and is included with the micro edits for each appropriate loan
**Column:** MSA/MD,State,County,Census Tract

If action taken type = 1, 2, 3, 4, 5, or 6; and if the HMDA respondent is required to report MSA/MD, state, county, census tract, then MSA/MD, state, county, census tract should equal a valid combination and not NA.

Additional resources:
- [FFIEC's 2016 HMDA File Spec](https://www.ffiec.gov/hmda/pdf/spec2016.pdf)
- [FFIEC's 2016 HMDA Edits](https://www.ffiec.gov/hmda/pdf/edit2016.pdf)
- [CFPB's 2017 HMDA File Spec](http://www.consumerfinance.gov/hmda/static/for-filers/2017/2017-HMDA-File-Specifications.pdf)

lesgou:
Updated language for clarification purposes.

hkeeler:
How do we know when a _HMDA respondent is **required** to report MSA/MD_?  Is this a check against  panel data, and if they are in a MSA/MD there, they also need to report it in the HMDA filing, or is it based on some other requirements?

lesgou:
This is related to branches located in an msa/md listed on panel. There is an update on this so I will report back with more information.

lesgou:
@hkeeler @lgdean @Kibrael 

Required institutions
1) If the filer is a CRA reporter
2) If the filer has a branch located in the MSA/MD that is being reported

hkeeler:
@lesgou, so this is what qualifies them as _required_ in the edit's description:

>...if the HMDA respondent is **required** to report MSA/MD, state, county, census tract...

lesgou:
@hkeeler Yes. That is what qualifies as required. I made a slight update to #2.

hkeeler:
Cool.  Thanks @lesgou.

hkeeler:
Labeling this as _Blocked_ since it requires institution data (branch and CRA).

Also, this is labeled as a _Macro_ edit.  Shouldn't it be _Quality_?  I realize this is a non-standard edit.  In the description it read:

>**Category:** Quality Edit Q030 is generated by the FFIEC and is included with the micro edits for each appropriate loan

I think _micro_ should be _macro_? :)

[...]

PR #564, opened 2016-09-12
Q030

lgdean:
addresses #248
addresses #439

I'd particularly like feedback on how to make this code readable enough to convince someone that it does the right things (and all the right things, particularly all the relevant test cases).  I wasn't sure how to structure it, to get all the contrasting pairs somewhat close to each other.

(For the prod code, I may also create a version that uses a bunch of `case`s, to see whether it ends up more readable/convincing.)


model/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -10,6 +10,7 @@ case class Institution(
     agency: Agency,
     institutionType: InstitutionType,
     hasParent: Boolean,
+    cra: Boolean = false, // TODO do we have this info when creating the institution? if so, then don't default here.
```
hkeeler: I think the answer is _yes_.  CRA is a separate data source, but  we should be able to merge it with NIC whenever creating/updating Institution records.

schbetsy:
The work you've done to make the `Q030` Spec readable is really slick. I like the nesting and the different implementations of `mustPass` and `mustFail`.

hkeeler:
For any other reviewers, I found the output of the test to be really useful when trying to understand all the branches of logic here.

@lesgou, can you take a peek at this?  I can walk you through it if it doesn't make sense, or interpret what's going on under-the-hood in the code.

```
[info] Q030Spec:
[info] Q030
[info]   when action taken type is 7 or 8 (preapproval)
[info]   - must pass
[info]   when action taken is 1-6
[info]     when institution is NOT a CRA reporter
[info]       when all 4 geography fields are NA
[info]       - must pass
[info]       when county is present and small
[info]         when tract is NA
[info]         - must pass
[info]         when tract is present (and matches)
[info]         - must fail
[info]     when institution is a CRA reporter
[info]       when all 4 geography fields are NA
[info]       - must fail
[info]       when state or county is NA (no matter what else is true)
[info]       - must fail
[info]       when county is present and small
[info]         when tract is NA
[info]         - must pass
[info]         when tract is present (and matches)
[info]         - must pass
[info]     when property MSA/MD is present
[info]       when tract is NA
[info]         when MSA/MD, state and county match
[info]         - must pass
[info]         when MSA/MD, state and county do not match
[info]         - must fail
[info]         when county is large
[info]         - must fail
[info]       when tract is present
[info]         when MSA/MD, state, county, and tract match
[info]         - must pass
[info]         when MSA/MD, state, county, and tract do not match
[info]         - must fail
[info]     when property MSA/MD is NA
[info]       when tract is NA
[info]         when state and county match
[info]         - must pass
[info]         when state and county do not match
[info]         - must fail
[info]         when county is large
[info]         - must fail
[info]       when tract is present
[info]         when state, county, and tract match
[info]         - must pass
[info]         when state, county, and tract do not match
[info]         - must fail
[info]     when state is present but county is NA
[info]     - must fail when CRA is true
[info]     - must fail when CRA is false
[info]     when county is present but state is NA
[info]     - must fail when CRA is true
[info]     - must fail when CRA is false
```

lesgou:
:+1: : wow, I think we've got it!

hkeeler:
I think we're good here.  Merging...

PR #567, opened 2016-09-12
Add activity year

### LINKED ISSUE #563: add year to ValidationContext
Add filing year to `ValidationContext`, now that it exists.  That way, edits that need it can get it that way.

We could perhaps also put the "timestamp" needed for S013 in there.  That one is puzzling because it's still not clear (at least to me) what it means, but the meaning is probably separate from the question of how to pass the information into the edit.  In any case, if it's blocked, then it doesn't need to be part of this ticket.  But if it's easy to go ahead and include, then we might as well.

nickgrippin:
Addresses #563

### DIFF
--- a/api/src/main/scala/hmda/api/http/LarHttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/LarHttpApi.scala
@@ -82,11 +82,12 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
 
   def validateRoute(lar: LoanApplicationRegister, checkType: String): scaladsl.server.Route = {
     val larValidation = system.actorSelection("/user/larValidation")
+    val vContext = ValidationContext(None, None)
     val checkMessage = checkType match {
-      case "syntactical" => CheckSyntactical(lar, ValidationContext(None, None))
-      case "validity" => CheckValidity(lar, ValidationContext(None, None))
-      case "quality" => CheckQuality(lar, ValidationContext(None, None))
-      case _ => CheckAll(lar, ValidationContext(None, None))
+      case "syntactical" => CheckSyntactical(lar, vContext)
+      case "validity" => CheckValidity(lar, vContext)
+      case "quality" => CheckQuality(lar, vContext)
+      case _ => CheckAll(lar, vContext)
     }
     onComplete((larValidation ? checkMessage).mapTo[ValidationErrors]) {
       case Success(xs) =>
--- a/api/src/main/scala/hmda/api/http/LarHttpApi.scala
+++ b/api/src/main/scala/hmda/api/http/LarHttpApi.scala
@@ -83,10 +83,10 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
   def validateRoute(lar: LoanApplicationRegister, checkType: String): scaladsl.server.Route = {
     val larValidation = system.actorSelection("/user/larValidation")
     val checkMessage = checkType match {
-      case "syntactical" => CheckSyntactical(lar, ValidationContext(None))
-      case "validity" => CheckValidity(lar, ValidationContext(None))
-      case "quality" => CheckQuality(lar, ValidationContext(None))
-      case _ => CheckAll(lar, ValidationContext(None))
+      case "syntactical" => CheckSyntactical(lar, ValidationContext(None, None))
+      case "validity" => CheckValidity(lar, ValidationContext(None, None))
+      case "quality" => CheckQuality(lar, ValidationContext(None, None))
+      case _ => CheckAll(lar, ValidationContext(None, None))
     }
     onComplete((larValidation ? checkMessage).mapTo[ValidationErrors]) {
       case Success(xs) =>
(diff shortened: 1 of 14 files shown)


lgdean:
I'd claim that this PR does not close #563 until the code has an actual client (such as Q022, or the recently-deleted edits). Otherwise, it seems hard to know whether it's working.
(That's not necessarily an argument against merging it; just a request to change "closes" to "addresses" or the like.)

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -75,7 +75,7 @@ class HmdaFileValidator(submissionId: String) extends HmdaPersistentActor with T
   override def receiveCommand: Receive = {
 
     case BeginValidation =>
-      val ctx = ValidationContext(None)
+      val ctx = ValidationContext(None, None)
```
lgdean: Does anything prevent us from using the actual year from the submission? If not, then let's do it. If so, then that seems like something we need to solve before calling this issue done.

If it can't be done, then it's arguably a fatal flaw in this approach. (In which case, thank you for doing the work to discover it!) Or maybe there's a work-around.
If it can be done, but something is blocking it at the moment, then I vote for hard-coding it here, perhaps with a big blaring comment explaining what's up.
It it can be done now, see above. :)

validation/src/main/scala/hmda/validation/context/ValidationContext.scala
```diff
@@ -2,4 +2,4 @@ package hmda.validation.context
 
 import hmda.model.institution.Institution
 
-case class ValidationContext(institution: Option[Institution])
+case class ValidationContext(institution: Option[Institution], filingYear: Option[Int])
```
lgdean: This line also seems like a potential site for hard-coding 2017.
Currently not my first choice, but I could imagine arguments that would sway me. And it is certainly better (more central) than the current location(s), such as line 13 of `LarQualityEngine`. (That 2017 is semantically the same year as this one, right?)

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -83,10 +83,10 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
...
     val larValidation = system.actorSelection("/user/larValidation")
     val checkMessage = checkType match {
-      case "syntactical" => CheckSyntactical(lar, ValidationContext(None))
-      case "validity" => CheckValidity(lar, ValidationContext(None))
-      case "quality" => CheckQuality(lar, ValidationContext(None))
-      case _ => CheckAll(lar, ValidationContext(None))
+      case "syntactical" => CheckSyntactical(lar, ValidationContext(None, None))
+      case "validity" => CheckValidity(lar, ValidationContext(None, None))
+      case "quality" => CheckQuality(lar, ValidationContext(None, None))
+      case _ => CheckAll(lar, ValidationContext(None, None))
```
lgdean: Please extract a val for this `ValidationContext`. Thanks! And sorry I missed it when reviewing previous PRs; I realize you didn't introduce the duplication here.

lgdean: Would it make sense to say `Some(2017)` for this context? I'm open to arguments for `None`. Just want to make sure it's a conscious decision.

Related: what is the intended behavior of edits when there's not a year? Are they just skipped? (Or, should that situation be impossible?)

nickgrippin:
I'd like to avoid having to hard-code the year as much as possible.  Otherwise, it seems like we're setting ourself up for yearly maintenance where we have to go through and change all the hard-coded values.  There should be a way for us to determine activity year programatically.  

I'm ok with changing this ticket to `addresses` instead of `closes`, since the next ticket on my list is the S100.

lgdean:
Yes, I also would prefer to avoid hard-coding as much as possible. The hard-coding (if any) would be a temporary approach, replacing the hard-coding that currently happens in `LarQualityEngine` and the like (and was until recently happening in code calling S100 and S013, until those were temporarily deleted).

And ideally, maybe hard-coding not needed at all, as noted in [this comment](https://github.com/cfpb/hmda-platform/pull/567/files#r78474569).

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
nickgrippin: So would this be using the `<period>` from the submission path?

lgdean: Yup, that period is the filing/activity year, as I understand it.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -75,7 +77,7 @@ class HmdaFileValidator(submissionId: String) extends HmdaPersistentActor with T
   override def receiveCommand: Receive = {
 
     case BeginValidation =>
-      val ctx = ValidationContext(None, None)
+      val ctx = ValidationContext(None, getFilingPeriod(submissionId))
```
kgudel: With #562 merged in you can just get this from, I believe, `submissionId.period`!

kgudel:
LGTM!!! I like taking the activity year from the SubmissionId in the validator. A very clean implementation!

jmarin:
LGTM

PR #568, opened 2016-09-12
Improve Authorization Specs

schbetsy:
This is a just a quick update. I implemented a better strategy for testing authorization in the institutions API. Now, I'm using Route.seal, which means the RejectionHandler will take over when a request is rejected and return the correct status code.


jmarin:
LGTM

kgudel:
LGTM!

PR #572, opened 2016-09-14
Supervisor

jmarin:
This PR introduces custom actor supervision to our project. Instead of creating all actors directly from the `guardian`, they are now children of `HmdaSupervisor`. This new actor is responsible for managing the lifecycle of all HMDA actors at this stage, with default supervision strategy (a custom strategy might be implemented in the future as the need arises). All actors are now created through the supervisor, and the persistent actors stop themselves if they haven't seen activity in a specific period of time.
In addition to this, some tests were updated to resolve a memory leak


api/src/main/scala/hmda/api/HmdaApi.scala
```diff
@@ -36,9 +41,27 @@ object HmdaApi
 
   //Start up API Actors
 
-  createSingleLarValidator(system)
+  val supervisor = HmdaSupervisor.createSupervisor(system)
+
+  (supervisor ? FindActorByName(SingleLarValidation.name))
```
nickgrippin: I remember you had mentioned that the actors would time out after a certain period of inactivity.  If that's the case, would we still want to start up all the actors right away (when they might be killed later), or should we wait until they receive activity before spinning them up?

api/src/main/scala/hmda/api/HmdaApi.scala
```diff
@@ -36,9 +41,27 @@ object HmdaApi
...
+      log.info(s"Started event processor at ${actor.path}")
+    }
+
+  (supervisor ? FindActorByName(InstitutionPersistence.name))
+    .mapTo[ActorRef]
+    .map { actor =>
+      log.info(s"Started institutions at ${actor.path}")
+    }
+
   createInstitutions(system)
```
kgudel: It looks like we are already creating the Institutions actor above through the supervisor so this one might be just a forgot to remove?

api/src/main/scala/hmda/api/HmdaApi.scala
```diff
@@ -36,9 +41,27 @@ object HmdaApi
 
   //Start up API Actors
 
-  createSingleLarValidator(system)
+  val supervisor = HmdaSupervisor.createSupervisor(system)
+
+  (supervisor ? FindActorByName(SingleLarValidation.name))
+    .mapTo[ActorRef]
+    .map { actor =>
+      log.info(s"Started validator at ${actor.path}")
```
kgudel: We are already logging something very similar when the actors start up in the `HmdaActor` class: `"Actor started at ${self.path}"`. Do we want to log this twice?

persistence/src/main/scala/hmda/persistence/HmdaSupervisor.scala
```diff
@@ -0,0 +1,104 @@
...
+}
+
+class HmdaSupervisor extends HmdaActor {
+
+  import HmdaSupervisor._
+
+  var hmdaPersistentActors = Map.empty[String, ActorRef]
+
+  override def receive: Receive = {
+    case FindActorByName(name) =>
```
kgudel: I think I would find this a bit more readable if it were called something like `StartActorByName` instead of `Find`. And the same for `FindFilings`, `FindSubmissions`, and `FindProcessingActor`. If other people like this naming schema more readable I'm totally fine with it though :)

kgudel: Oh I read the code slightly wrong the first time! This actually finds or starts!!! Maybe something like `FindOrStartActorByName`, its a bit long but I think it captures more fully what it does.

api/src/main/scala/hmda/api/HmdaApi.scala
jmarin: You are right about `InstitutionPersistence`, it will time out and shut itself down if it doesn't see activity. The persistent actors are the ones where I've added the behavior to shut themselves down. The other actors (event processor and single lar validation) are started at the beginning and remain in memory as long as the application is running. I'm going to push an update to better reflect how `InstitutionPersistence` should be instantiated in the API

kgudel: It looks like that is only true of persistent actors because that logic is contained in `HmdaPersistentActor`, but I think that this might still be a problem for the `InstitutionPersistence` clause.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -40,7 +42,7 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
   val institutionsPath =
     path("institutions") {
       val path = "institutions"
-      val institutionsActor = system.actorSelection("/user/institutions")
+      val institutionsActor = system.actorSelection("/user/supervisor/institutions")
```
kgudel: Switching this and all similar calls to `system.actorSelection("/user/supervisor/institutions")` to a `FindActorByName` might be more in keeping with the established pattern of accessing persistent actors through the supervisor. It would also solve the possible issue that Nick brought up about the `InstitutionActor` itself switching off.

api/src/main/scala/hmda/api/HmdaApi.scala
jmarin: Yes, as I stated in my previous comment, the logic for self shutdown is only present in persistent actors. We will have thousands of them, hence this logic. By the way, this will change when we go to a clustered application, but I don't want to get ahead of myself. Also as I said in my reply to @nickgrippin, I'm making some changes to the way `InstitutionPersistence` is being instantiated. Will push additional commits soon

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
jmarin: See comment above, this is being addressed in response to @nickgrippin comment

api/src/main/scala/hmda/api/HmdaApi.scala
kgudel: Sorry, I hadn't updated my github page and didn't see your previous comment. Sounds good!

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
kgudel: Got it 👍

api/src/main/scala/hmda/api/HmdaApi.scala
```diff
@@ -36,9 +41,25 @@ object HmdaApi
...
+    }
+
+  (supervisor ? FindActorByName(LocalHmdaEventProcessor.name))
+    .mapTo[ActorRef]
+    .map { actor =>
+      log.info(s"Started event processor at ${actor.path}")
+    }
+
+  (supervisor ? FindActorByName(InstitutionPersistence.name))
```
kgudel: Now that we are using `FindActorByName` for this actor whenever it is being used in the api do we still want to start a copy of it here?

jmarin: I am starting it here on purpose, since `/institutions` is the entry point. It doesn't hurt to have it running, since this actor is unique for the whole application, as opposed to institution specific ones, where there will be thousands of copies being instantiated.

api/src/main/scala/hmda/api/http/InstitutionsHttpApi.scala
```diff
@@ -40,16 +42,23 @@ trait InstitutionsHttpApi extends InstitutionProtocol with ApiErrorProtocol with
...
-          val fInstitutions = (institutionsActor ? GetInstitutionsById(ids)).mapTo[Set[Institution]]
-          onComplete(fInstitutions) {
-            case Success(institutions) =>
-              val wrappedInstitutions = institutions.map(inst => InstitutionWrapper(inst.id.toString, inst.name, inst.status))
-              complete(ToResponseMarshallable(Institutions(wrappedInstitutions)))
-            case Failure(error) => completeWithInternalError(path, error)
+          extractExecutionContext { executor =>
+            implicit val ec: ExecutionContext = executor
+            val ids = institutionIdsFromHeader(ctx)
+            val supervisor = system.actorSelection("/user/supervisor")
```
nickgrippin: This actor is used in every endpoint... Would it be better to have a single supervisor declaration at the top of this file?

persistence/src/main/scala/hmda/persistence/HmdaSupervisor.scala
```diff
@@ -0,0 +1,104 @@
...
+    case FindFilings(name, id) =>
+      sender() ! findFilings(name, id)
+
+    case FindSubmissions(name, institutionId, period) =>
+      sender() ! findSubmissions(name, institutionId, period)
+
+    case FindProcessingActor(name, submissionId) =>
+      sender() ! findProcessingActor(name, submissionId)
+
+    case Terminated(ref) =>
```
nickgrippin: Is there any restart logic?  If a child actor fails, how do we deal with the failure?

api/src/main/scala/hmda/api/HmdaApi.scala
kgudel: I just realized that it also has to be started for the demo data to load. So that sounds good! +1

persistence/src/main/scala/hmda/persistence/HmdaSupervisor.scala
jmarin: Or, to put it another way, there __is__ built in behavior for when a child actor fails, and we're taking that behavior to be ok for now (in most cases a restart will happen)

nickgrippin:
LGTM!  Definitely a very informative PR, learned a lot throughout the review.

jmarin:
👍 Glad to hear

PR #575, opened 2016-09-16
Submission status update

### LINKED ISSUE #497: Update submission status
Update submission status in `SubmissionPersistence` according to which step the processing pipeline is in (i.e. parsing, validated, etc.)

jmarin:
Closes #497

### DIFF
--- a/persistence/src/test/scala/hmda/persistence/processing/LocalHmdaEventProcessorSpec.scala
+++ b/persistence/src/test/scala/hmda/persistence/processing/LocalHmdaEventProcessorSpec.scala
@@ -1,15 +1,25 @@
 package hmda.api.processing
 
-import akka.actor.ActorSystem
+import akka.actor.{ ActorRef, ActorSystem }
 import akka.testkit.{ EventFilter, TestProbe }
+import akka.pattern.ask
+import akka.util.Timeout
+
+import scala.concurrent._
+import scala.concurrent.duration._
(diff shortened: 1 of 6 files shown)


persistence/src/test/scala/hmda/persistence/processing/LocalHmdaEventProcessorSpec.scala
```diff
@@ -71,13 +100,32 @@ class LocalHmdaEventProcessorSpec extends ActorSpec {
     "process validation completed from event stream" in {
       val msg = s"validation completed for submission $submissionId"
       checkEventStreamMessage(msg, ValidationCompleted(submissionId))
+      checkSubmissionStatus(Validated)
     }
   }
 
+  def checkSubmissionStatus(status: SubmissionStatus): Assertion = {
+    //TODO: find a way to avoid having to sleep here
+    Thread.sleep(500)
```
nickgrippin: Is there actually a way to avoid this race condition without sleeping?  It seems unavoidable, and we've had to use this same approach in other tests.

jmarin: I'm pretty sure there is a way, but I haven't thought about it in detail. As you say, we're using it across the board

nickgrippin:
Looks good!

PR #577, opened 2016-09-20
Use supervisor to access SingleLarValidation actor in LarHttpApi

### LINKED ISSUE #576: SingleLarValidation parseAndValidate not working
This [line](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/LarHttpApi.scala#L84) is pointing to the wrong ActorSelection (it should run now under the supervisor). `LarHttpApiSpec` will probably need to be updated

kgudel:
This pr also moves `completeWithInternalError` to `HmdaCustomDirectives` so that it can be used in `LarHttpApi` to make the api error response more consistent across routes.

Closes #576


api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -80,20 +79,27 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
...
-        complete(ToResponseMarshallable(aggregateErrors(xs)))
-      case Failure(e) =>
-        complete(HttpResponse(StatusCodes.InternalServerError))
+    val validationErrors = for {
+      larValidation <- fLarValidation
+      ve <- (larValidation ? checkMessage).mapTo[ValidationErrors]
+    } yield ve
+
+    onComplete(validationErrors) {
+      case Success(validationErrors) =>
```
nickgrippin: I'd rename this variable to something else, otherwise it's confusing whether this is the same `validationErrors` as in the `onComplete` call.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
```diff
@@ -80,20 +79,27 @@ trait LarHttpApi extends LarProtocol with ValidationResultProtocol with HmdaCust
       }
     }
 
-  def validateRoute(lar: LoanApplicationRegister, checkType: String): scaladsl.server.Route = {
-    val larValidation = system.actorSelection("/user/larValidation")
+  def validateRoute(lar: LoanApplicationRegister, checkType: String, path: String) = {
```
nickgrippin: What is the `path` variable used for?

api/src/main/scala/hmda/api/http/LarHttpApi.scala
kgudel: Good catch! I renamed the first `validationErrors` to `fValidationErrors`, to signify that it was a future.

api/src/main/scala/hmda/api/http/LarHttpApi.scala
kgudel: In writing it, it was meant to go in `completeWithInternalError`, I had forgotten to feed it into that function but I have resolved that.

nickgrippin:
LGTM!

schbetsy:
LGTM too. I'll merge after Travis turns green. 🍏

PR #581, opened 2016-09-21
Parser Events for Invalid Files

### LINKED ISSUE #552: Add ParsedWithErrors to Parser persistence
When a parsing fails, `HmdaFileParser` should create a `ParsedWithErrors` event and publish it to the event stream

schbetsy:
Closes #552.

Also, small updates to specs `HmdaFileParserSpec` and `LocalHmdaEventProcessorSpec`.

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
@@ -51,6 +51,7 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor wit
   import HmdaFileParser._
 
   var state = HmdaFileParseState()
+  var encounteredParsingErrors: Boolean = false
 
   override def updateState(event: Event): Unit = {
     state = state.updated(event)
@@ -62,7 +63,6 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor wit
 
     case ReadHmdaRawFile(persistenceId) =>
       publishEvent(ParsingStarted(submissionId))
-      var encounteredParsingErrors: Boolean = false
 
       val parsedTs = events(persistenceId)
         .map { case LineAdded(_, data) => data }
--- a/persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
@@ -25,6 +25,7 @@ object HmdaFileParser {
   case class CompleteParsingWithErrors(submissionId: SubmissionId) extends Command
   case class ParsingStarted(submissionId: SubmissionId) extends Event
   case class ParsingCompleted(submissionId: SubmissionId) extends Event
+  case class ParsingCompletedWithErrors(submissionId: SubmissionId) extends Event
 
   def props(id: SubmissionId): Props = Props(new HmdaFileParser(id))
 
@@ -125,6 +126,7 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor wit
 
     case CompleteParsingWithErrors =>
       log.debug(s"Parsing completed for $submissionId, errors found")
+      publishEvent(ParsingCompletedWithErrors(submissionId))
 
     case GetState =>
       sender() ! state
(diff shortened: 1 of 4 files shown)


persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -60,12 +61,16 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor wit
 
     case ReadHmdaRawFile(persistenceId) =>
       publishEvent(ParsingStarted(submissionId))
+      var encounteredParsingErrors: Boolean = false
```
schbetsy: Using this boolean variable, `encounteredParsingErrors`, seemed like the easiest/most efficient way to decide whether or not the file had any parsing errors. 

Is it bad form to use a `var`? If so, what would be better?

jmarin: Actors are one of the few places where using `var` is safe, since they are designed to keep state. I think it would probably be safest to put this variable at the top level, so it gets initialized (i.e. to false) and it's then changed accordingly as messages flow through the `receive` method

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -115,6 +123,9 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor wit
       log.debug(s"Parsing completed for $submissionId")
       publishEvent(ParsingCompleted(submissionId))
 
+    case CompleteParsingWithErrors =>
+      log.debug(s"Parsing completed for $submissionId, errors found")
```
jmarin: Are we missing a `publishEvent(ParsedWithErrors)` here so that it gets picked up by the event processor?

schbetsy: See what I wrote a bit about this in the original issue, #552. I'm sending the event `ParsingCompletedWithErrors`, because that follows the pattern that the `HmdaFileValidator` established [here](https://github.com/cfpb/hmda-platform/blob/master/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala#L134).

If the event should really be called `ParsedWithErrors`, should I change the `HmdaFileValidator` to follow the same pattern?

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
schbetsy: Thanks for the feedback. Could you say more about why having `encounteredParsingErrors` is safer as a class-level variable (that's what I think you're suggesting; correct me if I'm wrong), rather than as a method-level variable (which it is now)? That info will help me understand how to use vars in the future.

jmarin: It's more for convention than anything else, most Akka actor code you will see will follow that pattern. The idea is it's easy to spot where the state changes will happen, and there will be an initial state at actor instantiation before any messages are received (this could be necessary in some cases; I think it doesn't necessarily matter in this particular case)

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
jmarin: Sorry, probably used the wrong variable name. What I meant is that under `case CompleteParsingWithErrors` I see the log line, but no event is fired to the event stream. It looks like it's missing something like `publishEvent(ParsingCompletedWithErrors(submissionId))`. Is this being done elsewhere? (I couldn't find it on first look)

schbetsy: Ohh, I see. Can you see it [here](https://github.com/schbetsy/hmda-platform/blob/e53e31bace3d99143d1066a9c498b8a960b58b42/persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala#L129)? I think there might be a bug in GitHub, because I added that line, but sometimes it disappears in the PR Review view.

jmarin: Yeah, I see it. Weird, as you mention it wasn't visible from the PR review page

persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
schbetsy: Okay. I made the change in 45e9f59791fdac8d6b0764a4e7dd3525fa637db5.

schbetsy:
I believe one of our intermittent test failures was happening because we used the same `submissionId` in several test cases, when we were testing the log messages that contain `submissionId`. The latest commit on this branch should fix one instance where that was happening.

jmarin:
LGTM

PR #585, opened 2016-09-22
refactor InstitutionsHttpApiSpec

### LINKED ISSUE #584: Refactor InstitutionsHttpApiSpec
Break up big test file into smaller test files to reflect the structure as changed by https://github.com/cfpb/hmda-platform/pull/583

kgudel:
Closes #584

### DIFF
--- a/api/src/test/scala/hmda/api/http/InstitutionSpec.scala
+++ b/api/src/test/scala/hmda/api/http/InstitutionSpec.scala
@@ -1,0 +1,40 @@
+package hmda.api.http
+
+import java.io.File
+
+import akka.http.scaladsl.model.{ ContentTypes, HttpEntity, Multipart }
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import com.typesafe.config.ConfigFactory
+import hmda.api.RequestHeaderUtils
+import hmda.persistence.HmdaSupervisor
+import hmda.persistence.HmdaSupervisor.FindActorByName
+import hmda.persistence.demo.DemoData
+import hmda.persistence.institutions.InstitutionPersistence
+import org.iq80.leveldb.util.FileUtils
+import org.scalatest.{ BeforeAndAfterAll, MustMatchers, WordSpec }
+
+trait InstitutionSpec extends WordSpec with MustMatchers with BeforeAndAfterAll
+    with ScalatestRouteTest with RequestHeaderUtils with InstitutionsHttpApi {
+
+  override def beforeAll(): Unit = {
+    val supervisor = HmdaSupervisor.createSupervisor(system)
+    supervisor ! FindActorByName(InstitutionPersistence.name)
+    DemoData.loadTestData(system)
+    super.beforeAll()
+  }
+
+  override def afterAll(): Unit = {
+    system.terminate()
+    val config = ConfigFactory.load()
+    val snapshotStore = new File(config.getString("akka.persistence.snapshot-store.local.dir"))
+    FileUtils.deleteRecursively(snapshotStore)
+  }
+
+  def multiPartFile(contents: String, fileName: String) =
+    Multipart.FormData(Multipart.FormData.BodyPart.Strict(
+      "file",
+      HttpEntity(ContentTypes.`text/plain(UTF-8)`, contents),
+      Map("filename" -> fileName)
+    ))
+
+}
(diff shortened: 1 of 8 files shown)


api/src/test/scala/hmda/api/http/InstitutionSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  override def afterAll(): Unit = {
+    system.terminate()
+    val config = ConfigFactory.load()
+    val snapshotStore = new File(config.getString("akka.persistence.snapshot-store.local.dir"))
+    FileUtils.deleteRecursively(snapshotStore)
+  }
+
+  def multiPartFile(contents: String, fileName: String) =
```
jmarin: Shouldn't this function be included in the `UploadPathSpec` class instead?

kgudel: This function is used both in `UploadPathSpec` but also in `InstitutionHttpApiSpec`, that's why I put it in `InstitutionSpec`.

api/src/test/scala/hmda/api/http/institutions/SubmissionPathsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import hmda.api.http.InstitutionSpec
+import hmda.api.model.{ErrorResponse, SubmissionStatusWrapper, SubmissionWrapper}
+import hmda.model.fi._
+import hmda.persistence.demo.DemoData
+
+import scala.concurrent.duration._
+
+class SubmissionPathsSpec extends InstitutionSpec {
+
+  override val log: LoggingAdapter = NoLogging
```
schbetsy: Could it work to put these lines (initializing `log`, `timeout`, and `ec`) in the parent spec? If so, that would be nice, to reduce duplication.

api/src/test/scala/hmda/api/http/InstitutionsHttpApiSpec.scala
```diff
@@ -1,196 +1,20 @@
...
-import hmda.persistence.institutions.SubmissionPersistence.UpdateSubmissionStatus
-import hmda.persistence.institutions.{ InstitutionPersistence, SubmissionPersistence }
-import org.iq80.leveldb.util.FileUtils
-import org.scalatest.{ BeforeAndAfterAll, MustMatchers, WordSpec }
 
 import scala.concurrent.duration._
 
-class InstitutionsHttpApiSpec extends WordSpec with MustMatchers with ScalatestRouteTest
-    with InstitutionsHttpApi with BeforeAndAfterAll with RequestHeaderUtils {
+class InstitutionsHttpApiSpec extends InstitutionSpec {
```
schbetsy: It looks like all that's left in this file is authorization-related specs. If that's true, could we rename the spec to be more specific, like `InstitutionsAuthSpec` or similar?

api/src/test/scala/hmda/api/http/InstitutionSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.typesafe.config.ConfigFactory
+import hmda.api.RequestHeaderUtils
+import hmda.persistence.HmdaSupervisor
+import hmda.persistence.HmdaSupervisor.FindActorByName
+import hmda.persistence.demo.DemoData
+import hmda.persistence.institutions.InstitutionPersistence
+import org.iq80.leveldb.util.FileUtils
+import org.scalatest.{ BeforeAndAfterAll, MustMatchers, WordSpec }
+
+trait InstitutionSpec extends WordSpec with MustMatchers with BeforeAndAfterAll
```
schbetsy: A minor detail: I think `InstitutionApiSpec` (or something that includes `http` and/or `api` in it) would be a more descriptive name for this spec. There are other specs that cover other institution functionality, like persistence.

kgudel: Thanks for the suggestion here and with renaming `InstitutionsHttpApiSpec` to something like `InstitutionsAuthSpec`, I was having trouble figuring out the naming and I think that this is right on!

api/src/test/scala/hmda/api/http/InstitutionSpec.scala
jmarin: That makes sense

PR #586, opened 2016-09-23
Removing Thread.sleep

### LINKED ISSUE #579: Explore getting rid of Thread.sleep in tests

nickgrippin:
Several usages of `Thread.sleep()` are still in place.  The ones in `DemoData` are not a priority, due to the nature of them only being used during Demos (and not in production).  The other two usages exist in tests before `system.terminate` calls.  When these calls fire, they also kill all child actors, which results in some unusual test behavior.  I haven't found a solution to these outside of a `Thread.sleep()` call.

Closes #579


schbetsy:
There's one more instance of `Thread.sleep` that exists in the code that you didn't mention, in `UploadPathsSpec`. Did you get a chance to look at that one?

persistence/src/test/scala/hmda/persistence/processing/LocalHmdaEventProcessorSpec.scala
```diff
@@ -113,17 +113,14 @@ class LocalHmdaEventProcessorSpec extends ActorSpec {
...
-      a <- fSubmissions
-      b <- (a ? GetState).mapTo[Seq[Submission]]
-    } yield b.head.submissionStatus
-
-    val submissionStatus = Await.result(fSubmissionStatus, duration)
-    submissionStatus mustBe status
+    fSubmissions.map { subActor =>
+      val fSubmissionSeq = (subActor ? GetState).mapTo[Seq[Submission]]
+      fSubmissionSeq.map { subList =>
+        subList.head.submissionStatus mustBe status
```
schbetsy: Having this assertion happen async appears not to work, unfortunately. When the assertion fails, it doesn't show a failure in the test results.

nickgrippin: This should be resolved in my newest commit

schbetsy:
Small comment: Since the `LocalHmdaEventProcessor` is now part of the `hmda.persistence.processing` package, one of the imports (line 19) causes a warning in the console.

persistence/src/test/scala/hmda/persistence/processing/LocalHmdaEventProcessorSpec.scala
```diff
@@ -114,16 +114,10 @@ class LocalHmdaEventProcessorSpec extends ActorSpec {
...
-    val fSubmissionStatus = for {
-      a <- fSubmissions
-      b <- (a ? GetState).mapTo[Seq[Submission]]
-    } yield b.head.submissionStatus
-
-    val submissionStatus = Await.result(fSubmissionStatus, duration)
-    submissionStatus mustBe status
+    val subActor = Await.result(fSubmissions, 5.seconds)
+    val submissionSeq = Await.result((subActor ? GetState).mapTo[Seq[Submission]], 5.seconds)
+    submissionSeq.head.submissionStatus.code mustBe status.code
```
schbetsy: We'll get much more sensible test failure messages if you change this to
```
submissionSeq.head.submissionStatus mustBe status
```

api/src/test/scala/hmda/api/http/institutions/UploadPathsSpec.scala
```diff
@@ -48,12 +51,14 @@ class UploadPathsSpec extends InstitutionHttpApiSpec with UploadPaths {
...
       } yield {
         s ! UpdateSubmissionStatus(SubmissionId("0", "2017", 1), Signed)
-      }
 
-      Thread sleep 100
-      postWithCfpbHeaders("/institutions/0/filings/2017/submissions/1", file) ~> institutionsRoutes ~> check {
-        status mustBe StatusCodes.BadRequest
-        responseAs[ErrorResponse] mustBe ErrorResponse(400, "Submission already exists", "institutions/0/filings/2017/submissions/1")
+        val submission = Await.result((s ? GetSubmissionById(SubmissionId("0", "2017", 1))).mapTo[Submission], 5.seconds)
+        submission.submissionStatus.code mustBe 12
```
schbetsy: Similar to my above comment: I think it's much more readable to compare the actual status, rather than the status code.

nickgrippin:
@schbetsy, your suggested changes should be in now

schbetsy:
Nice work. LGTM.
 
`Await.result` looks like a good way to get around issues with specs that require future values, but I'm definitely not an expert in Scala best practices or concurrency.  I'm interested in hearing @jmarin 's opinion.

kgudel:
`Await.result` seems like a very cool way of handling this. Well done!

jmarin:
@nickgrippin @schbetsy @kgudel I'm late to this as it's already been merged. No issues with this implementation, good job @nickgrippin. Just wanted to state the obvious, which is, **never ever ever** use `Await.result` in production code unless there is no other option (there always is another option). It blocks the `Future` and throws away all the advantages of using async I/O. It's fine for test code and it is used in that context quite frequently.

PR #589, opened 2016-09-28
Update states doc

awolfe76:
The diff doesn't show it very well but this contains
- small changes in wording in state `9`, it isn't ready to be `Signed` until state `11`
- but notice that this proposes to swap states 5 and 6. this is to keep the states consistent in the order of:
  - `<state>ing`
  - `<state>edWithErrors`
  - `<state>ed`

The `Validating` states are currently in this order, `Parsed` states were not.

If the change to the order of the states is ok we'll need to create an issue to update the back-end. If we want to leave it as is, I'll remove that update.

### DIFF
--- a/Documents/submission-status.md
+++ b/Documents/submission-status.md
@@ -5,18 +5,30 @@ Every submission will store a `SubmissionStatus` object with the following struc
 * `code`: `Integer`
 * `message`: `String`
 
+For example:
+
+```json
+{
+  "id": 3,
+  "status": {
+    "code": 1,
+    "message": "created"
+  }
+}
+```
+
 In order to track the status of a filing for a financial institution, the following states are captured by the backend:
 
-* `1`: `Created` - The filing has been created and is ready to accept data.
+* `1`: `Created` - The submission for a filing period has been created and is ready to accept data.
 * `2`: `Uploading` - Data is currently being uploaded to the system.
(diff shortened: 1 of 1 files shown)


schbetsy:
LGTM. If this is accepted, https://github.com/cfpb/hmda-platform/issues/595 will be necessary.

awolfe76:
@schbetsy - more details added. Have a look when you get a chance.

awolfe76:
Also, to help document our standup conversation:

Its recommended that another state / other states be added to capture different types of errors, grouping syntactical and validity edits in one state and quality and macro edits in another.
_State names and descriptions just to get the point across._

For example:
- `7`: `Validating` - Submitted data is being run through the rules engine that checks for syntactical, validity, quality, and macro edits. All edit checks are performed in this state.
- `8`: `ValidatedWithSyntacticalValidity` - The validation process is complete, but syntactical and/or validity edits are present and a resubmission is required.
- `9`: `ValidatedWithQualityMacro` - The validation process is complete, but quality and/or macro edits are present and require verification.
- `10`: `Validated` - Data submitted passes all syntactical and validity edits and all quality and macro edits, if the existed, have been verified. The data is now considered valid and the IRS report can be generated.

---

I recommend that we continue with what we have and merge this PR and work on #595. If we run into a problem with the states in the future, or if someone feels strongly enough, PRs are always welcome! 😎

schbetsy:
@awolfe76 Thanks for your updates to the documentation. It's all much clearer to me now. 🚀

PR #590, opened 2016-09-28
Improve API error messaging

schbetsy:
This PR has several updates: 

- In the upload path, use the same directive that the other endpoints do (`timedPost`). So now it only responds to `POST`, which matches the specification. Previously, it would respond to any HTTP verb (e.g. `PUT`, `PATCH`), not just `POST`. 
- Update the ErrorResponse object so that the `path` field uses a `Path` object, instead of a simple `String`. This makes error responses more type safe.
- Update the `completeWithInternalError` method to use a `Uri` object as a parameter, instead of the `path` string.
- Extract the uri of each request within its endpoint, rather than reconstructing a path string from path parameters. This allows us to use `Uri` and `Path` objects instead of `String`s in several places.

### DIFF
--- a/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
+++ b/api/src/main/scala/hmda/api/http/HmdaCustomDirectives.scala
@@ -29,8 +29,8 @@ trait HmdaCustomDirectives extends ApiErrorProtocol {
       }
       .result()
 
-  def timedGet: Directive0 = get & time
-  def timedPost: Directive0 = post & time
+  def timedGet = get & time & extractUri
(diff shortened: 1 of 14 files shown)


kgudel:
LGTM! I really like these changes, I think it makes the API much cleaner to use the `extractUri` directive. Also, while you don't mention this change in your notes on the PR, I really the inclusion of a comment above each endpoint with the full path, it makes the api MUCH more legible.

schbetsy:
@jmarin @nickgrippin Any comments? If not, I'll merge tomorrow morning.

PR #591, opened 2016-09-29
Improved console output for test suite

schbetsy:
As we discussed in standup today.

You can read more about ScalaTest config (especially the flags that determine what gets printed to the console) [here](http://www.scalatest.org/user_guide/using_scalatest_with_sbt)


schbetsy:
Known issue: The spec runner still prints out success messages for `Q030`, presumably because it is confused by the nesting of the specs. I think this is acceptable, since the output is still vastly improved

kgudel:
LGTM

PR #587, opened 2016-09-26
Edits api

jmarin:
Adds edits endpoint to submission. Closes #334

PR #596, opened 2016-09-30
Single edit type api

### LINKED ISSUE #335: API - List edits, of a single type, for a submission
`/institutions/<institution>/submissions/period/<period>/<submission>/syntactial|validity|quality|macro`

`GET` - Edits for a given submission, of a single type.

Examples:
[syntactical](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/syntactical.json)
[validity](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/validity.json)
[quality](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/validity.json)
[macro](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/macro.json)

nickgrippin:
Most of these commits will be cleared up once #587 is merged

Closes #335

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -200,7 +200,37 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
     }
     ```
 
+* `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<syntactical|validity|quality|macro>`
+    * `GET`  - List of edits of a specific type, for a given submission
 
+    Example response, with HTTP code 200:
+
+```json
+{
+  "edits": [
+    {
+      "edit": "V555",
+      "lars": [
+        {
+          "lar": {
+            "loanId": "4977566612"
+          }
+        }
+      ]
+    },
+    {
+      "edit": "V550",
+      "lars": [
+        {
+          "lar": {
+            "loanId": "4977566612"
+          }
+        }
+      ]
+    }
+  ]
+}
+```
 
 ## Authorization
 Each endpoint that starts with `/institutions` is protected by three authorization requirements.
--- a/api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
+++ b/api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
@@ -115,28 +115,30 @@ trait SubmissionPaths
       val path = s"institutions/$institutionId/filings/$period/submissions/$seqNr/edits"
       val submissionId = SubmissionId(institutionId, period, seqNr)
       extractExecutionContext { executor =>
-        implicit val ec: ExecutionContext = executor
-        val supervisor = system.actorSelection("/user/supervisor")
-        val fActor = (supervisor ? FindProcessingActor(HmdaFileValidator.name, submissionId)).mapTo[ActorRef]
+        timedGet {
+          implicit val ec: ExecutionContext = executor
+          val supervisor = system.actorSelection("/user/supervisor")
+          val fActor = (supervisor ? FindProcessingActor(HmdaFileValidator.name, submissionId)).mapTo[ActorRef]
 
-        val fEditChecks = for {
-          a <- fActor
-          f <- (a ? GetState).mapTo[HmdaFileValidationState]
-        } yield f
+          val fEditChecks = for {
+            a <- fActor
+            f <- (a ? GetState).mapTo[HmdaFileValidationState]
+          } yield f
 
-        val fSummaryEdits = fEditChecks.map { editChecks =>
-          val s = validationErrorsToEditResults(editChecks.syntactical, Syntactical)
-          val v = validationErrorsToEditResults(editChecks.validity, Validity)
-          val q = validationErrorsToEditResults(editChecks.quality, Quality)
-          val m = validationErrorsToEditResults(editChecks.`macro`, Macro)
-          SummaryEditResults(s, v, q, m)
-        }
+          val fSummaryEdits = fEditChecks.map { editChecks =>
+            val s = validationErrorsToEditResults(editChecks.syntactical, Syntactical)
+            val v = validationErrorsToEditResults(editChecks.validity, Validity)
+            val q = validationErrorsToEditResults(editChecks.quality, Quality)
+            val m = validationErrorsToEditResults(editChecks.`macro`, Macro)
+            SummaryEditResults(s, v, q, m)
+          }
 
-        onComplete(fSummaryEdits) {
-          case Success(edits) =>
-            complete(ToResponseMarshallable(edits))
-          case Failure(error) =>
-            completeWithInternalError(path, error)
+          onComplete(fSummaryEdits) {
+            case Success(edits) =>
+              complete(ToResponseMarshallable(edits))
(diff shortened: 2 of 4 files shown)


api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
```diff
@@ -141,4 +141,43 @@ trait SubmissionPaths
...
+        val fValidationState = for {
+          s <- fHmdaFileValidator
+          xs <- (s ? GetState).mapTo[HmdaFileValidationState]
+        } yield xs
+
+        val fSingleEdits = fValidationState.map { editChecks =>
+          editType match {
+            case "syntactical" =>
+              val s = validationErrorsToEditResults(editChecks.syntactical, Syntactical)
+              SummaryEditResults(s, EditResults.empty, EditResults.empty, EditResults.empty)
```
nickgrippin: Is this the correct format for returning a single type of error?  Or should I simply return an `EditResults` type, rather than wrapping it in `SummaryEditResults`?

api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
```diff
@@ -141,4 +141,39 @@ trait SubmissionPaths
       }
 
     }
+
+  def submissionSingleEditPath(institutionId: String) =
```
schbetsy: This endpoint should be a  `GET`, right? In that case, you'll need to use the `get` directive somewhere in the endpoint (or preferably `timedGet`)

api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
```diff
@@ -141,4 +141,39 @@ trait SubmissionPaths
...
+    path("filings" / Segment / "submissions" / IntNumber / "edits" / Segment) { (period, submissionId, editType) =>
+      val path = s"institutions/$institutionId/filings/$period/submissions/$submissionId/edits/$editType"
+      extractExecutionContext { executor =>
+        implicit val ec: ExecutionContext = executor
```
schbetsy: It looks like these lines (149, 150, and maybe 151) are duplicated in the `submissionEditsPath` endpoint. Is it possible to extract them so the methods can share common code?

api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
nickgrippin: This should be done, along with the `/edits` endpoint

api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
```diff
@@ -115,30 +115,69 @@ trait SubmissionPaths
...
+            case Failure(error) =>
+              completeWithInternalError(path, error)
+          }
         }
       }
 
     }
+
+  def submissionSingleEditPath(institutionId: String) =
+    path("filings" / Segment / "submissions" / IntNumber / "edits" / Segment) { (period, submissionId, editType) =>
```
schbetsy: I think `seqNr` (as it's called in the edits endpoint) is a more accurate name for that variable than `submissionId`.

api/src/main/scala/hmda/api/http/institutions/SubmissionPaths.scala
schbetsy: The point isn't really to reduce line count (though that is often a nice benefit of sharing code), the goal is to abstract out ideas (to reduce code complexity) and reduce duplication (to make it more maintainable).

The next several lines (the entire `for` comprehension) are also shared with the previous method, so they could be included in the extracted code as well.

schbetsy: The point isn't really to reduce line count (though that is often a nice benefit of sharing code), the goal is to abstract out ideas (to reduce code complexity) and reduce duplication (to make it more maintainable).

The next several lines (the entire for comprehension) also seem to be shared with the previous method, so they could be included in the extracted code as well.

schbetsy: I tried it locally to make sure I wasn't lying, and I had luck making a method with this signature:
```scala
  def validationState(subId: SubmissionId)(implicit ec: ExecutionContext): Future[HmdaFileValidationState] = {
```

schbetsy:
I just want to confirm a couple design decisions:
- @awolfe76 + @wpears Is this API response in the format you're expecting?
- Currently, when the request doesn't match one of the edit types, it returns with a `500` error. Is that what we want? (e.g. `institutions/1/filings/2017/submissions/3/edits/otherType`)

Documents/API.md
```diff
@@ -200,7 +200,37 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
     }
     ```
 
+* `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<syntactical|validity|quality|macro>`
+    * `GET`  - List of edits of a specific type, for a given submission
 
+    Example response, with HTTP code 200:
+
+```json
+{
```
awolfe76: The mock-api has a "type" property. Like this:
```json
{
    "type": "validity",
    "edits": []
}
```

awolfe76:
That's not to say that we can't remove the `"type":` because that doesn't match what is currently being returned by the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits` endpoint. Am I right in thinking that the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits`  endpoint is just combination of these endpoints?

nickgrippin:
@awolfe76:  Correct, the `/edits` endpoint returns a wrapped object that contains all four types of edits.

awolfe76:
Do the rest of my comments make sense then? I think that removing the `"type":` makes sense. The app (or anyone/anything else) making the call would know what type was being asked for. Or, maybe the single endpoint could wrap the response in the type, like the `/edits` endpoint is doing?

nickgrippin:
I agree with your point about anyone making the call knowing which type is requested.  I think it might be redundant to include the type in the response... Unless someone wants verification that the syntactical edits they requested are, in fact, syntactical.

awolfe76:
Exactly, that was my thought. There is something to be said for being explicit. But, I think we can revisit that if we need to.

Are we good on the quality and macro responses? If so, 👍  from me.

schbetsy:
@awolfe76 We haven't added the feature of verifying Quality edits yet, and Macros aren't implemented. So those format updates will come with future PRs.

awolfe76:
Ahhh, thanks @schbetsy!

schbetsy:
Sounds like there's consensus that this PR is ready to merge. If I'm wrong, we can revert the merge or add corrections in a separate PR.

PR #603, opened 2016-10-04
Non-existent submission upload fix

### LINKED ISSUE #592: Upload Endpoint Doesn't Return Errors
When I try to upload a file to a submission that does not exist, e.g. `/institutions/1/filings/2017/submissions/77`, the endpoint still returns `200` and the word "uploaded", even though the back end logs an error: `[WARN ][hmda-akka.actor.default-dispatcher-24] Submission does not exist. Could not update submission with id 1-2017-77`.

Instead, the endpoint should return a sensible failure message. Probably a `404` and a message like "Submission 77 does not exist. Could not upload file".

NOTE: This issue is only valid if we keep the current strategy where creating a submission and uploading a file are two separate endpoints. If that API requirement changes, this change will be unnecessary.

nickgrippin:
The `SubmissionPersistence` actor was returning a `Submission` with the `Created` status when it failed to find a submission.  This behavior is changed so that it now returns a `Submission` object with a `Failed` status.

If anyone feels like this overloads the `Failed` status message, I can create a new status (something like `Not Found`).

Closes #592

### DIFF
--- a/api/src/test/scala/hmda/api/http/institutions/UploadPathsSpec.scala
+++ b/api/src/test/scala/hmda/api/http/institutions/UploadPathsSpec.scala
@@ -15,17 +15,18 @@ import akka.pattern.ask
 import scala.concurrent.Await
 
 class UploadPathsSpec extends InstitutionHttpApiSpec with UploadPaths {
+  val csv = "1|0123456789|9|201301171330|2013|99-9999999|900|MIKES SMALL BANK   XXXXXXXXXXX|1234 Main St       XXXXXXXXXXXXXXXXXXXXX|Sacramento         XXXXXX|CA|99999-9999|MIKES SMALL INC    XXXXXXXXXXX|1234 Kearney St    XXXXXXXXXXXXXXXXXXXXX|San Francisco      XXXXXX|CA|99999-1234|Mrs. Krabappel     XXXXXXXXXXX|916-999-9999|999-753-9999|krabappel@gmail.comXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX\n" +
+    "2|0123456789|9|ABCDEFGHIJKLMNOPQRSTUVWXY|20130117|4|3|2|1|10000|1|5|20130119|06920|06|034|0100.01|4|5|7|4|3|2|1|8|7|6|5|4|1|2|9000|0|9|8|7|01.05|2|4\n" +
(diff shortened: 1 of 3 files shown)


kgudel:
LGTM! Its a good new test as well to make sure the api performs as expected. I think that the change to `Failed` is fine, `NotFound` might be a bit more clear but since it is placed in a `getOrElse` and its message is `"No Submission Found"` I think that the logic is easy enough to follow.

api/src/test/scala/hmda/api/http/institutions/UploadPathsSpec.scala
```diff
@@ -15,35 +15,39 @@ import akka.pattern.ask
...
+      postWithCfpbHeaders("/institutions/0/filings/2017/submissions/1", badFile) ~> institutionsRoutes ~> check {
         status mustBe StatusCodes.BadRequest
         responseAs[ErrorResponse] mustBe ErrorResponse(400, "Invalid File Format", "institutions/0/filings/2017/submissions/1")
       }
     }
 
-    "return 400 when trying to upload to a completed submission" in {
-      val badContent = "qdemd"
-      val file = multiPartFile(badContent, "sample.txt")
+    "return a 404 when trying to upload to a non-existant submission" in {
```
schbetsy: `"return a 400 when trying to upload to a non-existent submission"`

api/src/test/scala/hmda/api/http/institutions/UploadPathsSpec.scala
```diff
@@ -15,35 +15,39 @@ import akka.pattern.ask
...
+      postWithCfpbHeaders("/institutions/0/filings/2017/submissions/987654321", file) ~> institutionsRoutes ~> check {
+        status mustBe StatusCodes.BadRequest
+        responseAs[ErrorResponse] mustBe ErrorResponse(400, "Submission 987654321 not available for upload", "institutions/0/filings/2017/submissions/987654321")
+      }
+    }
+
+    "return 404 when trying to upload to a completed submission" in {
```
schbetsy: This description should also say `400` instead of `404`.

schbetsy:
(I can't comment directly on the line, since it isn't part of the diff, but) I think that this update means that `checkSubmissionOverwrite` and any adjacent methods with "overwrite" in the name need to be renamed. That method now also checks that the submission exists, in addition to checking whether it's an overwrite.

PR #613, opened 2016-10-06
Update SubmissionStatusProtocol

### LINKED ISSUE #599: Update SubmissionProtocol to return SubmissionStatus code and message

nickgrippin:
Closes #599

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -110,15 +110,25 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
    "submissions": [
      {
        "id": 3,
-       "submissionStatus": "created"
+       "status": {
+         "code": 1
+         "message: "created"
+       }
      },
      {
        "id": 2,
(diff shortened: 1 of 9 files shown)


kgudel:
I tested this out and looks good for the most part. The only possible issue is that it returns `"submissionStatus"` instead of `"status"` as in the conversation in the issue.

@awolfe76

kgudel:
Also, remember to update the API documentation.

schbetsy:
I agree with @kgudel . According to the specification in the issue, the json field should be `status`, not `submissionStatus`.

schbetsy:
Also, (this may be out of scope for the current PR, but) it's a definite red flag that you didn't have to update any specs to make this change. We should have tests that ensure that our API is returning responses in the correct format.

awolfe76:
Looks 👍  to me.

kgudel:
Looks good Nick!!!

Issue #600, opened 2016-10-03
Update /institutions/<institution>endpoint


awolfe76:
From standup discussion we talked about making the "status" endpoints consistent.

The `/institutions/<institution>` endpoint currently returns a filing array:

```json
"filings": [
    {
      "period": "2017",
      "institutionId": "12345",
      "status": "not-started"
    },
    {
      "period": "2016",
      "institutionId": "12345",
      "status": "completed"
    }
]
```

It should probably return something like

```json
"filings": [
    {
        "period": "2017",
        "institutionId": "12345",
        "status": {
            "code": 1,
            "message": "not-started"
        }
    },
    {
        "period": "2016",
        "institutionId": "12345",
        "status": {
            "code": 2,
            "message": "in-progess"
        }
    }
]
```

to be more like the `/institutions/<institution>/filings/<period>/submissions/latest` endpoint. With the codes and messages being:
- `1` - `not-started`
- `2` - `in-progress`
- `3` - `cancelled`
- `4` - `completed`

__Note__: I'm not sure of the order of codes or the exact messages.

awolfe76:
Quick question, does the idea of `cancelled` seem better for the filing or for the submission? Seems like the submission is what gets cancelled, not the filing. I don't think they would cancel a filing ever, that would mean that they weren't required to file. Right?

If that is right we'll have to create an issue to add the `cancelled` state to the `/institutions/<institution>/filings/<period>/submissions` endpoint.

jmarin:
Yeah, I agree that `cancelled` probably doesn't belong in `filings`. I'm not even sure what action triggers a cancellation in submissions either. Is this something that the user would do?

awolfe76:
We have talked about the idea of a "Cancel Submission" button, but haven't implemented anything for it. Its a nice-to-have feature so at this point I think its ok to hold off it that for a bit and wait until later to implement. Just my opinion.

Another thing that could be called a cancellation maybe when the submit that first time and it has errors (or even without errors) and they choose to re-submit again. That first submission could be considered `cancelled`. Maybe we can call is something else or provide a message saying `Cancelled by resubmission`, but I think it works.

awolfe76:
Due to #604 removing the `/institutions/<institution>/summary` this issue should apply to `/institutions/<institution>`. I've updated the title and comment.

debseidner:
[...]

Issue #601, opened 2016-10-03
Update /institutions endpoint


awolfe76:
Like #600, `/institutions` also gives institutions a status.

```json
{
    "id": "12345",
    "name": "First Bank",
    "status": "active"
}
```

We should probably return something like

```json
{
    "id": "12345",
    "name": "First Bank",
    "status": {
        "code": 1,
        "message": "active"
    }
}
```

I'm guessing this is just an `active` or `inactive` so a `1` or a `2`.

PR #616, opened 2016-10-06
Add documentation for filing and institution status

### LINKED ISSUE #602: Documentation for states
Sorry if these already exist but I couldn't find them.

Like with the [submission states](https://github.com/cfpb/hmda-platform/blob/master/Documents/submission-status.md), it would be good to document the states for both the institutions (see #601) and for the filings (see #600).

kgudel:
Closes #602 

The filing status information is based on [this comment](https://github.com/cfpb/hmda-platform/pull/411#issuecomment-226338262) and [this comment](https://github.com/cfpb/hmda-platform/issues/600#issuecomment-251697290)

The institution status information is based on a preliminary conversation with @lgdean

### DIFF
--- a/Documents/filing-status.md
+++ b/Documents/filing-status.md
@@ -19,7 +19,7 @@ For example:
 
 In order to track the status of a filing for a financial institution, the following states are captured by the backend:
 
-* `1`: `NotStarted` - A filing has been opened but no submission has been submitted for the filing.
-* `2`: `InProgress` - At least one submission has been submitted for the filing but no submission has the status of signed.
-* `3`: `Completed` - A signed submission has been submitted for this filing, signifying that the filing is complete.
-* `4`: `Cancelled` - The filing has been cancelled, signifying that the filing is no longer required.
+* `1`: `not-started` - A filing has been opened but no submission has been submitted for the filing.
+* `2`: `in-progress` - At least one submission has been submitted for the filing but no submission has the status of signed.
+* `3`: `completed` - A signed submission has been submitted for this filing, signifying that the filing is complete.
+* `4`: `cancelled` - The filing has been cancelled, signifying that the filing is no longer required.
--- a/Documents/filing-status.md
+++ b/Documents/filing-status.md
@@ -1,0 +1,25 @@
+## Filing status
+
+Every filing will store a `FilingStatus` object with the following structure:
+
+* `code`: `Integer`
+* `message`: `String`
+
+For example:
+
+```json
+{
+  "id": 3,
+  "status": {
+    "code": 1,
+    "message": "not-started"
+  }
+}
+```
+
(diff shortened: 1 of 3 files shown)


Documents/institution-status.md
```diff
@@ -0,0 +1,20 @@
...
+
+* `status`: `String`
+
+For example:
+
+```json
+      "institution": {
+      "id": "12345",
+      "name": "First Bank",
+      "status": "active"
```
awolfe76: I think we would want this to be in the same format:
```json
"status": {
    "code": 1,
    "message": "active"
}
```

kgudel: I was taking the json directly from the API format, which I think that this documentation should follow. If we want to change how the bank status is returned I think that would be the purview of another ticket and the docs could be changed in that PR to match the new API response.

kgudel:
@awolfe76 By adding things about Panel, do you mean adding something like a description of what panel is?

awolfe76:
#601 is doing just that. Its fine if we leave it to that PR to make the change.

For panel, am just wondering if we should add a property to this to show if an institution is in panel or not? Again, another issue is probably best.

👍  from me.

awolfe76:
And sorry for the confusion. Just realized that `inactive` and `active` was in fact talking about panel.

Documents/institution-status.md
```diff
@@ -0,0 +1,20 @@
+## Institution status
+
+Every filing will store a `InstitutionStatus` object with the following structure:
+
+* `status`: `String`
+
+For example:
+
+```json
+      "institution": {
```
schbetsy: Looks like indentation is a little off here

schbetsy:
Just a quick formatting comment, then this looks good.

Good write-up, @kgudel! Very clear.

Documents/filing-status.md
```diff
@@ -0,0 +1,25 @@
...
+  "status": {
+    "code": 1,
+    "message": "not-started"
+  }
+}
+```
+
+In order to track the status of a filing for a financial institution, the following states are captured by the backend:
+
+* `1`: `NotStarted` - A filing has been opened but no submission has been submitted for the filing.
```
awolfe76: Should this be `not-started` to match the message above? Probably the same for the other statuses too.

kgudel: Can do!

Documents/institution-status.md
```diff
@@ -0,0 +1,20 @@
+## Institution status
+
+Every filing will store a `InstitutionStatus` object with the following structure:
```
schbetsy: Should this say "Every institution", rather than "Every filing"?

kgudel: Indeed it should!

awolfe76:
👍

PR #617, opened 2016-10-07
Separate LAR and TS validation errors

### LINKED ISSUE #605: Persist TS and LAR validation errors separately
Currently `HmdaFileValidator` stores validation errors for both `TransmittalSheet` and `LoanApplicationRegister` without differentiating them. Similar to how the parser errors are managed, create different events for `Lar` and `Ts` (`TsSyntacticalError`, `LarSyntacticalError`, etc)

nickgrippin:
Still considering this approach and looking for other side effects of this change.

So without splitting up `ValidationError` itself, which seems like a massive amount of work, I was considering just creating two wrapper classes (one for TS and one for LAR) before we call `persist` in `HmdaFileValidator`.  Right now I'm working through that possible approach.

UPDATE: Ready for review, also includes fixes to the `HmdaFileValidatorSpec` tests.

Closes #605


awolfe76:
@nickgrippin - would this change documentation?

nickgrippin:
Yes, it would... let me update the docs really quick.  

Forgot about it, thanks for the reminder!  I'll have to update the docs for my other PR as well.

nickgrippin:
Actually, on second thought, this shouldn't change the API.  My other PR definitely will, so I'll update the docs over there.

PR #625, opened 2016-10-14
Test Config Fix

nickgrippin:
Huge shoutout to @schbetsy for help!


schbetsy:
Shouldn't this also include removing the unneeded config that was added in 2b112532b971b8a8d9?

nickgrippin:
@schbetsy, this should be done in my open PR in f39be5ad5e46d4689d32aaa119092111dbb86aba

PR #631, opened 2016-10-15
Macro edits

### LINKED ISSUE #219: Validation (2017): Q007 (Action Taken - Type)
**Edit #:** Q007
**Type:** Macro
**Column:** Action Taken - Type

If action taken type = 2, then the total number of these loans should be ≤ 15% of the total number of loan applications.

Additional resources:

### LINKED ISSUE #220: Validation (2017): Q008 (Action Taken - Type)
**Edit #:** Q008
**Type:** Macro
**Column:** Action Taken - Type

If action taken type = 4, then the total number of these loans should be ≤ 30% of the total number of loan applications.

Additional resources:

### LINKED ISSUE #630: Add macro validation engine

jmarin:
Closes #219 
Closes #220 
Closes #630


validation/src/main/scala/hmda/validation/rules/AggregateEditCheck.scala
```diff
@@ -0,0 +1,28 @@
...
+abstract class AggregateEditCheck[A <: Source[T, NotUsed], T] {
+
+  def count(input: Source[T, NotUsed])(implicit materializer: ActorMaterializer): Future[Double] = {
+    input.runWith(sinkCount)
+  }
+
+  private def sinkCount(): Sink[T, Future[Double]] = {
+    Sink.fold[Double, T](0) { (acc, _) =>
+      val total = acc + 1
+      total.toDouble
```
kgudel: In playing with this pr, I don't think that this line is necessary.

validation/src/test/scala/hmda/validation/rules/lar/macro/Q008Spec.scala
```diff
@@ -0,0 +1,24 @@
...
+class Q008Spec extends MacroSpec {
+
+  "Q008" must {
+    "be valid if withdrawn <= 0.30 * total" in {
+      larSource.mustPass
+    }
+
+    "be invalid if withdrawn > 0.30 * total" in {
+      val badLar1 = lars.head.copy(actionTakenType = 4)
+      val badLar2 = lars.head.copy(actionTakenType = 4)
```
kgudel: Since all lines are the same I think it makes a bit more sense to remove `badLar2` and add `badLar1` on multiple times.

validation/src/main/scala/hmda/validation/rules/AggregateEditCheck.scala
kgudel: This line actually does seem to be necessary as without it I am getting Q008 failing given any number of lars with `actionTakenType = 4` above 0.

kgudel: It looks like I was wrong the second time and that this can be removed, sorry for all the confusion.

kgudel:
Other than some test changes this LGTM! I like the akka stream implementation of macro edits a lot.

validation/src/main/scala/hmda/validation/rules/lar/macro/Q007.scala
```diff
@@ -0,0 +1,37 @@
...
+  val multiplier = config.getInt("hmda.validation.macro.Q007.numOfLarsMultiplier")
+
+  override def name = "Q007"
+
+  override def apply(lars: LoanApplicationRegisterSource)(implicit system: ActorSystem, materializer: ActorMaterializer, ec: ExecutionContext): Future[Result] = {
+
+    val approvedButNotAccepted =
+      count(lars.filter(lar => lar.actionTakenType == 2))
+
+    val total = count(lars)
```
schbetsy: Do most Macro edits require calculating against the total number of lars in the file? If so, could it make sense to calculate the total number just once and send the value to each Macro edit as a parameter?

nickgrippin: Even better, if there was a way to use `totalLines` from the TS, we wouldn't even need to calculate the total number at all.

jmarin: @schbetsy I did think about this, it complicated the implementation on the first pass I did at this and decided to keep it simpler for now. Will try to put something together, see what it looks like. 

@nickgrippin I don't think we can really do that. There is at least one rule that checks that the `totalLines` field in TS is correct (there is no guarantee that this number is correct).

schbetsy: @jmarin Sounds good. If that's too big a change to put in this PR, I'm fine with making that a future improvement.

validation/src/test/scala/hmda/validation/rules/lar/macro/Q007Spec.scala
```diff
@@ -0,0 +1,33 @@
...
+import hmda.validation.rules.AggregateEditCheck
+import hmda.validation.rules.lar.`macro`.MacroEditTypes.LoanApplicationRegisterSource
+import org.scalacheck.Gen
+
+class Q007Spec extends MacroSpec {
+
+  val config = ConfigFactory.load()
+  val multiplier = config.getDouble("hmda.validation.macro.Q007.numOfLarsMultiplier") * 100
+
+  val multiplierGen: Gen[Int] = Gen.choose(1, multiplier.toInt)
```
schbetsy: I think property specs here are a good idea 👍 . Two quick things:

- I would change the name of this generator and the variable it generates (I think `multiplier` isn't quite accurate, since it's already been multiplied by 100). Maybe `proportion`? 
- Could we also get the inverse test? I'd like to see one that shows that when badlars is > multiplier, the edit fails

jmarin:
This is ready for another round of reviews. I've updated the tests a bit, fixed an issue with macro edits configuration and refactored them a bit

validation/src/main/scala/hmda/validation/engine/lar/macro/LarMacroEngine.scala
```diff
@@ -0,0 +1,42 @@
...
+      ).map(checkAggregate(_, larSource, "", Macro))
+    )
+      .map(checks => validateAll(checks, larSource))
+
+  }
+
+  private def checkAggregate(
+    editCheck: AggregateEditCheck[LoanApplicationRegisterSource, LoanApplicationRegister],
+    input: LoanApplicationRegisterSource,
+    inputId: String,
```
nickgrippin: This doesn't seem to be used anywhere, is it something that will be used in the future?

jmarin: At this point in time, I don't think so. That `id` is there to identify individual loan and respondent ids at the record level. For groups of `LARs` we haven't come up with an id to pass yet. It's like this to reuse the existing functionality of `convertResult`. I'm ok with refactoring this, but I also don't think this should hold this PR

nickgrippin:
LGTM

validation/src/test/scala/hmda/validation/rules/lar/macro/Q007Spec.scala
```diff
@@ -5,28 +5,38 @@ import com.typesafe.config.ConfigFactory
...
+    newLarSource.mustPass
+  }
+
+  property(s"be invalid if not accepted > $multiplier * total") {
+    val numOfGoodLars = (sampleSize * multiplier).toInt + 1
+    val newLarSource = newSource(numOfGoodLars)
+    newLarSource.mustFail
+  }
+
+  private def newSource(numOfGoodLars: Int) = {
```
kgudel: I think it would be cleaner to move this function into a utils since it will likely be used in all of our specs.

kgudel:
Other than a quick test fix LGTM!

validation/src/test/scala/hmda/validation/rules/lar/macro/Q007Spec.scala
nickgrippin: This specific implementation is only relevant to a certain edit based on `actionTakenType`.  There might be a way to pull out common code, but I wouldn't block the PR based on that.

PR #633, opened 2016-10-17
Macro edits persistence

jmarin:
Persist macro edits. Depends on #631


validation/src/main/scala/hmda/validation/engine/lar/LarEngine.scala
```diff
@@ -2,14 +2,15 @@ package hmda.validation.engine.lar
...
+import hmda.validation.engine.lar.`macro`.LarMacroEngine
 import hmda.validation.engine.lar.syntactical.LarSyntacticalEngine
 import hmda.validation.engine.lar.validity.LarValidityEngine
 import hmda.validation.engine.lar.quality.LarQualityEngine
 
 import scalaz._
 import Scalaz._
 
-trait LarEngine extends LarSyntacticalEngine with LarValidityEngine with LarQualityEngine {
+trait LarEngine extends LarSyntacticalEngine with LarValidityEngine with LarQualityEngine with LarMacroEngine {
```
jmarin: I don't think so, no. This was included here to encapsulate all the validation functions in one place. The place where they all show up currently is in `HmdaFileValidator`, which extends `LarEngine` with all the edit type checks (syntactical, validity, quality and macro). The macro checks are used in this file, after the stream has been passed through the other types of validation (see `ValidateMacro` message). Hope that makes sense

PR #638, opened 2016-10-19
Cross compile model

### LINKED ISSUE #636: Cross compile model project
Configure `model` to cross compile to `Scala.js`

jmarin:
Closes #636


kgudel:
Everything here LGTM as far as I can tell. In order to test that the javascript I simply ran `fastOptJS` and it didn't throw a compile error. I was wondering if there were other good ways to test something like this. And also if Travis is currently or if we can make it attempt to create the javascript as part of its tests going forward.

jmarin:
Testing for `JavaScript` is already happening. The whole point of a cross compiled project is to be able to minimize duplication and both compile and run tests for all target platforms. The code in `shared` gets compiled and tested for both the `JVM` and `JS` runtimes. Code in the `js` or `jvm` projects only gets executed for the corresponding platform.

In order to run `JavaScript` tests `node` is required. You can go to the `modelJS` project and run `test`, you will see that a `JavaScript` file gets generated before running the tests. This is also happening in the CI Server (I already installed `nodejs` in Travis). If you look at the [build](https://travis-ci.org/cfpb/hmda-platform/builds/168964957) most tests related to the `model` project are run twice, once for each platform (i.e. `InstitutionSpec`, line 582 for the `JVM`, line 616 for `JS`)

kgudel:
Thank you for pointing that out Juan, I was not aware of that feature which is a really good one. In that case this definitively LGTM!

schbetsy:
Testing my understanding: This PR takes everything that was in the `/model` project and moves it to `/model/shared` except the classes that use the CBSA files. The CBSA-related functionality can't be shared because ScalaJS doesn't have the same ability to store/read information in files. CBSA stuff either won't exist in ScalaJS, or it will have to be re-created using different, JS-compatible libraries.

Is that right?

schbetsy:
Also, I noticed that the JS-based tests don't honor the test formatting preferences that the JVM tests do. i.e. a single run of `sbt test` has both this:
```
[info] AgencySpec:
[info] Agency
```
and this:
```
[info] AgencySpec:
[info] Agency
[info] - must resolve Agency case class to the appropriate agencyId
[info] - must resolve agencyId to the appropriate Agency case class
[info] - must must have entries for all agencies
```

It would be nice to make those consistent, but that doesn't have to be this PR.

jmarin:
@schbetsy That is exactly right. File access is platform specific, so we need an implementation for each platform. Given how big the census files are, it's unlikely that these will load efficiently in the browser. Something to test in the future, perhaps.

nickgrippin:
LGTM.

Both modelJS and modelJVM compile and all tests pass.

PR #639, opened 2016-10-19
InstitutionStatus in JSON responses

### LINKED ISSUE #601: Update /institutions endpoint

schbetsy:
Closes #601 

1. Changes `InstitutionStatus` model to have a `code` and `message` field. `InstitutionStatus` is now modeled exactly like `SubmissionStatus`. This requires some updates, especially of test code and imports.
2. Adds institution status to JSON representation in all endpoints that return institution data (I believe just `/institutions` and `/institutions/<id>`).

### DIFF
--- a/model/src/main/scala/hmda/model/institution/InstitutionStatusMessage.scala
+++ b/model/src/main/scala/hmda/model/institution/InstitutionStatusMessage.scala
@@ -1,0 +1,6 @@
+package hmda.model.institution
+
+object InstitutionStatusMessage {
+  val activeMsg = "active"
+  val inactiveMsg = "inactive"
+}
--- a/persistence/src/main/scala/hmda/persistence/demo/DemoInstitutions.scala
+++ b/persistence/src/main/scala/hmda/persistence/demo/DemoInstitutions.scala
@@ -46,7 +46,10 @@ object DemoInstitutions extends ResourceUtils {
   }
 
   def toStatus(text: String): InstitutionStatus = {
-    InstitutionStatus.values.filter(status => status.entryName == text).head
+    text match {
+      case "active" => Active
+      case "inactive" => Inactive
+    }
   }
 
 }
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -32,12 +32,18 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
         {
           "id": "12345",
           "name": "First Bank",
-          "status": "active"
+          "status": {
+            "code": 1,
+            "message": "active"
+          }
         },
         {
           "id": "123456",
           "name": "Second Bank",
-          "status": "inactive"
+          "status": {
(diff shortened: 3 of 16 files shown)


awolfe76:
@schbetsy - should `API.md` also be updated with this? The `/institutions` and `/institutions/<id>` endpoints, like you mention.

model/src/main/scala/hmda/model/institution/InstitutionStatusMessage.scala
```diff
@@ -0,0 +1,6 @@
+package hmda.model.institution
+
+object InstitutionStatusMessage {
```
schbetsy: Whenever code depends on a specific string, it's always better to represent it as an object in the system, not as a string. That way the compiler can help us and prevent any mysterious bugs that may come from accidental misspelling or mis-capitalization. In this PR, the main benefit is [here in `InstitutionProtocol.scala`](https://github.com/cfpb/hmda-platform/pull/639/files#diff-3019bb0f3feec594ca0ddf5ebf517947R21), so that when we deserialize JSON, we are certain to match on the correct strings.

Also, this construction parallels the way `SubmissionStatus` uses `SubmissionStatusMessage`.

schbetsy:
Thanks @awolfe76 . I updated the API.md documentation in 23d6749907a4b9ad98c87d4c3e88ad005c16a89a.

awolfe76:
One other small thing. [This line](https://github.com/cfpb/hmda-platform/pull/639/files#diff-91970c103fe5d6ce41f695419d8db764R5) (line 5 of `institution-status.md`) should now represent the object instead of `String`.

persistence/src/main/scala/hmda/persistence/demo/DemoInstitutions.scala
```diff
@@ -46,7 +46,10 @@ object DemoInstitutions extends ResourceUtils {
   }
 
   def toStatus(text: String): InstitutionStatus = {
-    InstitutionStatus.values.filter(status => status.entryName == text).head
+    text match {
+      case "active" => Active
```
schbetsy: I know we're matching on specific strings here... My reasoning is that this file's purpose is to parse `persistence/src/main/resources/demoInstitutions.csv`, which uses its own system for representing institutions, not necessarily the strings that match `InstitutionStatusMessage`. 
But I could be convinced to change these to `activeMsg` and `inactiveMsg` if someone feels strongly.

PR #641, opened 2016-10-19
Cross compile parser

### LINKED ISSUE #637: Cross compile parser project
Configure `parser` to cross compile to `Scala.js`

jmarin:
Closes #637


parser/js/src/test/scala/hmda/parser/fi/DateGenerators.scala
```diff
@@ -0,0 +1,15 @@
...
+import scalajs.js
+
+object DateGenerators {
+  def randomDate: Gen[Int] = {
+    val beginDate = new js.Date("2017-01-01")
+    val endDate = new js.Date("2020-12-31")
+
+    for {
+      randomDate <- Gen.choose(beginDate.getTime(), endDate.getTime())
+    } yield new js.Date(randomDate).getTime().toInt
```
nickgrippin: The bug you're encountering with the javascript test is caused here.  The `getTime` method returns a number of milliseconds, not a formatted date.  In turn, this causes the created DAT to be offset on every field after the Date, since this returns a longer number than was expected (with a length of 10 instead of 8).  Which in turn leads to the weird `NumberFormatException`s.

jmarin: Are you sure that is where the problem is?. The `JVM` version produces the same output on the `randomDate` generator, an `Int` (milliseconds) but the tests pass in that case. See [here](https://github.com/jmarin/hmda-platform/blob/scalajs-parser/parser/jvm/src/test/scala/hmda/parser/fi/DateGenerators.scala#L9)

nickgrippin: The `JVM` version produces an `Int` of a specific date format.  There, a `Date` object is turned into a formatted `String` (`"yyyyMMdd"`), then converted that to an `Int` (e.g. `20161102`).

The `JS` version is producing a random Date in milliseconds, which have a different length than the `JVM` output.

jmarin: Commit fixing this coming up. Thanks for your help getting to the bottom of this

nickgrippin:
LGTM

PR #643, opened 2016-10-19
Update parser to recognize empty fields at the end of lines

### LINKED ISSUE #642: Parser Does Not Properly Parse Strings that End in Multiple Pipes
This issue was found in GHE/hmda-otis-an-erika

This is an issue in how `.split('|')` functions as:

`"1|1||` returns `Array("1","1")`, instead of the expected `Array("1","1","","")`.

As the last field of both the ts and lar is a required this is an edge case, however, someone may run into while throwing test data at the parser to check if it is well formed.

kgudel:
Closes #642


jmarin:
Haven't looked at the code yet, but from the description this PR seems to be undoing the work committed [here](https://github.com/cfpb/hmda-platform/commit/b5f9605f1625607bf4a0485a0c82adadee000343)

kgudel:
Thanks @jmarin , I'll talk to @schbetsy about this one and think about how we want to handle situations like this.

schbetsy:
Hey all. Just bumping this since it's been a week. I agree with the need for this PR, and Kenneth's implementation looks good to me. Anyone else want to comment before I merge?

PR #645, opened 2016-10-20
Mock Endpoint Fix

nickgrippin:
To avoid having to deal with relative paths in different runtime environments, I'm just dumping the static JSON into a String.  Not as clean, but this is only temporary.


nickgrippin:
Test failure is in `LocalHmdaEventProcessorSpec`, unrelated to the changes in this PR

jmarin:
Did some timeout change in the tests we're running lately? I'm also seeing issues in my `scalajs-model` branch with timeouts, and other branches were having this sort of failure more often than usual after yesterday's commits to master. It might be worthwhile checking that we didn't change anything by accident there

jmarin:
@nickgrippin +1 on putting the JSON in a string. 

As for the event processor, take a look at my last PR (#646). It achieves the same (increased timeout for filter log), but also consolidates the test configuration with other actors. In that case it's not necessary to add `testKit.filter-leeway` (and if you want to add it, it would belong in `src/test/resources` and not in `src/main/resources`, as this setting is related to tests exclusively)

schbetsy:
I think this is fine, since these mock responses are temporary and will be removed anyway. But FYI, the code has an example of how to read content from a file correctly, so it won't break in different environments. Let's use [this pattern](https://github.com/cfpb/hmda-platform/blob/master/model/src/main/scala/hmda/model/ResourceUtils.scala#L8) next time.

nickgrippin:
@schbetsy: Thanks for pointing that out!  I'll keep that in mind for next time.

@jmarin:  I'll take out that config then, and give your PR a look-through.

PR #646, opened 2016-10-21
Consolidate test config (event processor)

jmarin:
Consolidates configuration with other actors in the `persistence` project. Might help with sporadic test failures in CI server (event stream log capture timeout is increased)


nickgrippin:
LGTM!

PR #653, opened 2016-10-31
Admin api

### LINKED ISSUE #609: Provide admin API for institution persistence to internal clients
Clients to the application (i.e. the panel creation process) need to be able to insert, update, etc. institution data. Separate implementation from the existing API

jmarin:
Closes #609


api/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -0,0 +1,82 @@
...
+            val supervisor = system.actorSelection("/user/supervisor")
+            val fInstitutionsActor = (supervisor ? FindActorByName(InstitutionPersistence.name)).mapTo[ActorRef]
+            val fCreated = for {
+              a <- fInstitutionsActor
+              i <- (a ? CreateInstitution(institution)).mapTo[Institution]
+            } yield i
+
+            onComplete(fCreated) {
+              case Success(i) =>
+                complete(ToResponseMarshallable(StatusCodes.Created -> i))
```
kgudel: I think that it would be good to return a response that informs that user if the institution already exists, if it does. I think that this should be relatively easy to implement using the same pattern as the put.

api/src/main/scala/hmda/api/HmdaPlatform.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private def startActors(system: ActorSystem, supervisor: ActorRef)(implicit ec: ExecutionContext): Unit = {
+    lazy val actorTimeout = config.getInt("hmda.actor.timeout")
+    implicit val timeout = Timeout(actorTimeout.seconds)
+
+    (supervisor ? FindActorByName(SingleLarValidation.name))
+      .mapTo[ActorRef]
+      .map { actor =>
+        //log.info(s"Started validator at ${actor.path}")
```
schbetsy: I'm curious about these `log` commands. Should they be uncommented?

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -56,6 +56,7 @@ class InstitutionPersistence extends HmdaPersistentActor {
         persist(InstitutionCreated(i)) { e =>
           log.debug(s"Persisted: $i")
           updateState(e)
+          sender() ! e.i
```
kgudel: Now that this is sending a response when an institution is added `DemoData` should probably be updated to accept or at least acknowledge this response. I'll make another ticket for it though as the only effect I can see is that the log displays 4 dead letters when demo is set to true.

api/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -0,0 +1,82 @@
...
+  implicit val system: ActorSystem
+  implicit val materializer: ActorMaterializer
+  implicit val timeout: Timeout
+
+  val log: LoggingAdapter
+
+  val institutionsWritePath =
+    path("institutions") {
+      timedPost { uri =>
+        entity(as[Institution]) { institution =>
```
schbetsy: It looks like these lines are duplicated and could be moved up outside of the `timedPost`/`timedPut`:
```scala
entity(as[Institution]) { institution =>
  extractExecutionContext { executor =>
    implicit val ec: ExecutionContext = executor
    val supervisor = system.actorSelection("/user/supervisor")
    val fInstitutionsActor = (supervisor ? FindActorByName(InstitutionPersistence.name)).mapTo[ActorRef]
```
(Though then the system wouldn't check for the appropriate verbs before it starts processing the request; not sure whether that's a concern)

api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
```diff
@@ -0,0 +1,140 @@
...
+      case JsString(FRS.name) => FRS
+      case JsString(HUD.name) => HUD
+      case JsString(NCUA.name) => NCUA
+      case JsString(OCC.name) => OCC
+      case _ => throw DeserializationException("Unable to deserialize")
+    }
+
+  }
+
+  implicit object InstitutionTypeJsonFormat extends RootJsonFormat[InstitutionType] {
```
jmarin: Unfortunately this implementation fails the JSON serialization test in `WriteInstitutionProtocolSpec`

hkeeler: Huh.  I made these changes locally, and I _thought_ all the tests passed.  Anyway, not worth holding up the PR over.  I'll create another issue to remind me to look back into it.

lgdean: That seems to be because of the quotes around the string.  This version works (though you might prefer to clean it up):
```
    override def read(json: JsValue): InstitutionType = {
      json match {
        case JsString(name) =>
          try {
            InstitutionType.withName(name)
          } catch {
            case iae: NoSuchElementException => throw DeserializationException(
              s"Unable to translate JSON string into valid InstitutionType value: $name"
            )
          }
        case _ => throw DeserializationException("Unable to deserialize")
      }
    }
```

Issue #620, opened 2016-10-11
Encode response on all API endpoints


jmarin:
Encode the response with the encoding that is requested by the client via the `Accept-Encoding` header. 

See documentation [here](http://doc.akka.io/docs/akka/2.4.7/scala/http/routing-dsl/directives/coding-directives/encodeResponse.html#encoderesponse)

schbetsy:
- [ ] 
- [ ] 
- [ ] 
- [ ] 
- [ ]

PR #658, opened 2016-11-01
Encode all API responses

schbetsy:
Addresses #620


nickgrippin:
@schbetsy: does this fix the test issue we found?

schbetsy:
@nickgrippin Yep. As you pointed out, my previous testing strategy didn't work because the system wasn't recognizing the `Accept-Encoding` header correctly. I found a different way to add the header that the `encodeResponse` directive does handle correctly.

nickgrippin:
LGTM!

PR #659, opened 2016-11-02
Query institutions

### LINKED ISSUE #660: Separate write from read for Institutions (CQRS)

jmarin:
Closes #660

Some notes:

* DAO implementation is specific to `Institution`. A generic implementation will be included at a later date when the `query` project is augmented. This will also be properly abstracted with a `Repository` data access model
* `InstitutionView` reads the journal as institution events come in. It stores the state and keeps track of what it has seen (to recover from crashes). It also sends events to `InstitutionDBProjection` and is now responsible for the institutions API endpoints, while `InstitutionPersistence` is responsible for writing (separation of writes from reads). 
* `InstitutionDBProjection` is a projection of the institution events in the journal, it stores them in an H2 database. This will be later changed to use Postgres


query/src/main/scala/hmda/query/projections/institutions/InstitutionView.scala
```diff
@@ -0,0 +1,101 @@
...
+        case InstitutionModified(i) =>
+          val others = institutions.filterNot(_.id == i.id)
+          InstitutionViewState(others + i, seqNr)
+      }
+    }
+  }
+
+}
+
+class InstitutionView extends HmdaPersistentActor {
```
kgudel: To check my knowledge, I believe that in the future this actor will a different backend from `InstitutionPersistence` for storing data. This one will be optimized for reads, whereas the one for `InstitutionPersistence` will be optimized for writes. Is that correct?

query/src/main/scala/hmda/query/projections/institutions/InstitutionView.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def createInstitutionQuery(system: ActorSystem): ActorRef = {
+    system.actorOf(InstitutionView.props(), "institutions-view")
+  }
+
+  case class InstitutionViewState(institutions: Set[Institution] = Set.empty[Institution], seqNr: Long = 0L) {
+    def updated(event: Event): InstitutionViewState = {
+      event match {
+        case InstitutionCreated(i) =>
+          InstitutionViewState(institutions + i, seqNr)
```
kgudel: It looks like `seqNr` always remains at `0`.

query/src/test/scala/hmda/query/projections/institutions/InstitutionViewSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    institutionQuery ! EventWithSeqNr(1, InstitutionCreated(i1))
+    institutionQuery ! EventWithSeqNr(2, InstitutionCreated(i2))
+    institutionQuery ! EventWithSeqNr(3, InstitutionCreated(i3))
+    institutionQuery ! EventWithSeqNr(4, InstitutionModified(i4))
+  }
+
+  "Institutions Projection" must {
```
kgudel: I think that it would be useful to have a test that `InstitutionsView` will successfully read events that it doesn't have from the `InstitutionPersistence` journal, though I don't know if this is the right place for it as that seems like more of an integration test.

query/src/main/scala/hmda/query/projections/institutions/InstitutionView.scala
jmarin: Not exactly, but close. The view is responsible for reading the journal and it is also a persistent actor, it keeps track of what it has seen (through the sequence number) and starts reading the journal again from there. As events come in, it reads the latest event and sends it to other actors to process (the projector, `InstitutionDBProjection`). These other actors will use a different backend, optimized for read and / or analysis. As it stands right now the view is serving the read only API, the data is stored in memory. This will probably work for some data feeds but not others.

query/src/main/scala/hmda/query/HmdaQuerySupervisor.scala
```diff
@@ -0,0 +1,26 @@
...
+  def createQuerySupervisor(system: ActorSystem): ActorRef = {
+    system.actorOf(HmdaQuerySupervisor.props(), "query-supervisor")
+  }
+}
+
+class HmdaQuerySupervisor extends HmdaSupervisorActor {
+
+  override protected def createActor(name: String): ActorRef = name match {
+    case id @ InstitutionView.name =>
+      println(name)
```
schbetsy: This `println` is unnecessary.

query/src/main/scala/hmda/query/projections/institutions/InstitutionView.scala
schbetsy: Now I'm confused. So the `InstitutionDBProjection` isn't used to query institution data in the api (that's what `institutionView` is for). Is the DB Projection for future publication/analysis of HMDA data?

query/src/test/scala/hmda/query/projections/institutions/InstitutionViewSpec.scala
schbetsy: Not sure what you mean by "read events that it doesn't have". Could you clarify?

query/src/main/scala/hmda/query/projections/institutions/InstitutionView.scala
jmarin: That's how it is in this PR, yes. So `InstitutionPersistence` is exclusively concerned with writing stuff. `InstitutionView` reads created / modified events as they get to the journal and updates state accordingly and exposes the current state through the API. This part was previously done also in `InstitutionPersistence`, now it's separated. At the same time, it sends these events to `InstitutionDBProjection`, which stores the current state in a relational database. At this point I don't know if we will be able to serve current institution data from memory. I _think_ so, but I am not sure. At the same time, that is not enough to satisfy some of the panel and edits requirements. We need to know what the panel looks like every year. So a SQL view (filtering by year) on that SQL table makes a lot of sense. The SQL table can also serve the API if it gets too big to fit in memory in the future (in that case, the view would just be responsible for reading events from journal and streaming them to consumers). Having this data in a relational database will also help with analytics in the future.

jmarin: Oh, and in case you are missing some design there this is a known issue, the institutions model is not complete. For example the SQL view I'm referring to can't be done with the data that we have right now, but it will be there in the future (i.e. we will know when a create / modify event happened for an institution, and will store that).

kgudel:
This LGTM!

PR #662, opened 2016-11-02
Changed macro API output 

### LINKED ISSUE #634: Change output for macro edits API
Macro edits don't have a list of `LAR` that fail the edit check (they are applied to all the file). Modify the API so that it only lists the macro edits, without an empty list of `LAR` in each reported error.
Depends on #633

nickgrippin:
Closes #634

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -255,10 +255,10 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
       },
       "validity": {},
       "quality": {},
-      "macro": {
+      "macro": [
           "Q007",
           "Q008"
-      }
+      ]
     }
     ```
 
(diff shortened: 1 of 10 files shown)


awolfe76:
@nickgrippin - can we update the API docs because of this PR? To add a macro response?

nickgrippin:
@awolfe76: sure thing!

nickgrippin:
So, this isn't going to match what the front end is currently using, because we don't have a model for justifications yet.  I'm not sure if that should belong in a separate ticket, or if the ticket should be changed to "match front-end api" as opposed to "list macro edits".

awolfe76:
I think create a new issue to get the format right. This is a great first step. Thanks!

nickgrippin:
Created #663

PR #665, opened 2016-11-03
Remove yield block that was causing test to not be evaluated

### LINKED ISSUE #657: UploadPathsSpec test not failing
Currently, the last test in `UploadPathsSpec` ("return 400 when trying to upload to a completed submission") will not fail no matter what object the test compares it to.  Replacing the `ErrorResponse` object with `Nil`, for example, will not cause the test to fail.

The test should be fixed so that the comparison is actually tested.

kgudel:
Closes #657


nickgrippin:
LGTM!  Nice fix!

PR #668, opened 2016-11-08
clean up InstitutionType

### LINKED ISSUE #534: clean up InstitutionType
1. Make its `depositoryType` field no longer an `Option`.  We have determined that every InstitutionType is made up entirely of either depository or non-depository institutions. (One `InstitutionSpec` test case should go away as a result.)
2. Remove the two temporary ones used for testing (indicated by FIXME).  We have determined that the others cover all the necessary real-world cases. Some tests will need to be updated to use other existing Institution Types.
3. Remove no-longer-needed case class from `InvalidRespondentId` children.

kgudel:
Closes #534


nickgrippin:
LGTM!

PR #669, opened 2016-11-09
Add timestamp to Filing and Submission

### LINKED ISSUE #644: Add timestamps to endpoints
I think it would be nice if the following endpoints returned timestamps (start and end). One place we were going to use this data on the front-end is on the list of institutions to show when a filing/submission began and ended. [Here's an example](http://awolfe76.github.io/hmda-platform-prototype/) from a very early prototype. You can see things like "Filing started about 22 minutes ago" or "Filing started on Feb 21, 2016" or "Filing started Feb 9, 2016 and completed Feb 9, 2016". Also there is a timestamp beside available downloads for previous submission attempts.

_Property names can be changed. Just used `start` and `end` for now._
- [ ] `/institutions/<institution>` `GET`
  Example to show when a filing began and was completed:

``` json
"filings": [
    {
        "period": "2017",
        "institutionId": "12345",
        "status": {
            "code": 3,
            "message": "completed"
        },
        "start": 1457494448191,
        "end": 1457494448191
    }
]
```
- [ ] `/institutions/<institution>/filings/<period>` `GET`
  Example to show when filing began and was completed, along with when a submission began and was completed:

``` json
{
    "filing": {
        "period": "2017",
        "institutionId": "12345",
        "status": {
            "code": 1,
            "message": "not-started"
        },
        "start": 1457494448191,
        "end": 1457494448191
    },
    "submissions": [{
        "id": {
            "institutionId": "12345",
            "period": "2017",
            "sequenceNumber": 1
        },
        "status": {
            "code": 1,
            "message": "created"
        },
        "start": 1457494448191,
        "end": 1457494448191
    }]
}
```

---

Other possible places to include the timestamps (_don't think that these are required but may make sense to keep things consistent, especially if some of these responses are created from the same code base_):
- in the response from `/institutions/<institution>/filings/<period>/submissions` `POST`, when a new submission is created

``` json
{
    "id": {
        "institutionId": "0",
        "period": "2016",
        "sequenceNumber": 1
    },
    "status": {
        "code": 1,
        "message": "created"
    },
    "start": 1457494448191,
    "end": null
}
```
- in the response from `/institutions/<institution>/filings/<period>/submissions/latest` `GET`, while we are polling

``` json
{
    "id": 3,
    "status": {
        "code": 1,
        "message": "created"
    },
    "start": 1457494448191,
    "end": null
}
```

nickgrippin:
Closes #644


persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
```diff
@@ -54,7 +54,7 @@ class SubmissionPersistence(institutionId: String, period: String) extends HmdaP
     case CreateSubmission =>
       val seqNr = state.submissions.size + 1
       val submissionId = SubmissionId(institutionId, period, seqNr)
-      val newSubmission = Submission(submissionId, Created)
+      val newSubmission = Submission(submissionId, Created, 0L, 0L)
```
kgudel: I believe that this line assigns all `Submission`s a creation time of `0L`. Is that correct?

kgudel:
I believe in `submissionPersistence` we also have to update the `SubmissionStatusUpdated` method in the `SubmissionState` class so that we can change the end time to the current time if the status is changed to `Signed`.

nickgrippin:
@kgudel: Your suggestion should be done in 3d85191afaa029428e76dfbe03a5598c8b39456f

api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala
```diff
@@ -85,17 +85,17 @@ trait UploadPaths extends InstitutionProtocol with ApiErrorProtocol with Submiss
           case Success(response) =>
             processingActor ! CompleteUpload
             processingActor ! Shutdown
-            complete(ToResponseMarshallable(StatusCodes.Accepted -> Submission(id, Uploaded)))
+            complete(ToResponseMarshallable(StatusCodes.Accepted -> Submission(id, Uploaded, System.currentTimeMillis(), 0L)))
```
kgudel: Now that a start time stamp is being saved for each submission when it is created we should probably return that time here instead of the current time.

api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala
```diff
@@ -67,12 +67,11 @@ trait UploadPaths extends InstitutionProtocol with ApiErrorProtocol with Submiss
       }
     }
 
-  private def checkSubmissionIsCreated(submissionsActor: ActorRef, submissionId: SubmissionId)(implicit ec: ExecutionContext): Future[Boolean] = {
-    val submission = (submissionsActor ? GetSubmissionById(submissionId)).mapTo[Submission]
-    submission.map(_.status == Created)
+  private def checkSubmissionIsCreated(submissionsActor: ActorRef, submissionId: SubmissionId)(implicit ec: ExecutionContext): Future[Submission] = {
```
kgudel: Since this function is now just returning a `Future[Submission]` it probably shouldn't be called `checkSubmissionIsCreated`. It might also just be worthwhile to just remove it all together and just put its logic in the `for` block since it's only the one line.

PR #672, opened 2016-11-09
Sbt upgrade

jmarin:
Faster builds!
Preparation for upgrade to Scala 2.12 (which we can't do yet)


kgudel:
LGTM! I'm looking forward to the faster builds!

PR #682, opened 2016-11-17
Remove local event stream processor

jmarin:
Replaces `LocalHmdaEventProcessor`, which is gone now, with `SubmissionManager`, an actor that plays ping pong sending messages with its children as the filing process progresses (upload -> parse -> validate). Also introduces a [Finite State Machine](https://en.wikipedia.org/wiki/Finite-state_machine) for keeping track of the submission state. This is implemented in the `SubmissionFSM` class as an `Akka` persistent `FSM` actor. This class is now responsible for updating the status of submissions in `SubmissionPersistence`.


nickgrippin:
I'm getting about 10 dead letter messages when I run the api:
```
api [2016-11-23 13:08:22,769][INFO ][hmda-akka.actor.default-dispatcher-5] Message [scala.Some] from Actor[akka://hmda/user/$a#1961947287] to Actor[akka://hmda/deadLetters] was not delivered. [1] dead letters encountered. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'
```

persistence/src/main/scala/hmda/persistence/processing/SubmissionFSM.scala
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  when(ValidatedWithErrors) {
+    case Event(GetState, data) =>
+      stay replying data
+    case Event(_, data) =>
+      stay replying data
+  }
+
+  whenUnhandled {
```
kgudel: I think that it would be good to remove the `Event(_,data)` cases throughout the `FSM` and let `whenUnhandled` take care of `GetState` calls and unexpected events, since currently in many states it is not possible get the unhandled error. Also, I think it will be good to make it explicate that only a `GetState` call will return the state as currently, for most states, any unhandled call will return the state.

jmarin:
@nickgrippin The `dead letters` error is expected since the `Filing` semantics were changed, see #696

persistence/src/main/scala/hmda/persistence/processing/SubmissionFSM.scala
jmarin: I think there might be a misunderstanding here. From the Akka docs: _"we make use of the fact that any event which is not handled by the when() block is passed to the whenUnhandled() block"_. So `whenUnhandled` is sort of a "catch all" for events that might happen in any state, hence why `GetState` is placed here as default. As for the `Event(_, data)` those are present whenever the `SubmissionPersistence` actor is called through the `updateStatus` message, if you follow the code you will see that a response with a `Some(submission)` is returned before the state machine transitions to the next state. Leaving that out causes errors to be logged in the console. Putting it into the `whenUnhandled` block masks the `case Event(e,d)` for catching all other events that might be sent to this state machine by mistake (which I would like to log separately).

nickgrippin:
LGTM, this will be immensely useful. I'm glad to be rid of the event processor.

PR #686, opened 2016-11-18
Remove unused parsers

### LINKED ISSUE #611: Okay to remove `FIDataCsvParser` class?
We've changed our strategy for reading and parsing HMDA files between the beginning of the project and now. Currently, we persist each line of the file individually (using `HmdaRawFile`), then parse each line separately (using `TsCsvParser`).  So, `FIDataCsvParser`, which has methods to break up a file into an `Iterable` and parse the whole thing, is not used. I believe it's a relic of our previous strategy.

Can `FiDataCsvParser` (and corresponding `FiDataDatParser`) be removed?

cc @jmarin

jmarin:
Closes #611


kgudel:
This LGTM!

PR #689, opened 2016-11-21
Clean up SubmissionBasePathsSpec

### LINKED ISSUE #684: SubmissionBasePathSpec failures
This test seems to fail occasionally, perhaps due to ordering of checks for data

nickgrippin:
The only way I can see this test failing is if the latest submission somehow (perhaps from a previous test) does not have a `Created` status.  I moved the `/latest` endpoint test beneath the one creating a new submission, so we should never run into that situation.

Closes #684


schbetsy:
I think there are some ways we could consider improving this test suite in the future (like cleaning out the database before it, or using a different test institution that isn't manipulated by the other specs), but for now, this fix makes sense to me.

kgudel:
I agree with @schbetsy . This is certainly something that we can revisit in the future, but this is a good fix! LGTM

PR #690, opened 2016-11-21
Return 404 for invalid requests to Edits / Parser Errors

### LINKED ISSUE #593: Return 404 When Appropriate from Edits Endpoint
On the Edits endpoint, currently if the request is for an institution/period/submission that doesn't exist, the endpoint simply responds with an empty set of edits. Instead, it should respond with a `404` and a message saying which part of their request was invalid.

e.g. `/institutions/66/filings/2017/submissions/1/edits` should return "Insitution 66 not found" or similar.

schbetsy:
Closes #593

### DIFF
--- a/api/src/main/scala/hmda/api/http/institutions/submissions/RequestVerificationUtils.scala
+++ b/api/src/main/scala/hmda/api/http/institutions/submissions/RequestVerificationUtils.scala
@@ -1,8 +1,15 @@
 package hmda.api.http.institutions.submissions
 
 import akka.actor.{ ActorRef, ActorSystem }
+import akka.http.scaladsl.marshalling.ToResponseMarshallable
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import akka.http.scaladsl.model.{ StatusCodes, Uri }
+import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.Directives._
 import akka.pattern.ask
 import akka.util.Timeout
+import hmda.api.http.HmdaCustomDirectives
+import hmda.api.model.ErrorResponse
 import hmda.model.fi.{ Filing, Submission, SubmissionId }
 import hmda.model.institution.Institution
 import hmda.persistence.HmdaSupervisor.{ FindFilings, FindSubmissions }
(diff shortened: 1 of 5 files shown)


api/src/main/scala/hmda/api/http/institutions/submissions/RequestVerificationUtils.scala
```diff
@@ -0,0 +1,78 @@
...
+    val submissionId = SubmissionId(institutionId, period, seqNr)
+
+    val supervisor = system.actorSelection("/user/supervisor")
+    val querySupervisor = system.actorSelection("/user/query-supervisor")
+
+    val fInstitutionsActor = (querySupervisor ? FindActorByName(InstitutionView.name)).mapTo[ActorRef]
+    val fFilingsActor = (supervisor ? FindFilings(FilingPersistence.name, institutionId)).mapTo[ActorRef]
+    val fSubmissionsActor = (supervisor ? FindSubmissions(SubmissionPersistence.name, institutionId, period)).mapTo[ActorRef]
+
+    for {
```
kgudel: From my current knowledge of how for comprehensions work with futures in scala I believe that they are created and evaluated sequentially in the comprehension and that they have to be created before the comprehension in order to be able to be run in parallel (I get that from [here](http://buransky.com/scala/scala-for-comprehension-with-concurrently-running-futures/)).  Therefor I believe that it would better to remove the `?` calls in the for comprehension and instead include them as a `flatMap` on each of the individual call to find `ActorRef`s so that they can be run as soon as their `ActorRef` is evaluated instead of waiting for the earlier calls to complete.

kgudel:
It might be useful to test the order of the hierarchy of errors, that `no institution` overrides `no submission` which overrides `no filing`.

api/src/main/scala/hmda/api/http/institutions/submissions/RequestVerificationUtils.scala
jmarin: @kgudel is right, those calls to `Future` should be moved out of the `for` comprehension to take advantage of parallel processing (as you have done with `fInstitutionsActor`, for instance)

jmarin: Actually, scratch that. This code is fine. The futures inside the for comprehension are fine, since they depend on the completion of the previous futures, they are not independent of each other and can't run all in parallel. Sorry for the confusion

api/src/main/scala/hmda/api/http/institutions/submissions/RequestVerificationUtils.scala
```diff
@@ -0,0 +1,78 @@
...
+      sa <- fSubmissionsActor
+      s <- (sa ? GetSubmissionById(submissionId)).mapTo[Submission]
+      msg <- getErrorMessage(i, f, s, institutionId, period, submissionId)
+    } yield msg
+  }
+
+  private def getErrorMessage(i: Institution, f: Filing, s: Submission, iid: String, p: String, sid: SubmissionId)(implicit ec: ExecutionContext): Future[Option[String]] = Future {
+
+    if (s.id == sid) None
+    else if (f.period == p) Some(s"Submission ${sid.sequenceNumber} not found for $p filing")
```
schbetsy: I clarified the logic a bit. Let me know whether it's enough.

jmarin:
LGTM

PR #693, opened 2016-11-22
Simplify Macro Tests

kgudel:
This PR is in response to the fact that many of our macro test suites are essentially the same with only changes to a couple of variables. In order to address this, this PR introduces a new trait `SimpifiedMacroSpec` which contains a function to run the three tests in most macro specs with the only a change to the necessary variables.


validation/src/test/scala/hmda/validation/rules/lar/macro/SimplifiedMacroSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+import hmda.model.fi.lar.LoanApplicationRegister
+
+trait SimplifiedMacroSpec extends MacroSpec {
+
+  def multiplier: Double
+
+  def relevantLar(lar: LoanApplicationRegister): LoanApplicationRegister
+  def irrelevantLar(lar: LoanApplicationRegister): LoanApplicationRegister
+
+  protected def simplifiedPropertyTests(name: String, multiplier: Double, relevantLar: LoanApplicationRegister => LoanApplicationRegister, irrelevantLar: LoanApplicationRegister => LoanApplicationRegister): Unit = {
```
nickgrippin: This is definitely a great way to trim down on duplicated code!  One thing I would suggest would be to rename this... Something like `lessThanOrEqualToPropertyTests`.  Just to give some idea of what we're expecting from this function, since it's specific to that behavior.

kgudel: Thanks! And good suggestion. I was having trouble thinking of a good name for it!

nickgrippin:
LGTM!

validation/src/test/scala/hmda/validation/rules/lar/macro/lessThanOrEqualToPropertyMacroSpec.scala
```diff
@@ -0,0 +1,37 @@
+package hmda.validation.rules.lar.`macro`
+
+import com.typesafe.config.ConfigFactory
+import hmda.model.fi.lar.LoanApplicationRegister
+
+trait lessThanOrEqualToPropertyMacroSpec extends MacroSpec {
+
```
jmarin: This goes against naming conventions, the name of a `trait` is usually capitalized

PR #700, opened 2016-11-29
Q048 and Q006

### LINKED ISSUE #216: Validation (2017): Q006 (Loan Purpose)
**Edit #:** Q006
**Type:** Macro
**Column:** Loan Purpose

If the total number of loan applications with loan purpose = 1 and action taken type = 1 is > 25, then the total number of these loan applications should be ≤ 95% of the total number of home purchase loan applications.

Additional resources:

### LINKED ISSUE #218: Validation (2017): Q048 (Preapproval)
**Edit #:** Q048
**Type:** Macro
**Column:** Preapproval

If preapproval = 1, and action taken type = 5, then the total number of these loans should be ≤ 5% of the total number of loan applications.

Additional resources:

kgudel:
Closes #216 
Closes #218


nickgrippin:
LGTM aside from merge conflicts

validation/src/test/scala/hmda/validation/rules/lar/macro/Q006Spec.scala
```diff
@@ -0,0 +1,61 @@
...
+  def relevantLar(lar: LoanApplicationRegister) = {
+    val relevantLoan = lar.loan.copy(purpose = 1)
+    lar.copy(actionTakenType = 1, loan = relevantLoan)
+  }
+
+  val irrelevantAmount: Gen[Int] = Gen.chooseNum(1, numOfOriginatedHomePurchaseLoans)
+  val relevantAmount: Gen[Int] = Gen.chooseNum(numOfOriginatedHomePurchaseLoans + 1, 10000)
+
+  property(s"be valid if fewer than $numOfOriginatedHomePurchaseLoans originated loans") {
+    forAll(irrelevantAmount) { (x) =>
```
schbetsy: Most other macro specs don't use `forAll` to run with randomly generated numbers of loans, they just choose a few values that test the edge cases of the specific edit.

I like that these specs cover more of the possible cases. I just want to make sure that it's okay that they don't follow the same pattern as the others, and that they'll be slower, since they're testing hundreds of cases instead of a few.

(for the record: I don't think you should have to change them to single tests; I just wanted to put that decision in the open)

validation/src/test/scala/hmda/validation/rules/lar/macro/Q006Spec.scala
```diff
@@ -0,0 +1,61 @@
...
+      val lars = larNGen(x).sample.getOrElse(Nil)
```
schbetsy: Could these be `get` rather than `getOrElse`?

validation/src/test/scala/hmda/validation/rules/lar/macro/Q006Spec.scala
```diff
@@ -0,0 +1,61 @@
...
+      val validLarSource = newLarSource(lars, x, relevantLar, irrelevantLar)
+      validLarSource.mustPass
+    }
+  }
+
+  property(s"be valid if more than $numOfOriginatedHomePurchaseLoans originated loans and originated < $multiplier * total") {
+    forAll(relevantAmount) { (x) =>
+      val numOfLars = (x / multiplier).toInt + 1
```
schbetsy: I got confused for a while by the `x / multiplier` logic. Maybe adding `total` to the name of that variable could prevent some future confusion?

validation/src/test/scala/hmda/validation/rules/lar/macro/Q006Spec.scala
kgudel: `sample` does return an `option` so it would make sense to provide for a case when the `option` is empty. However, I'm not actually sure why it returns an `option` and so why the option might be empty, so it might not be necessary.

@jmarin

nickgrippin: Most of these will return options so that if the requirements for generation are too strict, it can possibly return empty values.  Our LAR generators are pretty open ended, so we should never encounter a case where we can't get a full sample.

schbetsy: `get` will throw an error if the `Option` is empty. Also, the test will throw errors if we try to run it with `Nil` as the set of lars. I don't see a benefit to adding `Nil` into the spec.

validation/src/test/scala/hmda/validation/rules/lar/macro/Q006Spec.scala
schbetsy: Are you leaving the tests as-is?

kgudel: I think that I would prefer to leave the test as is for as I have not seen it take a substantial length of time.

PR #705, opened 2016-11-30
Institution repository

jmarin:


schbetsy:
After a first pass of review, the code here is looking good. But I'm not sure I understand the reasoning behind the changes here. Would you mind adding a few sentences to explain? (Or you could link to a resource that explains this pattern, whatever you prefer).

kgudel:
The is also LGTM with my understanding.

jmarin:
This is basically a refactor of our previous code. It separates concerns by extracting the operations in `InstitutionRepository` and leaving the table structure in `InstitutionsTable`. It also makes it easier to extend for future repository implementations (though repository implementation is still not fully generic; that will come in the near future). More importantly, it makes it a lot easier to configure the database being used in the projection. Now it's just a matter of configuring in `application.conf`, the database driver can be loaded through `DatabaseConfig`.

schbetsy:
Thanks for the explanation. LGTM.

Issue #709, opened 2016-12-01
Clarify wording in `RequestVerificationUtils`


schbetsy:
When writing error messages that could be visible to the user, let's use "filing period" instead of "filing" for `Filing` objects. That will be more consistent with the language used on the front end. ( @awolfe76 and I settled on this decision)

This should be a quick change. Just update all instances in `RequestVerificationUtils` and any corresponding specs.

PR #713, opened 2016-12-02
Update filing to say filing period

kgudel:
#709


schbetsy:
@awolfe76 Do you want to double check this?

PR #716, opened 2016-12-05
S270

### LINKED ISSUE #131: Validation (2017): S270 (Action Taken - Date)
**Edit #:** S270
**Type:** Validity
**Column:** Action Taken - Date

Century (CC) and Year (YY) of action taken date must = activity century/year (CCYY) for period being processed.

Additional resources:

schbetsy:
Closes #131

I've implemented S270 to use the exact same pattern as S100 (TS). The activity year comes from the validation context. If the validation context doesn't provide a year, this edit is treated as optional and automatically passes.

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/syntactical/S270.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/syntactical/S270.scala
@@ -1,0 +1,23 @@
+package hmda.validation.rules.lar.syntactical
+
+import hmda.model.fi.lar.LoanApplicationRegister
+import hmda.validation.context.ValidationContext
+import hmda.validation.dsl.PredicateCommon._
+import hmda.validation.dsl.PredicateSyntax.PredicateOps
+import hmda.validation.dsl.Result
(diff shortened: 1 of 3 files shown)


validation/src/main/scala/hmda/validation/rules/lar/syntactical/S270.scala
```diff
@@ -0,0 +1,23 @@
...
+import hmda.validation.rules.{ EditCheck, IfYearPresentIn }
+
+object S270 {
+  def inContext(ctx: ValidationContext): EditCheck[LoanApplicationRegister] = {
+    IfYearPresentIn(ctx) { new S270(_) }
+  }
+}
+
+class S270 private (year: Int) extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "S270"
```
kgudel: I don't believe that this needs the `override` tag.

schbetsy: The method does override a method on the abstract class, though. Why do you think I should remove it?

PR #727, opened 2016-12-07
Generic repository

jmarin:
This PR refactors the DB access layer to make it generic and promote code reuse and sensible default implementation for commonly used database access methods. With these changes, mapping a new entity to a RDBMS table now requires:

1. Creating a new case class modeling the entity. See `hmda.query.model.institutions.InstitutionQuery` for an example
2. Creating a new trait that includes the `Slick` definition for a table, as well as the repository for this entity. This repository extends from the generic `hmda.query.repository.Repository` and it inherits default access methods (find by id, insert, update, etc.). See `hmda.query.repository.institutions.InstitutionComponent` for an example. 
3. Use the component in client code (see `InstitutionComponentSpec` or `InstitutionDBProjection` for an example

The implementation remains agnostic as to which database technology is used (defined in configuration)


schbetsy:
This is a smart change, I like it! I just had a small question (above) about the organization of methods.

query/src/main/scala/hmda/query/repository/institutions/InstitutionComponent.scala
```diff
@@ -0,0 +1,34 @@
...
+    def filingPeriod = column[Int]("period")
+
+    override def * = (id, name, cra, agency, institutionType, hasParent, status, filingPeriod) <> (InstitutionQuery.tupled, InstitutionQuery.unapply)
+  }
+
+  class InstitutionRepository(val config: DatabaseConfig[JdbcProfile]) extends Repository[InstitutionsTable, String] {
+    val table = TableQuery[InstitutionsTable]
+    def getId(table: InstitutionsTable) = table.id
+
+    def createSchema() = db.run(table.schema.create)
```
jmarin: The difference is that I could not figure out an easy way to make them generic

schbetsy:
LGTM

kgudel:
Other than some merge conflicts LGTM!

nickgrippin:
Tested, looks good to me.

Issue #720, opened 2016-12-05
Change institution status?


awolfe76:
_More of a question to get other thoughts but **no rush**_

The `institution` endpoint returns:
```json
{
  "id": "12345",
  "name": "First Bank",
  "status": {
    "code": 1,
    "message": "active"
  }
}
```
where `active` means "on panel". 

Can we change that message to `on-panel` or something similar to make it more clear? To me, `active` means the bank is still doing business. 

Or, as @hkeeler [points out](https://github.com/cfpb/hmda-platform/issues/720#issuecomment-264975710) maybe we remove it from the `/institutions` endpoint and instead make changes to the `/institutions/<institution>` endpoint and the `institutions/<institution>/filings/<period>` endpoint.

Same for the `code: 0` where the message is `inactive`. To me, that sounds like the institution is not doing business anymore.

Maybe 
`0`: `not-on-panel`
`1`: `on-panel`

Or maybe it should just be something like (take from https://github.com/cfpb/hmda-platform/pull/411#issuecomment-226312782):
```json
"filings": [
  {
    "period": "2017",
    "institutionId": "12345",
    "required": true,
    "start": 0,
    "end": 0
  }
]
```
That is found in both the `/institutions/<institution>` and  `institutions/<institution>/filings/<period>` endpoints.

This would require updates to:
- [ ] [institution status](https://github.com/cfpb/hmda-platform/blob/master/Documents/institution-status.md)
- [ ] [api](https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md)

hkeeler:
`institution.status` was not (originally) intended to indicate whether an institution is on/off panel.  It was a more general indicator of whether an institution that is in our system and perhaps _used_ to file, but is no longer _active_, likely due to it being merged with some other institution, going out of business, etc.  There have been several questions about this in the past, and seems more confusing than useful at this point.  I think we may want to remove it all together for now.  We'd won't be in this situation in year-one anyway.

In my opinion, the of/off panel indicator should be on the `institution/<institution_id>/filing/<period>` resource.  That way you can look at each filing over time for a given institution, and see if it was required or not for that period.  Way back on https://github.com/cfpb/hmda-platform/pull/411#issuecomment-226312782 we discussed something like this.

awolfe76:
> In my opinion, the of/off panel indicator should be on the institution/<institution_id>/filing/<period> resource.
👍

schbetsy:
[...]

PR #728, opened 2016-12-07
Remove InstitutionStatus

schbetsy:
This resolves half of #720. This PR removes `InstitutionStatus` from the Institution model (and therefore from the API responses, specs, and everything in between). See the linked issue for more info.

PR #731 completes the second half of #720, where we're adding a "filing-required" field to the Filing object.


awolfe76:
Should the [API.md](https://github.com/cfpb/hmda-platform/blob/9eb815e8c7792e3f1c4bf587eaf605164113bc53/Documents/API.md) be updated with the PR?

schbetsy:
Good call! Updated in 23d53dc.

Issue #717, opened 2016-12-05
Adding field data to edits endpoints


awolfe76:
See https://github.com/cfpb/hmda-platform-ui/issues/96

To help the filers read edits easier on the front-end, the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits` and `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<syntactical|validity|quality|macro>` endpoints need field metadata added.

I _think_ we would need a `desc` for the edit itself,  a list of `fields` (with the `name` and `friendlyName`) the edit is applied to, and a list of `values` (with the `value` and the `field`) entered for each.

__Possible__ response example:
```json
{
    "edit": "S025",
    "ts": true,
    "desc": "A friendly description of this edit",
    "fields": [
        {
            "name": "fieldName",
            "friendlyName": "Field name"
        },
        {
            "name": "fieldName2",
            "friendlyName": "Field name 2"
        }
    ],
    "lars": [
         {
            "lar": {
                "loanId": "s1",
                "values": [
                    {
                        "value": "value",
                        "field": "fieldName"
                    },
                    {
                        "value": "value",
                        "field": "fieldName2"
                    }
                ]
            }
        }
    ]
}
```

_Note: I'm sure there are other ways to implement this, including adding new endpoints. Happy to hear suggestions._

wpears:
I'd probably prefer the `lars` structured like:

```
"lars": [
  {
    "loanId": "s1",
    "field1": "field1value",
    "field2": "field2value"
   }
]
```

which, instead of having an object with a single `lar` key which refers to the actual lar object, just includes the lar object. Also, rather than having the lar object's field values be stuck in the `values` array, they are just included directly as part of the object.

schbetsy:
So the new format for the edits endpoint looks like this?

I wanted to confirm the format of the endpoint in general. Also, specifically, I'm wondering about the field names in the `"rows"` objects. Should they be the friendly names?

```json
{
  "syntactical": {
    "edits": [
      {
        "edit": "S010",
        "description": "The first record identifier in the file must = 1 (TS). The second and all subsequent record identifiers must = 2 (LAR)",
        "rows": [
          {
            "loanId": "Transmittal Sheet",
            "Record Identifier": "2"
          },
          {
            "loanId": "s11234",
            "Record Identifier": "8"
          }
        ]
      },
    ]
  },
  "validity": {
    "edits": [
      {
        "edit": "S215",
[...]

Issue #724, opened 2016-12-06
Add data for edit/field metadata


awolfe76:
Add the data required to fill in the metadata setup in #717.

Attached is a `.csv` (uploaded as a `.txt` because that's what Github likes) with the required data. There are some extra fields in the file that may be used later. To complete this task the "Field Names", "Edit Number", and "User Friendly Edit Description" will be needed.

The "User Friendly Edit Description"  will be the `"desc": "A friendly description of this edit",` part of #717.

The "Field Names" (comma separated in the string) will be used to fill in the `"friendlyName":` in:
```json
"fields": [
        {
            "name": "fieldName",
            "friendlyName": "Field name"
        },
        {
            "name": "fieldName2",
            "friendlyName": "Field name 2"
        }
    ],
```

The `.csv` does **not** contain the `name` for the field. I'm not sure where that will be defined, maybe already has been on the backend, but can be added if needed.

[edit-metadata1.txt](https://github.com/cfpb/hmda-platform/files/646387/edit-metadata1.txt)

schbetsy:
I noticed in the attached file that the `fieldNames` column only lists one field for each edit, though most edits have several relevant fields. For example, V275 depends on `applicationDate` AND `actionTakenDate`, but only "Action Taken Date" is listed in the file. 

So, (if my understanding of the desired UI is correct) I don't think we can rely on that file to get all of the fields we need for each edit. 

(I already added this comment [here](https://github.com/cfpb/hmda-platform/pull/732#discussion_r91789843), but this is a more relevant place to put it).

schbetsy:
Bumping this. I'm currently working on adding an edit's relevant fields to the API, so it would be nice to know where the edit metadata will come from.

Will we be able to get a file that lists all of an edit's relevant fields? If so, how will it be formatted?

schbetsy:
Resolution: @lesgou will add relevant fields to the edit-metadata file, separated by semicolons.

lesgou:
Updated file:

[edit-metadata1.txt](https://github.com/cfpb/hmda-platform/files/693694/edit-metadata1.txt)

awolfe76:
@schbetsy - can this one be closed?

schbetsy:
Yep, good catch. This was necessary for #770 and is safe to close.

PR #732, opened 2016-12-08
Add edit description

kgudel:
Address #717 

This PR is the first piece of #717 it adds a description field to each edit which is persisted through when an edit fails and is passed through the `ValidationErrorConverter` to be returned through the API.


jmarin:
I've only taken a quick look at this, but I think this approach could be improved. Ticket #717 is asking for additional metadata about the edits. This is all about providing the user with additional, human readable descriptions of the edits (as per https://github.com/cfpb/hmda-platform-ui/issues/96). I don't think this needs to be baked into the edit model, but rather be a lookup that gets included into the `json` response. At the persistence and validation level, we don't care about these messages. It's only on the presentation layer (api project) where we need to incorporate these. On ticket #724 the additional fields and content are described in a file. A possible approach might be to load this when necessary, and link to the validation errors by edit id. I don't think we want to persist all these fields, since they will most likely change (and it's extra stuff in the journal that we don't need)

kgudel:
@jmarin Those are really good points. I hadn't considered the burden of persisting all of that extra information. I have refactored it to get the descriptions from a flat file which has the added benefit of already having all of the edit descriptions in it preventing the need for a task to add descriptions to each edit individually.

Documents/API.md
```diff
@@ -275,7 +277,8 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
         {
             "edits": [
                 {
-                   "edit": "q007",
+                   "edit": "Q007",
+                   "description: "Description of Q007",
```
schbetsy: This `"description` and two others below are missing closing quotes, which is why the highlighting here is weird.

validation/src/main/scala/hmda/validation/engine/ValidationError.scala
```diff
@@ -7,7 +7,8 @@ case object Quality extends ValidationErrorType
 case object Macro extends ValidationErrorType
 
 // errorID = Loan ID (LAR) or Agency Code + Respondent ID (TS)
-case class ValidationError(errorId: String, name: String, errorType: ValidationErrorType)
+case class ValidationError(errorId: String, metaData: ValidationErrorMetaData, errorType: ValidationErrorType)
+case class ValidationErrorMetaData(name: String)
```
schbetsy: I don't understand yet what the `ValidationErrorMetaData` class is adding. Will it eventually hold more metadata than just the edit name?

model/jvm/src/main/scala/hmda/model/edits/EditMetaDataLookup.scala
```diff
@@ -0,0 +1,42 @@
...
+      )
+
+    }.toSeq
+  }
+}
+
+case class EditMetaData(
+  category: String,
+  editType: String,
+  fieldNames: String,
```
schbetsy: I noticed in the file that `fieldNames` only lists one field for each edit, though most edits have several relevant fields. (for example, `V275` depends on `applicationDate` and `actionTakenDate`, but only "Action Taken Date" is listed in the file.)

This isn't relevant for this PR, but it seems like it will become relevant soon.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -1,35 +1,45 @@
...
+import hmda.model.edits.{ EditMetaData, EditMetaDataLookup }
 import hmda.validation.engine._
 
 trait ValidationErrorConverter {
 
+  val editDescriptions = EditMetaDataLookup.values
+
   def validationErrorsToEditResults(tsErrors: Seq[ValidationError], larErrors: Seq[ValidationError], validationErrorType: ValidationErrorType) = {
+
+    def findEditDescription(editName: String): String = {
```
kgudel: Absolutely, this also opens up the ability to individually test this function.

validation/src/main/scala/hmda/validation/engine/ValidationError.scala
schbetsy: Would it be possible (without a large time commitment) to remove `ValidationErrorMetaData` from this PR? Since that class won't be used until a future PR, I think it doesn't make sense to code review and merge it now.

I don't want to waste your time doing git stuff, but I also want to wait and evaluate `ValidationErrorMetaData` in context, since it's a big change in a confusing part of the code.

api/src/test/scala/hmda/api/model/ModelGenerators.scala
```diff
@@ -12,7 +12,7 @@ trait ModelGenerators {
     for {
       status <- Gen.oneOf("OK", "SERVICE_UNAVAILABLE")
       service = "hmda-api"
-      time = Calendar.getInstance().getTime().toString
+      time = Calendar.getInstance().getTime.toString
```
jmarin: Why this change? I find that with `Java` getters it's easier to read if we keep the parens as in the original language / api. Either that, or also change `getInstance()` to `getInstance`

validation/src/main/scala/hmda/validation/rules/lar/quality/Q068.scala
```diff
@@ -19,4 +19,5 @@ object Q068 extends EditCheck[LoanApplicationRegister] with ApplicantUtils {
         (app.coSex not equalTo(4))
     }
   }
+
```
jmarin: Same comment on these lines, seem to unnecessarily add a blank space to existing code, without changing behavior

schbetsy:
Any more comments on this? @jmarin ?

jmarin:
No. I'm good

Issue #718, opened 2016-12-05
Allow edits to be returned by row


awolfe76:
See https://github.com/cfpb/hmda-platform-ui/issues/272

We need to allow filers to be able to sort edits by LAR row.

One way to handle this would be to add a `sortBy` parameter to the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits` endpoint. This endpoint would default to the current response, but a request to `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits?sortBy=row` would return the edits grouped by LAR row.

__Possible__ json response:
```json
{
  "rows": [
    {
      "rowId": "s1",
      "edits": [
        {
          "edit": "S025",
          "desc": "A friendly description of this edit",
          "fields": [
            {
              "name": "fieldName",
              "friendlyName": "Field name"
            },
            {
              "name": "fieldName2",
              "friendlyName": "Field name 2"
            }
          ],
          "values": [
            {
              "value": "value",
              "field": "fieldName"
            },
            {
              "value": "value",
              "field": "fieldName2"
            }
          ]
        }
      ]
    }
  ]
}
}
```

schbetsy:
This should wait until #717 is finished to avoid conflicts.

schbetsy:
_Not sure if this was already discussed elsewhere. If so, let me know_

I'm curious: why are the `fields` and `values` separate JSON objects in the response? Is that better than combining them, from an API usability perspective? Here's a possible example where they're combined (I substituted realistic values into the fields, so I could understand better):

```json
{
  "rowId": "s1",
  "edits": [
    {
      "edit": "V210",
      "desc": "Date application received must be in YYYYMMDD format; month and day must be in the range 01-12 and 01-31, respectively or equal NA. YYYY must be numeric and the first two digits of the number must equal 20.",
      "fields": [
        {
          "name": "loan.applicationDate",
          "friendlyName": "Date Application Received",
          "value": "19991231"
        }
      ]
    }
  ]
}
```

schbetsy:
Another question: Does this endpoint not need to separate the edits by edit type? From your example, it looks like we can put all of the edits in the same list, regardless of whether they're Syntactical, Validity, or Quality. 

Should the Edits By Row endpoint return Macro edits at all?

awolfe76:
@schbetsy for the first question, I think the conversation in #717 is related.

> Does this endpoint not need to separate the edits by edit type?

[...]

PR #734, opened 2016-12-08
Use same protocol as edits for single lar

kgudel:


schbetsy:
I wonder if the single LAR validation should use the same protocol as the Edits By Row endpoint (which isn't created yet. It's described in #718). 

Using `SummaryEditResults` has a few usability issues that we could avoid if we use the Edits By Row protocol. (I've pasted a sample response below for reference). Since this endpoint only analyzes one LAR, the `"ts": false` is unnecessary, as is the list of `"lars":` and list of Macro edits.

```json
{
  "syntactical": {
    "edits": []
  },
  "validity": {
    "edits": [
      {
        "edit": "V325",
        "ts": false,
        "lars": [
          {
            "lar": {
              "loanId": "ABCDEFGHIJKLMNOPQRSTUVWXY"
            }
          }
        ]
      }
    ]
  },
  "quality": {
    "edits": []
  },
  "macro": {
    "edits": []
  }
}
```

kgudel:
This makes sense to me! Closing.

PR #735, opened 2016-12-08
WIP: Add lar level fields

### LINKED ISSUE #717: Adding field data to edits endpoints

### LINKED ISSUE #721: Update edits response to let TS be part of the edits []
_Just a question, no rush_

#623 added `ts: true` to the `/edits` and `/edits/<type>` endpoints, like this:

```json
"edits": [
  {
    "edit": "S025",
    "ts": true,
    "lars": [
      {
        "lar": {"loanId": "s1"}
      },
      {
        "lar": {"loanId": "s2"}
      },
      {
        "lar": {"loanId": "s3"}
      }
    ]
  }
]
```

Rather than have `ts: true` it might be better remove the `ts:true` and rename the `lars: []` to `rows: []` and then change the `lar: {}` to `row: {}` and `loanId` to `rowId`. (`rowId` would either be "Transmittal Sheet" or the loan ID.) This would let us just have the transmittal sheet be a normal part of the edits.

```json
"edits": [
  {
    "edit": "S025",
    "rows": [
      {
        "row": {"rowId": "Transmittal Sheet"}
      },
      {
        "row": {"rowId": "s2"}
      },
      {
        "row": {"rowId": "s3"}
      }
    ]
  }
]
```

Thoughts?

kgudel:
Close #717 

This PR finishes implementing #717 by adding a list within each lar of each edit which gives the names of relevant fields and the values which caused the edit to fail.

As a side effect, in order to return the fields for the transmittal sheet the transmittal sheet had to be treated like any other line and so this PR as Closes #721

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -249,17 +249,34 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
                         "description": "None"
                       }
                     ],
-            "ts": true,
             "lars": [
+               {
+                "larId": "Transmittal Sheet",
+                "fields": [
+                  {
+                    "name": "None",
+                    "value": ""
+                  }
+                ]
+               },
               {
-                "lar": {"loanId": "s1"}
-              },
-              {
-                "lar": {"loanId": "s2"}
-              },
-              {
-                "lar": {"loanId": "s3"}
-              }
+                          "larId": "ABCDE12345ABCDE12345ABCDE",
+                          "fields": [
+                            {
+                              "name": "None",
+                              "value": ""
+                            }
+                          ]
+                        },
+                        {
+                          "larId": "ABCDE12345ABCDE12345ABCDE",
+                          "fields": [
+                            {
+                              "name": "None",
+                              "value": ""
+                            }
+                          ]
+                        }
             ]
           },
           {
@@ -274,11 +291,23 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
             "ts": false,
             "lars": [
               {
-                "lar": {"loanId": "s4"}
-              },
-              {
-                "lar": {"loanId": "s5"}
-              }
+                          "larId": "ABCDE12345ABCDE12345ABCDE",
+                          "fields": [
+                            {
+                              "name": "None",
+                              "value": ""
+                            }
+                          ]
+                        },
+                        {
+                          "larId": "ABCDE12345ABCDE12345ABCDE",
+                          "fields": [
+                            {
+                              "name": "None",
+                              "value": ""
+                            }
+                          ]
+                        }
             ]
           }
         ]
@@ -328,10 +357,14 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
               ],
       "lars": [
         {
-          "lar": {
-            "loanId": "4977566612"
-          }
-        }
+                     "larId": "ABCDE12345ABCDE12345ABCDE",
+                     "fields": [
+                       {
+                         "name": "None",
+                         "value": ""
+                       }
+                     ]
+                   }
       ]
     },
     {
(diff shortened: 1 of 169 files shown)


jmarin:
See my comments on the previous PR of this series

kgudel:
@jmarin because the front end wants the relevant values for each row that failed the check I believe that we do need to be persisting some amount of information. At least the names of the relevant fields and the values of the fields. Since this is almost all that the front end wants (just friendly name of the field as well) it seems reasonable to me to persist everything through and return the information from the `ValidationError`, what do you think?

jmarin:
I disagree with the way this requirement is laid out and / or interpreted. The edits tell you the valid values. that spec is clear. We will give the user a list of edits that failed, and user friendly language explaining why the edit failed. Storing the fields and values that produced that error is redundant imho. Especially considering that many FIs take the list (i.e. as a csv) and go back to their system of record (where they can see the values that produced the error).

@awolfe76 let me know what you think. Adding this bloats up the persistence in an unnecessary way (we're not adding anything new). Another way to think about this is as follows: we can produce an API endpoint that allows the user to see all the fields and values for a loan id, if that is what they want. The ui could then filter based on the list of fields, or highlight the values for that loan id. There is no need to store the fields and values with each and every edit.

awolfe76:
Just want to make sure I understand.

The `/edits` endpoints would return a list of edits, the edit friendly description, and the fields names and friendly description associated with each edit.

Then the front-end would go through each `row` (currently `lar`), make an API call to get details about that row which would return `field:value` pairs. The front-end could then show only the necessary values, the values for the fields that apply to the edit.

That seems ok. It could be a good bit of traffic is all, one call for each lar per edit.

Might help if that endpoint could be filtered based on `editId`. So if we add `?editId=editId` we could only get back the relevant fields. Would still have to make a call for each lar and TS that had edits though, it would just trim down the response.

---

Is it possible to do something similar to what this PR is proposing, but instead of storing the fields and values populate them from the records that are already stored? 

That seems like a combination of your proposal and what is already there. So rather than the front-end looping through the results, making the API calls, and matching to what we need, can that be done on the back-end? Seems like that would save some traffic between front and back-end.

jmarin:
>Is it possible to do something similar to what this PR is proposing, but instead of storing the fields and values populate them from the records that are already stored?

Yes, and this is kind of where I was going with this. Another way to do this is to project the data into a table (or elasticsearch document) where we have the structure that we need for the UI. This will require more work but is way more efficient. We're lacking some stuff in the backend to support this at this point. I think the approach proposed here takes a shortcut, but the long term costs are just too high

awolfe76:
Thanks @jmarin. Until we have what we need on the back-end to do it better, is it possible to just mock this? Put some field:value pairs so we can get the UI working?

kgudel:
I'm going to close this one due to the approach that is takes. @jmarin do we want to create some kind of mocking on edit returns as suggested by @awolfe76 ?

PR #743, opened 2016-12-15
Macro verify

### LINKED ISSUE #725: Create the POST for verification for macro edits
Each macro edit requires verification. The UI will allow for multiple verifications for each edit. When each verification is selected the UI will `POST` to the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/macro` endpoint and send:
```json
{
  "edit": "M1",
  "justification": "The string for the selected justification."
}
```

The response should be:
```json
{
  "edit": "M2",
  "justifications": [
    "One string",
    "for every justification",
    "that has been selected",
    "including all previous selections"
  ]
}
```

The question I have for this is how does the UI know what the options are for the selections? Should we have them in the UI? Should they be part of the API?

In the docs for the `GET` for edits we have:
```json
"justifications": [
  {
    "value": "don't worry",
    "selected": false
  },
  {
    "value": "be happy",
    "selected": false
  }
]
```
Should that response have each possible selection in it? If so the json response to a `POST` would also need to have that same structure. (Maybe we change `"selected"` to `"verified"`.) 

@wpears - is that enough to allow us to update the state and be able render what we need?

_Related to See https://github.com/cfpb/hmda-platform-ui/issues/281_

jmarin:
Closes #725


Documents/API.md
```diff
@@ -318,6 +318,54 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
 }
 ```
 
+    * `POST` - Provides verification for macro edits
```
awolfe76: __Minor__: There looks like some odd formatting issue [here](https://github.com/jmarin/hmda-platform/blob/b1256c7dcae1c40c6d306b2f24902daa96a29797/Documents/API.md) ... below  the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<syntactical|validity|quality|macro>`

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -29,7 +29,12 @@ trait ValidationErrorConverter {
   }
 
   def validationErrorsToMacroResults(errors: Seq[ValidationError]): MacroResults = {
-    val macroValidationErrors: Seq[ValidationError] = errors.filter(_.errorType == Macro)
-    MacroResults(macroValidationErrors.map(x => MacroResult(x.name, List())))
+    val macroValidationErrors: Seq[ValidationError] = errors.filter(_.errorType == Macro).asInstanceOf[Seq[MacroValidationError]]
+    val macroEditNames = macroValidationErrors.map(x => x.name)
+    val macroJustifications = MacroEditJustificationLookup().justifications
```
nickgrippin: This value isn't being used anywhere

persistence/src/test/scala/hmda/persistence/processing/HmdaFileValidatorSpec.scala
```diff
@@ -93,14 +93,57 @@ class HmdaFileValidatorSpec extends ActorSpec with BeforeAndAfterEach with HmdaF
...
+          MacroValidationError("Q008", Nil),
+          MacroValidationError("Q010", Nil),
+          MacroValidationError("Q016", Nil),
+          MacroValidationError("Q023", Nil)
+        )
       ))
-
     }
 
+    "veify macro edits" in {
```
schbetsy: small typo in "verify"

Documents/API.md
```diff
@@ -289,34 +289,82 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
...
+    {
+      "edit": "Q023",
+      "justification": {
+      "id": 1,
+      "value": "Most of the loan activity are in areas outside of an MSA/MD",
+      "verified": true
+      }
+    }
+    ```
+        Example response, with HTTP code 200:
```
nickgrippin: Formatting on this line looks a little weird.

Documents/API.md
```diff
@@ -289,34 +289,82 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
...
+
+    * `POST` - Provides verification for macro edits
+
+    Example payload, in `JSON` format:
+
+    ```json
+    {
+      "edit": "Q023",
+      "justification": {
+      "id": 1,
```
nickgrippin: I think this might need to be indented again

nickgrippin:
Is there any kind of collision detection on adding new verifications?  In testing, I was able to add new verifications with duplicate IDs or duplicate values (but obviously not both).  This was my final result after a couple different `POST`s.
```
{
      "edit": "Q007",
      "justifications": [
        {
          "id": 2,
          "value": "There were a large number of applications, but few loans were closed",
          "verified": true
        },
        {
          "id": 3,
          "value": "Loan activity for this filing year consisted mainly of purchased loans.",
          "verified": false
        },
        {
          "id": 1,
          "value": "There were a large number of applications, but few loans were closed",
          "verified": true
        },
        {
          "id": 1,
          "value": "Applicants decided not to proceed with the loan.",
          "verified": true
        },
        {
          "id": 1,
          "value": "There were a large number of applications, but few loans were closed",
          "verified": false
        }
      ]
    }
```

jmarin:
@nickgrippin I didn't add any check for that. My understanding is that the id and value don't change, they are fixed. @awolfe76 is this assumption correct?

awolfe76:
We won't be changing them on the front-end. We'll be using both, but not changing them.
Using them like this:
```html
<select>
  <option value="id">value</option>
</select>
```

nickgrippin:
Duplication bug is fixed. 

LGTM!

schbetsy:
Since the front-end won't update the text of the justifications (as Andrew said), should we not enable that from the API? Currently, I can edit justifications and add new ones from the API:

```json
    {
      "edit": "Q008",
      "justifications": [
        {
          "id": 1,
          "value": "Changed this justification text to something nonsensical",
          "verified": true
        },
        {
          "id": 2,
          "value": "There were a large number of applications, but few loans were closed",
          "verified": false
        },
        {
          "id": 11234,
          "value": "added a new justification that isn't in the list",
          "verified": false
        }
      ]
    }
```

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
```diff
@@ -66,36 +67,56 @@ trait SubmissionEditPaths
   def submissionSingleEditPath(institutionId: String) =
     path("filings" / Segment / "submissions" / IntNumber / "edits" / Segment) { (period, seqNr, editType) =>
```
schbetsy: What would you think about splitting this endpoint definition into two separate ones? Just a question to consider. Also, if this is out of scope for this ticket, we could make it a separate PR.

It seems like we could have one for the Macro edits `GET` and `POST`
```scala
     path("filings" / Segment / "submissions" / IntNumber / "edits" / "macro" ) { (period, seqNr) =>
```
and one for the remaining edit types `GET`

```scala
     path("filings" / Segment / "submissions" / IntNumber / "edits" / Segment) { (period, seqNr, editType) =>
```

nickgrippin:
@schbetsy:  I definitely see the benefits of your suggestion to prevent changes to the descriptions, but I don't think we need to add in those types of safeguards.  If we were dealing with an unknown user, then we would probably want to lock down the descriptions.  But with a known user (our front-end), it seems like an unnecessary complication.

schbetsy:
Hmm. I see protecting this endpoint as **removing** unnecessary complication. @nickgrippin 

But that's a good point. As we discussed in sprint planning, we're no longer shooting for flawless, just for MVP. In that case, this LGTM, and we can add improvements later.

PR #751, opened 2016-12-21
Sort Edits By Row

schbetsy:
This completes most of #718. 

Done: 
- You can now request the Edits endpoint with the query param `sortBy=row`, and the response will be sorted by row, (almost) as shown in the documentation (see below, about the "fields" node).

Major To Dos: 
- I still need to update the "Single Edit Type" endpoint (aka, the endpoint that shows only a single category of edits--Syntactical, Validity, Quality, or Macro) to allow `sortBy=row`. I'd like to put this in a separate PR, so this one doesn't get too big.
- Each failed edit needs a "fields" node in the Json that shows the edit's relevant fields and their values for that row. This will also come in a separate PR. (This could possibly be combined with the work that adds "field" info to this endpoint when it's sorted by edit)

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -223,10 +223,13 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
     ```
 
 * `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits`
-    * `GET`  - List of all edits for a given submission, grouped by edit type
+    * `GET`  - List of all edits for a given submission
+       * By default, results are grouped by edit type, then by named edit.
+       * Use `sortBy=row` as a query parameter to group by the row in the submitted file.
 
     Example response, with HTTP code 200:
 
+    Default Sorting:
     ```json
     {
       "syntactical": {
@@ -268,17 +271,18 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
         {
             "edits": [
                 {
-                   "edit": "Q007",
-                   "description": "Description of Q007",
-                   "justifications": [
-                     {
-                       "value": "don't worry",
-                       "selected": false
-                     },
-                     {
-                       "value": "be happy",
-                       "selected": false
-                     }
+                    "edit": "Q023",
+                    "justifications": [
+                        {
+                            "id": 1,
+                            "value": "Most of the loan activity are in areas outside of an MSA/MD",
+                            "verified": true
+                        },
+                        {
+                            "id": 2,
+                            "value": "Most branches or the main branch is located outside of an MSA/MD, therefore many loans are located outside of an MSA/MD.",
+                            "verified": false
+                        },
                    ]
                 }
             ]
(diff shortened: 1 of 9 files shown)


Documents/API.md
```diff
@@ -287,6 +291,86 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
...
+              "description": "If action taken type = 1-5, 7 or 8, then lien status must = 1, 2, or 3.",
+              "fields": {
+                "Action Taken Type": 3,
+                "Lien Status": 0
+              }
+            }
+          ]
+        }
+      ],
+      "macroResults": {
```
awolfe76: Does it hurt to just keep this as `macro` instead of `macroResults`?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
```diff
@@ -46,18 +46,33 @@ trait SubmissionEditPaths
...
-              val m = validationErrorsToMacroResults(editChecks.larMacro)
-              SummaryEditResults(s, v, q, m)
-            }
+            parameters("sortBy".?) {
+              case Some("row") =>
+                val fRowSummary: Future[RowResults] = fEditChecks.map { e =>
+                  val tsErrors = e.tsSyntactical ++ e.tsValidity ++ e.tsQuality
+                  val larErrors = e.larSyntactical ++ e.larValidity ++ e.larQuality
+                  val macroErrors = e.larMacro
+                  validationErrorsToRowResults(tsErrors, larErrors, macroErrors)
```
nickgrippin: I think the creation of these three fields (`tsErrors`, `larErrors`, and `macroErrors`) could be moved inside the `validationErrorsToRowResults` function.  It would result in more concise code here, and a smaller method signature.

persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
```diff
@@ -85,6 +85,7 @@ object DemoData {
       case (instId: String, filings: Seq[Filing]) =>
         val filingActor = system.actorOf(FilingPersistence.props(instId))
         filings.foreach { filing => filingActor ? CreateFiling(filing) }
+        Thread.sleep(100)
```
nickgrippin: I'm curious about this.  Were you seeing test failures in the `InstitutionHttpApiSpec`?

schbetsy: Yeah, I kept getting failures in `SubmissionParseErrorsPathsSpec` on my branch, until I added the `Thread.sleep`. I'm as puzzled as you 😕 

I added `Thread.sleep` out of superstition, and it hasn't failed since. I can remove it again, to see whether the failures continue.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
schbetsy: That's true. My thought process was that `validationErrorsToEditResults` set a pattern for how these methods should work, and `validationErrorsToRowResults` should follow it. (though I recognize that there are other ways I could have done it)

If the goal is having the smallest method signature (which isn't necessarily the case), then both methods could just take a `HmdaFileValidationState` (the result of  `getValidationState`) as a parameter. Then the individual conversion methods could do the work of reorganizing that data into the proper format. I'd be happy to make that refactor if that would help

persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
schbetsy: Actually, I tested again locally, and I'm no longer seeing the errors. I'll update to remove that `Thread.sleep`.

Documents/API.md
schbetsy: Good question.  `macro` is a keyword in Scala, so it's hard to name things just that. I'll try to find out what it takes

schbetsy: Updated in 6fb5984.

persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
schbetsy: Updated in a514a74.

nickgrippin:
LGTM!

PR #754, opened 2016-12-21
Filing aggregation

### LINKED ISSUE #738: Project read data for filing
Projects validated LAR data and stores it in a format suitable for reading and / or aggregating

jmarin:
Closes #738

**NOTES**
* This PR does not perform aggregation yet (#758)
* There is probably some fairly straightforward refactoring to extract common code for "views", which is not included here (#757)
* This PR does not take into account resubmissions yet (#759)


query/src/main/scala/hmda/query/repository/filing/FilingComponent.scala
```diff
@@ -0,0 +1,291 @@
...
+
+import hmda.query.DbConfiguration
+import hmda.query.model.filing.LoanApplicationRegisterQuery
+import hmda.query.repository.Repository
+import slick.basic.DatabaseConfig
+import slick.jdbc.JdbcProfile
+import slick.collection.heterogeneous._
+import slick.collection.heterogeneous.syntax._
+
+trait FilingComponent { this: DbConfiguration =>
```
schbetsy: Quick question to help my understanding: I'm confused because this trait has `Filing` in the name, but it seems to just be dealing with individual LARs. (Though I do see that `period` is an added field, which means that this table does have most of the information about the filing)

Does this component contain all of the filing data we need? (i.e. Do we not need to have a DB projection for the Transmittal Sheet?)

jmarin: This PR doesn't deal with TS, but yes, naming it this way is because we will probably want to also aggregate TS data. In that case I would include the relevant classes / traits in this component

schbetsy:
My review wasn't the most in-depth, but I think this is good to merge. As we discussed in standup, I'm in favor of merging this before the holiday to avoid conflicts.

nickgrippin:
I'm also OK with having this merged sooner.  LGTM!

PR #755, opened 2016-12-21
Update ts return

kgudel:
Closes #721 

This work was completed in work that was done for #717 . It is broken out here.

PR #760, opened 2016-12-27
Treat TS like LARs in Edits Response

### LINKED ISSUE #721: Update edits response to let TS be part of the edits []

schbetsy:
Closes #721. (This replaces @kgudel's #755, since he didn't get to resolve merge conflicts and test failures before vacation)

This PR:
- Renames `lars`, `lar`, `loanId` to be `rows`, `row`, `rowId` in the JSON response
- Includes TransmittalSheet failures in the list of failures for that edit
- Updates the objects with trait `ValidationError` to have a `ts` property, so they keep track of whether the failure was for TS or a LAR.


Documents/API.md
```diff
@@ -232,58 +232,71 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
...
-                       "selected": false
-                     },
-                     {
-                       "value": "be happy",
-                       "selected": false
-                     }
-                   ]
-                }
+      "quality": {
+        "edits": []
```
nickgrippin: Is this a new change to always have an empty `edits` object?  Or is it a correction that we just haven't had in the docs?

schbetsy: Good question. I checked on master, and it has the same output as what's shown here.:
```json
  "quality": {
    "edits": []
  }
```

So evidently the docs just hadn't caught up yet.

kgudel:
@schbetsy really good cleanup! I also think that adding whether an error comes from a ts or lar on ValidationError is a really good change since trying to determine it after the fact is a bit iffy.

PR #764, opened 2017-01-03
Single Edit Type endpoint - sort by row

schbetsy:
Completes more of #718.

Now, the "Single Edit Type" endpoint (aka, the endpoint that shows only a single category of edits--Syntactical, Validity, Quality, or Macro) allows sortBy=row.

I assumed that the format of the response should be the same as the Edits endpoint when using sortBy=row. Let me know if that's not the case.

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
+++ b/persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
@@ -85,6 +85,7 @@ object DemoData {
       case (instId: String, filings: Seq[Filing]) =>
         val filingActor = system.actorOf(FilingPersistence.props(instId))
         filings.foreach { filing => filingActor ? CreateFiling(filing) }
+        Thread.sleep(100)
         filingActor ! Shutdown
     }
   }
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -399,6 +399,8 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
 
 * `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<syntactical|validity|quality|macro>`
     * `GET`  - List of edits of a specific type, for a given submission
+       * By default, results are grouped by named edit.
+       * Use `sortBy=row` as a query parameter to group by the row in the submitted file.
 
     Example response, with HTTP code 200:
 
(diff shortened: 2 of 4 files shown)


kgudel:
Do we want to enable users to get csv results if they are sorting by row?

schbetsy:
Good question. I think that the user won't need their csv formatted by row (when they already have it sorted by edit). I bet that, once the user has the edit summary in CSV format, they can use Excel (or whatever data tool they prefer) to sort and group the results however they want.

schbetsy:
Oh, you might have been asking, What if we receive a request with both parameters, "sortBy=row" and "format=csv", what should happen? 

In this PR, the sortBy=row overrides the format=csv. So if the request contains both, the response is JSON, formatted by row. @awolfe76 is that okay?

kgudel:
@schbetsy the second one is what I'm talking about! Sorry that that wasn't clear.

awolfe76:
> In this PR, the sortBy=row overrides the format=csv. So if the request contains both, the response is JSON, formatted by row. @awolfe76 is that okay?

I think that's ok for now. For what we're doing we won't make a call that has both.

But, that seems like an ok request to make, IMO. Maybe we can create an issue for it to be handled in another PR (probably a bit later)?

awolfe76:
I haven't given it much thought but it seems like it should return the csv sorted differently. So rather than the __by edit type__ (what we have now)

```csv
editType, editId, loanId
syntactical, S025, Transmittal Sheet
syntactical, S025, s1
syntactical, S025, s2
syntactical, S025, s3
syntactical, S010, s4
syntactical, S010, s5
macro, Q007
```

maybe something like this, __by `rowId`__

```csv
rowId, editType, editId
Transmittal Sheet, syntactical, S025
Transmittal Sheet, syntactical, S100
1, syntactical, S025
2, syntactical, S025
2, syntactical, S010
, quality, Q595
, macro, Q007
```

It seems like the ordering by default would be by editType > editId > rowId.
Then, with the `?sortBy=row`, rowId > editType > editId.

_Note: The `rowId` is empty for the last two rows (macro/quality). Having it empty is fine but would it better with `NA` or something similar? This would apply to any case with `format=csv`._

_Note: I think in the csv `loanId` should be `rowId` for all of these, like in #760. If easy go for it, but can be another issue. Again, this would apply to any case with `format=csv`._

There could be a few new issues here unless this is all straightforward.

schbetsy:
Okay, I guess it would be best to move this discussion to an issue.

nickgrippin:
Tested this out, LGTM

persistence/src/main/scala/hmda/persistence/demo/DemoData.scala
```diff
@@ -85,6 +85,7 @@ object DemoData {
       case (instId: String, filings: Seq[Filing]) =>
         val filingActor = system.actorOf(FilingPersistence.props(instId))
         filings.foreach { filing => filingActor ? CreateFiling(filing) }
+        Thread.sleep(100)
```
jmarin: Wondering why the `Thread.sleep`?

schbetsy: I was intermittently getting these failures:
```
These are the intermittent failures that this is meant to fix:
[info] Submission Parse Errors Path
[info] - must return no errors for an unparsed submission *** FAILED ***
[info]   404 Not Found was not equal to 200 OK (SubmissionParseErrorsPathsSpec.scala:27)
[info] - must return errors for a parsed submission *** FAILED ***
[info]   404 Not Found was not equal to 200 OK (SubmissionParseErrorsPathsSpec.scala:44)
[info] - must Return 404 for nonexistent submission *** FAILED ***
[info]   "[2017 filing period not found for institution 0]" was not equal to
"[Submission 0 not found for 2017 filing period]" (SubmissionParseErrorsPathsSpec.scala:69)
```

I traced them back to the demo Filings not loading correctly. This seems to solve the intermittent failures.

`Thread.sleep` isn't the ideal solution, but in-depth research into a better solution was out of scope for this sprint. It seems to do the trick for now.

jmarin: Sounds good to me

schbetsy:
@awolfe76 LGTY? :) If so, I think this is ready to merge.

awolfe76:
@schbetsy - the checks failing, is this still ok to merge?

schbetsy:
We never merge if Travis is failing, but we're more flexible on CodeCov. We generally just use it to help notice what code is covered by tests, not to accept/reject PRs.

PR #744, opened 2016-12-15
Close resources after use

jmarin:
Closes #737

PR #765, opened 2017-01-04
Update institutions model

### LINKED ISSUE #753: Update institutions model with panel spec
1. Update institutions model with panel spec
2. Update institution persistence code with new model
3. Update institution read projection code (i.e. `InstitutionRepository`, `InstitutionDBProjection`)

nickgrippin:
Still a WIP, main goals will be to update the rest of the demo data and generators, and fix up the rest of the compile warnings

Closes #753


nickgrippin:
Couple of possible improvements that I'm seeing:

- Both `TransmittalSheet` and `Institution` have a parent, and they're fairly similar.  These could be combined.
- Currently, running the parser won't actually do anything aside from just trying to parse the file.  Not sure if fixing this falls under the scope of updating the Institutions model.
- The `Parent` respondent ID is a `String`, while the `Respondent` respondent ID is an External ID.  This was done because there is no way to determine the External ID type for the parent, as we don't know their agency or institution type.  This could also be changed so that the `Parent` respondent ID is an External ID with an undetermined type.

validation/src/main/scala/hmda/validation/rules/ts/syntactical/S025.scala
```diff
@@ -20,10 +20,6 @@ class S025 private (institution: Institution) extends EditCheck[HasControlNumber
   def apply(input: HasControlNumber): Result = compare(input.respondentId, input.agencyCode)
 
   private def compare(filingRespId: String, filingAgencyCode: Int): Result = {
-    institution.respondentId match {
-      case Left(invalid) => Failure()
-      case Right(validRespId) =>
-        (filingRespId is equalTo(validRespId.id)) and (filingAgencyCode is equalTo(institution.agency.value))
-    }
+    (filingRespId is equalTo(institution.respondent.id.id)) and (filingAgencyCode is equalTo(institution.agency.value))
```
jmarin: This reads a little weird and looks like a typo, though it isn't. Maybe rename respondent id to externalId?

query/src/test/scala/hmda/query/model/institutions/InstitutionQueryGenerators.scala
```diff
@@ -7,20 +7,60 @@ object InstitutionQueryGenerators {
   implicit def institutionQueryGen: Gen[InstitutionQuery] = {
```
schbetsy: I couldn't find anywhere in the project that this is used. Will there be a kind of test that uses query generators?

model/shared/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -1,40 +1,38 @@
...
+  otherLenderCode: Int,
+  topHolder: TopHolder
+)
+case object Institution {
+  def empty: Institution = Institution("", UndeterminedAgency, -1, UndeterminedInstitutionType, cra = false, Set(),
+    EmailDomains("", "", ""), Respondent(ExternalId("", UndeterminedExternalId), "", "", "", ""), hmdaFilerFlag = false,
+    Parent("", -1, "", "", ""), -1, -1, TopHolder(-1, "", "", "", ""))
 }
 
+case class EmailDomains(email2015: String, email2014: String, email2013: String)
```
schbetsy: Why do we collect an institution's email domains from 2013 to 2015? In future years, will we need different years of email domain data?

query/src/test/scala/hmda/query/model/institutions/InstitutionQueryGenerators.scala
nickgrippin: I'm not sure about this one, but I assume we will be testing this in the future.

model/shared/src/main/scala/hmda/model/institution/Institution.scala
nickgrippin: So, it looks like David did this based off of Debbie's work.  @hkeeler, what will you need in terms of email domains?  I'm assuming we'll need future years of emails, so it would probably be best to change this data structure... will you only ever need the previous three years?  Do you need to know which year the email is from, or can I just return a list of all emails for an institution?

hkeeler: Yeah, I think instead of having a separate `EmailDomains` case class, we should just:

```scala
emailDomains: Set[String]
```

In most cases, those email domains would be the same, but sometimes the wouldn't, and sometimes they'd be partially null, and...its likely to get complicated if we stick to the current model.  We may even decide to go back farther in time, or look for other sources of potential email domains.  Having just a simple set of them will make this much more flexible.

model/shared/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -1,40 +1,49 @@
...
-        extIdsByType.get(extIdType) match {
-          case None => Left(RequiredExternalIdNotPresent(id, extIdType))
-          case Some(extId) => Right(ExternalId(extId.id, extIdType))
-        }
-    }
-  }
-
+  id: String,
+  agency: Agency,
+  activityYear: Int,
```
hkeeler: What is `activityYear`?  It sounds analogous to [`Filing.period`](https://github.com/cfpb/hmda-platform/blob/f68c574cca6e9ed6de18972b06a3641b57bcb18b/model/shared/src/main/scala/hmda/model/fi/Filing.scala#L25).  It doesn't _seem_ like something that belongs on `Institution` as it will change each year.  Or does this mean something else?

model/shared/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -1,40 +1,49 @@
...
+  institutionType: InstitutionType,
+  cra: Boolean,
+  externalIds: Set[ExternalId],
+  emailDomains: Set[String],
+  respondent: Respondent,
+  hmdaFilerFlag: Boolean,
```
hkeeler: Similar to `activityYear`, isn't this represented by [`Filing.filingRequired`](https://github.com/cfpb/hmda-platform/blob/f68c574cca6e9ed6de18972b06a3641b57bcb18b/model/shared/src/main/scala/hmda/model/fi/Filing.scala#L28)?

model/shared/src/main/scala/hmda/model/institution/InstitutionType.scala
```diff
@@ -7,13 +7,13 @@ import enumeratum.{ Enum, EnumEntry }
...
 sealed trait DepositoryType extends EnumEntry
 
-// FIXME: This could also be modeled as `isDepository: Option[Boolean]`.  Which is cleaner?
 object DepositoryType extends Enum[DepositoryType] {
 
   val values = findValues
 
   case object Depository extends DepositoryType
   case object NonDepository extends DepositoryType
+  case object UndeterminedDepositoryType extends DepositoryType
```
hkeeler: Do we want institution records in our system where we can't determine the `DepositoryType` and `InstitutionType`.  If you don't have them these set, you cannot derive the `respondentId` or do the related edits. We previously decided that `DepositoryType` would always present, and that's why we changed it to non-Optional (https://github.com/cfpb/hmda-platform/commit/abd7314d7e107b08fd2baa28bf461c74c1580e6d)

I think the logic in [Q033](https://github.com/cfpb/hmda-platform/blob/8341ef8b181f321fe9956ef842e92dc7158667b1/validation/src/main/scala/hmda/validation/rules/ts/quality/Q033.scala#L17) and [V110](https://github.com/cfpb/hmda-platform/blob/8341ef8b181f321fe9956ef842e92dc7158667b1/validation/src/main/scala/hmda/validation/rules/ts/validity/V110.scala#L17) assumes only known types will be set.  What do we do in these edits if `UndeterminedInstitutionType` shows up?

I guess I'm just generally concerned with all of these _undetermined_ types.  It feels like we're letting bad data in that will lead to downstream consequences.  Can we just not put these institutions in our system until we've resolved whatever upstream data sources that are causing these values to be _undetermined_?

query/src/test/scala/hmda/query/model/institutions/InstitutionQueryGenerators.scala
schbetsy: @jmarin How do you foresee using Query generators like `institutionQueryGen`?

jmarin:
@nickgrippin what is the status of this PR? Travis seems unhappy with some tests, and you have some comments from others, not sure if you are addressing them at this point in time. It would be good to push this one over the edge soon.

nickgrippin:
@jmarin: working on the tests, just an issue with the JSON reading and writing.  As far as the questions, I was going to have a quick discussion with Hans and David after standup today to address them.  I'll write up responses right now, then we can finish any discussion later today.

model/shared/src/main/scala/hmda/model/institution/InstitutionType.scala
nickgrippin: We won't have an issue deriving `respondentId`, since we'll no longer be determining that ourselves (I deleted the code [here](https://github.com/cfpb/hmda-platform/pull/765/files#diff-454df7adab09d0875ec5f470ac598c1aL23)).

The logic in the edits you mentioned doesn't assume only known types will be set, it checks to see if the type is one of several, and if not, the edit will not fail.  So, an `UndeterminedInstitutionType` will not cause any unexpected behavior in these validations.

As far as the concern for bad data, we can have a quick talk with David to discuss this further.  In short (from my understanding), it would be extremely difficult for us to resolve every single one of these undetermined types, and coding the backend to accept these seems like an easier approach without any huge consequences.

model/shared/src/main/scala/hmda/model/institution/Institution.scala
nickgrippin: We're not currently using `filingRequired` anywhere, so I don't have an issue removing it from the `Filing` class.  Or we can keep it in both places, and remove it from one of the classes once we find out how we're using it.

Basically, I'd like to keep extra data from the panel.  As we figure out more about how we're using it, we can remove duplicate fields.

model/shared/src/main/scala/hmda/model/institution/InstitutionType.scala
Kibrael: To expand on @nickgrippin's statement about data: the reason we have undetermined types is that the HMDA filer population in NIC is a very small subset (like 7,000 of 160,000 when looking at head offices). While typing each entity category is possible it would either require us to incorporate unnecessary types (e.g. edge corporation), create a new schema that encompassed all types in the NIC attributes file, or create an 'other' category.

Happy to discuss this one further. I think it's a by-product of populating the institutions model with all the NIC institutions and not just HMDA filers.

api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
```diff
@@ -84,40 +85,46 @@ trait WriteInstitutionProtocol extends InstitutionProtocol {
...
-            id.convertTo[String],
-            name.convertTo[String],
-            externalIds.convertTo[Set[ExternalId]],
-            agency.convertTo[Agency],
-            institutionType.convertTo[InstitutionType],
-            hasParent.convertTo[Boolean],
-            cra.convertTo[Boolean]
-          )
+        case Seq(id, agency, activityYear, institutionType, cra, externalIds, emailDomains, respondent, hmdaFilerFlag, parent, assets, otherLenderCode, topHolder) =>
+          Institution(id.convertTo[String], agency.convertTo[Agency], activityYear.convertTo[Int], institutionType.convertTo[InstitutionType], cra.convertTo[Boolean], externalIds.convertTo[Set[ExternalId]], emailDomains.convertTo[Set[String]], respondent.convertTo[Respondent], hmdaFilerFlag.convertTo[Boolean], parent.convertTo[Parent], assets.convertTo[Int], otherLenderCode.convertTo[Int], topHolder.convertTo[TopHolder])
```
jmarin: Could you format this line so it's a bit more readable? It's too wide

parser/shared/src/main/scala/hmda/parser/fi/panel/PanelCsvParser.scala
```diff
@@ -0,0 +1,33 @@
...
+  val file = pickFile()
+  val lines = file.getLines().toList.tail
+
+  for (line <- lines) {
+    val i = InstitutionParser(line)
+    println(i)
+  }
+
+  def pickFile(): Source = {
+    val chooser = new JFileChooser()
```
jmarin: I think this file needs to be moved to the `panelJVM` project. It contains `Java Swing` desktop components and file access methods that are `JVM` only, and while we will never run this from the `panelJS` if we attempted to do that (i.e. by mistake) it would fail.

query/src/test/scala/hmda/query/model/institutions/InstitutionQueryGenerators.scala
jmarin: One possible usage might be converting from `InstitutionQuery` to `Institution`, though I don't think we are doing that right now (or have a need to in the immediate future)

jmarin:
@nickgrippin Good job on this one, it's a beast!. I have finished my review and have a couple of minor comments, once those are addressed I think it's good to go.

parser/shared/src/main/scala/hmda/parser/fi/panel/PanelCsvParser.scala
nickgrippin: The file chooser was just a temporary thing, I already had the code for it from an earlier attempt at parsing.  I was having issues with it though, so I might switch this back to a simple command line input.

api/src/test/scala/hmda/api/protocol/admin/WriteInstitutionProtocolSpec.scala
```diff
@@ -20,19 +20,110 @@ class WriteInstitutionProtocolSpec extends PropSpec with PropertyChecks with Mus
...
+            ("state", JsString(i.topHolder.state)),
+            ("country", JsString(i.topHolder.country))
+          ))
         )
     }
   }
 
 }
+/*
+{"otherLenderCode":7,
```
schbetsy: Should this comment stay here? If so, it would be nice if it had a label/explanation.

query/src/main/scala/hmda/query/model/institutions/InstitutionQuery.scala
```diff
@@ -2,10 +2,30 @@ package hmda.query.model.institutions
...
-  cra: Boolean,
   agency: Int,
+  filingPeriod: Int,
+  activityYear: Int,
+  respondentId: String,
   institutionType: String,
-  parent: Boolean,
-  filingPeriod: Int
+  cra: Boolean,
+  emailDomain2015: String,
```
schbetsy: These `emailDomain20XX` fields should be removed, correct? Is there a test that should have caught this mismatch between the `Institution` model and the `InstitutionQuery`?

query/src/main/scala/hmda/query/repository/institutions/InstitutionComponent.scala
```diff
@@ -4,21 +4,202 @@ import hmda.query.DbConfiguration
...
     def agency = column[Int]("agency")
-    def institutionType = column[String]("type")
-    def hasParent = column[Boolean]("parent")
     def filingPeriod = column[Int]("period")
+    def activityYear = column[Int]("activity_year")
+    def respondentId = column[String]("respondent_id")
+    def institutionType = column[String]("type")
+    def cra = column[Boolean]("cra")
+
+    def emailDomain2015 = column[String]("email_2015")
```
schbetsy: Same question as above about these `emailDomain20XX` fields.

query/src/main/scala/hmda/query/model/institutions/InstitutionQuery.scala
nickgrippin: The `InstitutionConverter` converts an `Institution` to an `InstitutionQuery`, which is what is stored in the database.  I based the database columns off of the columns in the panel CSV, so that included the year labels.  I could switch the names to something more generic, like `emailDomain1`, if that would be less confusing.

schbetsy: Sure. I think removing the year labels on those fields would make them more accurate and descriptive.

schbetsy:
I agree with @jmarin. Nice work on a really complex issue. I'll make issues out of @hkeeler 's outstanding comments, so we remember to look into them. After that, assuming Travis is green, LGTM.

schbetsy:
Do you want to preserve any of your to-dos from [this comment](https://github.com/cfpb/hmda-platform/pull/765#issuecomment-271393493) as issues?

nickgrippin:
@schbetsy: One of those is already an issue (#744), and the other two (the parent and external ID) are more just nice-to-haves.  Feel free to make those issues, but I don't see us needing to do them anytime soon.

PR #771, opened 2017-01-10
Postgres view

jmarin:
Uses `PostgreSQL` database as query backend


schbetsy:
The code is looking good from the review I've done so far. I like the refactors you've included. 

What documentation should be included? I'm working now on setting it up locally and installing postgres, but it will be good to have setup instructions.

docker-compose.yml
```diff
@@ -7,6 +7,10 @@ services:
       - '9090:8080'
     volumes:
     - ./target/scala-2.12/hmda.jar:/opt/hmda.jar
+    depends_on:
+      - query_db
+    environment:
+      JDBC_URL: jdbc:postgresql://192.168.99.100:54321/hmda?user=postgres&password=postgres
```
hkeeler: This will be more portable if you reference the service name and container port.  I _think_ the following should work:

```yaml
JDBC_URL: jdbc:postgresql://query_db:5432/hmda?user=postgres&password=postgres
```

query/src/main/resources/application.conf
```diff
@@ -20,12 +20,13 @@ hmda {
...
 db {
-  driver = "slick.driver.H2Driver$"
+  driver = "slick.driver.PostgresDriver$"
 
   db {
-    url = "jdbc:h2:mem:test1;DB_CLOSE_DELAY=-1"
-    driver=org.h2.Driver
-    connectionPool = disabled
-    keepAliveConnection = true
+    # url = "jdbc:postgresql://localhost/hmda?user=postgres&password=postgres"
```
hkeeler: Do you want to remove this commented out `url`?

docker-compose.yml
jmarin: I tried the configuration you suggest and it didn't work for me, so I ended up settling on this

query/src/main/resources/application.conf
jmarin: I left it because it answers the question that @schbetsy is asking above, about how to set up the `JDBC_URL` environment variable. Kind of obfuscated and it would be better to have docs I suppose

schbetsy:
The documentation looks good. Thanks for adding it.

When I run this branch locally, I'm still having the issue where none of the demo data exists in the system. Has anyone else had any issues like that? I'm not sure if it's an issue with my setup.

jmarin:
How are you running the project? If using `sbt`, make sure you do a `clean` before `reStart` on the `api` project. With `docker`, I usually wipe out containers with `docker rm $(docker ps -a -q)` before running `docker-compose`

schbetsy:
I use these commands to run it:
```shell
$ sbt
> clean
> compile
> project api
> reStart
```

And here's the log after I `reStart`. It doesn't show any errors:
```
> reStart
[info] Application api not yet started
[info] Starting application api in the background ...
api Starting hmda.api.HmdaPlatform.main()
[success] Total time: 1 s, completed Jan 18, 2017 10:14:52 AM
> api [2017-01-18 10:14:53,162][INFO ][hmda-akka.actor.default-dispatcher-3] Slf4jLogger started
api [2017-01-18 10:14:53,180][INFO ][main] ...LOADING DEMO DATA...
api [2017-01-18 10:14:53,187][INFO ][hmda-akka.actor.default-dispatcher-5] Started validator at akka://hmda/user/supervisor/larValidation
api [2017-01-18 10:14:53,190][INFO ][hmda-akka.actor.default-dispatcher-6] Started institutions at akka://hmda/user/supervisor/institutions
api [2017-01-18 10:14:53,318][WARN ][hmda-akka.actor.default-dispatcher-5] Use `db.profile` instead of `db.driver`. The latter is deprecated since Slick 3.2 and will be removed.
api [2017-01-18 10:14:53,454][INFO ][hmda-akka.actor.default-dispatcher-5] HikariCP pool db.db is starting.
api [2017-01-18 10:14:54,235][INFO ][hmda-akka.actor.default-dispatcher-9] hmda-filing-api started on /0:0:0:0:0:0:0:0:8080
api [2017-01-18 10:14:54,237][INFO ][hmda-akka.actor.default-dispatcher-9] hmda-admin-api started on /0:0:0:0:0:0:0:0:8081
```

... but when I'm on the `master` branch, it includes these lines, which are missing when I run this branch:
```
...
api [2017-01-18 10:18:38,326][WARN ][hmda-akka.actor.default-dispatcher-2] Use `db.profile` instead of `db.driver`. The latter is deprecated since Slick 3.2 and will be removed.
api [2017-01-18 10:18:38,641][INFO ][hmda-akka.actor.default-dispatcher-10] Created: InstitutionQuery(0,Bank 0,true,3,bank,true,2017)
api [2017-01-18 10:18:38,819][INFO ][hmda-akka.actor.default-dispatcher-10] Created: InstitutionQuery(1,Bank 1,false,9,credit-union,true,2017)
api [2017-01-18 10:18:38,821][INFO ][hmda-akka.actor.default-dispatcher-10] Created: InstitutionQuery(2,Bank 2,false,1,credit-union,false,2017)
api [2017-01-18 10:18:38,823][INFO ][hmda-akka.actor.default-dispatcher-10] Created: InstitutionQuery(3,Bank 3,false,7,credit-union,true,2017)
```

jmarin:
Ok, that second output looks like successful output to me, I get the same when running on the `postgres-view` branch. What's probably missing is that you need to create a new database in your postgres server, named `hmda`, and make the `postgres` user the owner of that database (or at least grant privileges on it to read, write, create views, etc.). You can use `pgAdmin3` or if you don't like a GUI also perform those tasks from a `psql` prompt

schbetsy:
Okay, I'll try that. 

Is there a way for the code to create the necessary database automatically? If not, the instructions should go in the documentation.

And is there a way to show an error when this failure is happening? It's confusing when it fails silently.

awolfe76:
> Is there a way for the code to create the necessary database automatically?

👍  - assuming running the app locally will require it.

jmarin:
I don't think the software should create the database, this wasn't left out randomly. The project has 2 "local" modes. One with docker, and the one you are trying, without it. 

If you run things with `docker-compose`, the database is automatically created when spinning up the container, since this functionality is a feature of the `PostgreSQL` container that we are using. The `Keycloak` database operates the same way. This is configured in the `POSTGRES_DB` environment variable. 

In the "backend only" development mode, you need to provide a running `PostgreSQL` database, with the database created. I have not included instructions on how to do that because it's trivial and well documented [elsewhere](https://www.postgresql.org/docs/9.1/static/manage-ag-createdb.html).

In production, we won't require this "database management code" either. The database should be created before the application is run.

schbetsy:
That makes sense. 

In that case, I agree that we don't need instructions in the documentation, but we definitely need to document that a `hmda` database is required, with correct permissions.

schbetsy:
🎉  It works! Thanks for all of your guidance.

Now, when I sign a submission, I see the lars populating the postgres database.

query/src/main/resources/application.conf
schbetsy: Now that the env variable is documented, should the commented URL here still stay?

hkeeler:
I think we're good here.  Merging...

PR #772, opened 2017-01-10
Store all lars when validating

### LINKED ISSUE #768: Store LAR data even if it has errors
`HmdaFileValidator` should store LAR data even if it has validation errors, so a projection to the read side includes all the submitted data (in addition to this, `LARs` affected by quality errors will always be "validated with errors")

jmarin:
Closes #768


schbetsy:
I think we also need to save the Transmittal Sheet, even if it errors. (I know this isn't necessary for the publishing side, but I need it for #717 and #718, so we can show the submitted values that caused validation failures)

jmarin:
I've created a separate ticket for that, #780. I need the LAR data in place for the other PRs that I have in the queue and would prefer not to hold this one

PR #784, opened 2017-01-18
Include actual LAR/TS values in edits API

### LINKED ISSUE #717: Adding field data to edits endpoints

### LINKED ISSUE #718: Allow edits to be returned by row

schbetsy:
closes #717
closes #718


api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -62,13 +70,29 @@ trait ValidationErrorConverter {
       .getOrElse("")
   }
 
-  private def relevantFields(err: ValidationError): JsObject = {
+  private def relevantRow(err: ValidationError, vs: HmdaFileValidationState): HmdaFileRow = {
+    if (err.ts) vs.ts.get
+    else vs.lars.find(lar => lar.loan.id == err.errorId).get
+  }
```
jmarin: I think using `get` in production code is dangerous in general, since it can cause an exception if nothing is found. Can these expressions ever return something other than a `Some`?. If so, `getOrElse` should be used.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -62,13 +70,29 @@ trait ValidationErrorConverter {
...
-    val jsVals: Seq[(String, JsValue)] = fieldNames.map(n => (n, JsNumber(1)))
+    val jsVals: Seq[(String, JsValue)] = fieldNames.map { fieldName =>
+      val fieldValue = row.valueOf(fieldName)
+      (fieldName, jsonify(fieldValue))
+    }
 
     JsObject(jsVals: _*)
   }
 
+  private def jsonify(value: Any) = {
```
jmarin: I think `toJson` reads better for this method name

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
schbetsy: Good point. when the code gets to this point, the `HmdaFileValidationState` _should_ include the TS and LARs, but that doesn't mean it always will. 

I added code to keep it from failing in e1b81efb8e7a3c819aa00ea58b8d11a7e3dd98b6.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
schbetsy: Agreed. A more readable name is valuable. 

In 2e23a1de5637a61c4217e1216596406659ae88b0, I changed the name to `toJsonVal`, since it doesn't return actual JSON, just a `JsString` or `JsNumber`. Feedback/counter-offers welcome.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -62,13 +70,29 @@ trait ValidationErrorConverter {
...
   }
 
-  private def relevantFields(err: ValidationError): JsObject = {
+  private def relevantRow(err: ValidationError, vs: HmdaFileValidationState): HmdaFileRow = {
+    if (err.ts) vs.ts.getOrElse(HmdaRowError())
+    else vs.lars.find(lar => lar.loan.id == err.errorId).getOrElse(HmdaRowError())
+  }
+
+  private def relevantFields(err: ValidationError, row: HmdaFileRow): JsObject = {
     val fieldNames: Seq[String] = editDescriptions.find(e => e.editNumber == err.ruleName)
```
kgudel: While I realize that this comment refers to legacy, I figure it might be useful anyway since you are already deeply submerged in this part of the code.

We are performing the search `editDescriptions.find(e => e.editNumber == err.ruleName)`, though coded slightly differently, both here and in `findEditDescription` and using them to find the same object. It would probably be better to find that object separately and just extracted the needed pieces from it here and in  `findEditDescription`.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
```diff
@@ -50,9 +50,17 @@ trait ValidationErrorConverter {
 
   val editDescriptions = EditMetaDataLookup.values
 
-  private def rowDetail(err: ValidationError): RowEditDetail = {
+  private def editDetail(err: ValidationError, vs: HmdaFileValidationState): EditResultRow = {
+    val row = relevantRow(err, vs)
+
+    if (err.ts) EditResultRow(RowId("Transmittal Sheet"), relevantFields(err, row))
```
kgudel: Here you are passing `err, row` which expands to `err, err, vs`. Therefor I think it might be cleaner to pass either only `row` or just `err, vs`.

schbetsy: I think I disagree with this one. In this case, `row` is a `HmdaFileRow` (i.e. a TS or a LAR). It doesn't just "expand" to the types of its parameters, since the `relevantRow` method has complex logic built in.

schbetsy: Good idea. Done in 4914546f3ad5a068c32f8af19cbf9883cfef7a1c.

api/src/main/scala/hmda/api/http/ValidationErrorConverter.scala
schbetsy: You're right. This is much more streamlined. Updated in 2d0a294395e50d3d37b25536e172fbf6e9058975.

kgudel:
LGTM!

schbetsy:
@jmarin or @nickgrippin Do you want to do any more review of this?

jmarin:
LGTM

Issue #787, opened 2017-01-23
Clean up demo database upon restart


jmarin:
The application should clean up the demo database (i.e. delete `institutions`, `lar` tables if present) when starting. If this isn't done automatically, it's necessary to run `clean` from the `api` project before running it

PR #790, opened 2017-01-24
Akka & Akka HTTP version upgrade

jmarin:
Fixes vulnerability described [here](http://akka.io/news/2017/01/23/akka-http-10.0.2-security-fix-released.html)


schbetsy:
This update theoretically looks good to me.
But, when I run the project locally, I'm (once again) having the problem where the demo data doesn't get populated. Any idea why this would be happening? (I made sure to `sbt clean` and I made sure postgres is running)

jmarin:
Most likely, you still have the `institutions` table lying around in `Postgres`. Make sure you delete it before running the project again. This has been [ticketed](https://github.com/cfpb/hmda-platform/issues/787) so that we don't require that manual step, but we haven't gotten to it yet.

schbetsy:
Good call. Thanks for the tip.

LGTM.

PR #792, opened 2017-01-25
Fix parser errors in test file

schbetsy:


kgudel:
Tested this out and LGTM

schbetsy:
@kgudel. Good question. It's possible these files are stored elsewhere, like our shared drive. I'll take a look.

PR #795, opened 2017-01-27
npm publish

### LINKED ISSUE #776: Explore publishing Scala.js parser to npm

jmarin:
Closes #776

### DIFF
--- a/parser/js/README.md
+++ b/parser/js/README.md
@@ -139,7 +139,7 @@ which parses into the following structure:
     "reason3": ""
   },
   "rateSpread": "NA",
-  "hoepaStatus: 2,
+  "hoepaStatus": 2,
   "lienStatus": 1
 }
 ```
--- a/parser/js/README.md
+++ b/parser/js/README.md
@@ -2,7 +2,29 @@
 
 Standalone parser for the Home Mortgage Disclosure Act (HMDA) submission file, as per the 2017 [File Specification Guide](http://www.consumerfinance.gov/data-research/hmda/static/for-filers/2017/2017-HMDA-FIG.pdf)
 
-To run in a Node.js environment, do the following:
+## Building
+
+The standalone HMDA file parser is built with [Scala.js](https://www.scala-js.org/).
+To build this project, clone the repository from GitHub:
+
+```shell
+git clone https://github.com/cfpb/hmda-platform.git
+```
+
+[`SBT`](https://github.com/sbt/sbt) is required to build the library.
+
+```shell
+cd hmda-platform
+$ sbt
+> project parserJS
+> fullOptJS
+```
+
+This will build the production ready `JavaScript` library containing the parser, and place it in parser/js/target/scala-2.12/parserjs-opt.js
+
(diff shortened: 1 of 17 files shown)


parser/js/README.md
```diff
@@ -0,0 +1,145 @@
...
+```javascript
+var js = require('hmda-file-parser');
+var parser = new js.hmda.parser.fi.CsvParser();
+```
+Here are a couple of examples parsing HMDA data
+
+* Transmittal Sheet
+
+```javascript
+var ts = parser.parseTs('1|21|1|201503111034|2017|35-0704860|10|CENTRAL FIRST BANK|221 Hesburgh Library|Notre Dame|IN|46556|FIRST BANK|2566 S. Kinnickinnic Ave.|Milwaukee|WI|53207|Anne Shirley|574-555-2000|574-555-2001|as@centerfirst.COM');
```
schbetsy: I just want to double check: Is this all faked data? No real names, phone numbers, etc?

parser/js/README.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+Which parses into the following structure:
+
+```json
+{
```
schbetsy: The parser actually outputs a `TransmittalSheet` object, not JSON, right? If so, how should this documentation reflect that? I like the key-value format you have, since it's explicit about the names of the fields. On the other hand, if we used something like below, the user would know what kind of object to expect.

```
TransmittalSheet(1,
  1,
  201503111034,
  2017,
  35-0704860,
  10,
  Respondent(21, CENTRAL FIRST BANK, 221 Hesburgh Library, Notre Dame, IN, 46556),
  Parent(FIRST BANK, 2566 S. Kinnickinnic Ave., Milwaukee, WI, 53207),
  Contact(Anne Shirley, 574-555-2000, 574-555-2001, as@centerfirst.COM)
)
```

parser/js/README.md
jmarin: This has been taken from this [file](https://github.com/cfpb/hmda-platform/blob/master/parser/jvm/src/test/resources/txt/passes.txt) in public GitHub. I don't know the origin of that data but I assume it has been checked before committing. 
/cc @awolfe76

parser/js/README.md
jmarin: The parser outputs a `TransmittalSheet` object when used in `Scala`. In `Javascript`, it outputs native structures for that language as documented. This documentation is meant to described usage of the library as published in [npm](https://www.npmjs.com/package/hmda-file-parser) and be a guide for JavaScript developers, independently of what language was used to write the original code.

parser/js/README.md
awolfe76: I created that file based on one of the other test files in the repo, not sure which one, but if the others were good this should be too.

nickgrippin: Can we alter it a bit?  We've been caught once by the assumption that data elsewhere has already been checked.  It would be safer to have data that is obviously not realistic.

PR #797, opened 2017-02-01
Parse institution data

### LINKED ISSUE #774: Institution data loader
Extend `PanelCsvParser` and load real institution data through the event stream (`InstitutionPersistence`). Depends on #765

nickgrippin:
Closes #774 

Testing instructions:

* Make sure the postgres "institutions" table is removed with `drop table institutions;`
* In the hmda-platform project, run `sbt`, then `clean`, then select the panel project with `project panel`
* Run the panel project with `run /path/to/panelData.csv` (an example test file is on Google Drive, under `HMDA/Panel/inst_data_2016_1_trim.csv`)
* You currently have to force-stop the program when it's finished with control+C (this might get fixed later)
* Restart `sbt`, the select `project api` and `reStart`. (Don't clean!!!)
* Make sure your authorization header is updated with a few real IDs (which you can get from the log output, it's the first field argument for `InstitutionQuery`)
* Try out the endpoint `localhost:8080/institutions`, and you should get a response with real data!


kgudel:
I haven't reviewed the code but this did work for me as advertised by following the instructions!

schbetsy:
The console gets a bunch of deadletter warnings when I run the panel project. Is there a quick way to prevent those from being generated?

Here's an example:
```
12] [akka://hmda/user/StreamSupervisor-0/flow-0-2-actorRefSink] Message [scala.Some] from Actor[akka://hmda/user/supervisor/institutions#2087848179] to Actor[akka://hmda/user/StreamSupervisor-0/flow-0-2-actorRefSink#-1649578356] was not delivered. [4] dead letters encountered.
```

nickgrippin:
@schbetsy: Unfortunately, I don't think there is.  The `CreateInstitution(i)` command returns a response with the created institution, which we can't capture here with Akka streams (as far as I know).

Hopefully it shouldn't be much of a problem anyways, since this will only be run every month or so.

schbetsy:
It might be valuable to have a spec for the `PanelCsvParser`, but I don't think it should block this PR from being merged. I created an issue to remind us to look into it later: #804.

PR #798, opened 2017-02-03
Institution search

jmarin:


api/src/main/scala/hmda/api/http/public/InstitutionSearchPaths.scala
```diff
@@ -0,0 +1,60 @@
...
+      timedGet { uri =>
+        parameter('domain.as[String]) { domain =>
+          val institutionsF = for {
+            v <- institutionViewF
+            institutions <- (v ? FindInstitutionByPeriodAndDomain(domain)).mapTo[Set[Institution]]
+          } yield institutions
+          onComplete(institutionsF) {
+            case Success(institutions) =>
+              if (institutions.nonEmpty) {
+                val institutionSearch = institutions.map(i => institutiontoInstitutionSearch(i))
```
nickgrippin: This line currently doesn't do anything.

nickgrippin:
Everything works as expected.  The only parts left that I can see are tests and documentation.  Aside from that, the code LGTM.

nickgrippin:
The one other suggestion I would make would be to change one or two demo institutions to a different domain, so that we can try out other domains aside from "test.com".

nickgrippin:
LGTM!  Is this still a WIP?

persistence/src/main/resources/demoInstitutions.csv
```diff
@@ -1,4 +1,4 @@
-2015|0|3|2|1|1234|123|1234|1234|1234|test@test.com|test@test.com|test@test.com|Bank 0|VA|Test|1|1|123|123||||1|100|1|||||
-2015|1|2|2|0|4321|123|1234|1234|1234|test@test.com|test@test.com|test@test.com|Bank 1|VA|Test|1|1|123|123||||1|100|1|||||
-2015|2|3|2|1|1234|123|1234|1234|1234|test@test.com|test@test.com|test@test.com|Bank 2|VA|Test|1|1|123|123||||1|100|1|||||
-2015|4|3|2|1|1234|123|1234|1234|1234|test@test.com|test@test.com|test@test.com|Bank 3|VA|Test|1|1|123|123||||1|100|1|||||
+2015|0|3|2|1|1234|123|1234|1234|1234|test@bank0.com|test@bank0.com|test@bank0.com|Bank 0|VA|Test|1|1|123|123||||1|100|1|||||
```
hkeeler: It's probably worth having a test where a given institution has different email domains, that way you can test that all unique values are returned in the `domains` attribute, and that the same institution can be found when searching for each of its domains.

Documents/public-api.md
```diff
@@ -0,0 +1,34 @@
...
+
+   Example response, with HTTP code 200:
+
+   ```json
+   [
+      {
+        "id": "0",
+        "name": "Bank 0",
+        "domains": ["test@bank0.com"],
+        "externalIds":[
```
hkeeler: `externalIds` are a little different in the current API.  

```scala
      "externalIds": [
        { "name": "RSSD ID", "value": "3333333" }, 
        { "name": "Federal Tax ID", "value": "33-3333333" }, 
        { "name": "NCUA Charter Number", "value": "333333" }
      ]

```

As you can see, there's a human-friendly `name` attribute, no `idType`, and `id` is called `value`.  I'm cool switching to `id` if you prefer that naming, and `idType` could be useful in the future too.  However, I do think adding `name` would be really helpful.  Without this, we'll have to map `idType` to the friendly name in the Keycloak frontend code.  If we add it to the API, it will be consistent across Keycloak and any other places we show this in the future.

    To do this, looks like we'd need to add a new attribute to [`ExternalId`](https://github.com/cfpb/hmda-platform/blob/379a6f021178c59636527388904630f7fc8b2f0f/model/shared/src/main/scala/hmda/model/institution/ExternalId.scala).

    I think the names could pretty closely follow the names of the case classes:

    ```scala
    case object FdicCertNo extends ExternalIdType("fdic-certificate-number", "FDIC Certificate Number")
    case object FederalTaxId extends ExternalIdType("federal-tax-id", "Federal Tax ID")
    case object NcuaCharterId extends ExternalIdType("ncua-charter-id", "NCUA Charter Number")
    case object OccCharterId extends ExternalIdType("occ-charter-id", "OCC Charter Number")
    case object RssdId extends ExternalIdType("rssd-id", "RSSD ID")
    case object UndeterminedExternalId extends ExternalIdType("undetermined-external-id", "???")
    ```

    I guess we also need to figure out what to do with `UndeterminedExternalId`?  That doesn't seem like something we'd want the end-user to see.  I'm not quite sure what an `UndeterminedExternalId` would even be.  Don't we always know where a given ID comes from?

jmarin: I think this discussion has come up before, and yes, there are cases where the id is undetermined for institutions if I remember correctly. @nickgrippin might know, if not this is a question for @Kibrael

Documents/public-api.md
```diff
@@ -0,0 +1,34 @@
...
+
+This documenatation describes de public HMDA Platform HTTP API
+
+## Institutions
+
+### Search
+
+* `/institutions?domain=<domain>`
+
+   * `GET` - Returns a list of institutions filtered by their email domain. If none are found, an HTTP 404 error code (not found) is returned
```
awolfe76: I don't have a real preference. Can we provide a message in the case of no results?

hkeeler:
I spun off the `UndeterminedExternalId` question off into its own issue.  No reason to jam up this PR over that.

Also, I'm close to having auth-proxy serving the institution search over HTTPS.  If we're OK with that approach, we can check that off the list here as well.

Documents/public-api.md
jmarin: @awolfe76 do you mean in the API response? We can certainly do that. If that is the case, can you specify what it should look like?

jmarin:
@hkeeler SSL termination should happen in a proxy in front of this API. We could add HTTPS support, it's not that hard, but I have concerns about that architecture (and security, too).

Documents/public-api.md
awolfe76: Exactly. So instead of the normal 404 message `Not Found` maybe we could make it a bit more clear. Maybe `Domain Not Found` would do?

The only reason being is that I'm assuming if someone misspells `institutions`, eg `insttttitutions`, we are returning the `Not Found`.

Can we do that? Have a different message for different types of 404s?

If that's too much for now we can always create a new issue later.

hkeeler:
>@hkeeler SSL termination should happen in a proxy in front of this API. We could add HTTPS support, it's not that hard, but I have concerns about that architecture (and security, too).

Agreed, let's not do SSL in hmda-platform.  I have an institution-search vhost setup in auth-proxy that's proxying to the new API, terminating SSL, and handling CORS.  I think that should suffice for local dev for now.  I should have a PR up for that shortly.  It'll require some updates to `docker-compose.yml` as well, so there'll be a PR to this repo too.

Documents/public-api.md
```diff
@@ -0,0 +1,36 @@
...
+   {
+     "institutions":
+     [
+        {
+          "id": "0",
+          "name": "Bank 0",
+          "domains": ["test@bank0.com"],
+          "externalIds":[
+            {
+              "name": "1234",
```
hkeeler: `name` and `value` are switched.  I _think_ this happened when the parameters were renamed on https://github.com/cfpb/hmda-platform/pull/798/files#diff-1f02427f497c8439ae1bb2fbf7070c65R8.  I'm guessing when the data gets read in, it is assuming that order for the params?

jmarin: Reversed in the implementation, documentation, or both?

hkeeler: Confirmed this is fixed now. 👍

PR #799, opened 2017-02-07
Increase timeout for probe.expectMsg tests

jmarin:
Increases default timeout for `probe.expectMsg` tests. Might help with random failures that happen sometimes in CI server under load


nickgrippin:
LGTM

PR #800, opened 2017-02-07
Endpoint for Verifying Quality Edits

### LINKED ISSUE #747: Verify quality edits

schbetsy:
Closes #747

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -534,7 +534,7 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
 
     ```json
     {
-      "verified": true,
+      "verified": true
     }
     ```
 
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -316,6 +316,7 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
         ]
       },
       "quality": {
+        "verified": false,
         "edits": []
       },
       "macro": {
@@ -527,7 +528,29 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
     }
     ```
 
-    * `POST` - Provides verification for macro edits
+    * `POST` - Provides verification for Quality edits
+
+    Example payload, in `JSON` format:
+
+    ```json
+    {
+      "verified": true,
+    }
+    ```
+
+    Example response, with HTTP code 200:
+
+    ```json
+    {
+      "verified": true,
+      "status": {
+        "code": 8,
+        "message": "validated with errors"
+      }
+    }
+    ```
+
+    * `POST` - Provides justification for Macro edits
 
     Example payload, in `JSON` format:
 
--- a/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
+++ b/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
@@ -92,27 +92,28 @@ trait SubmissionEditPaths
             }
           }
         }
-      } ~ timedPost { uri =>
-        if (editType == "macro") {
-          entity(as[MacroEditJustificationWithName]) { justifyEdit =>
-            completeVerified(institutionId, period, seqNr, uri) {
-              val supervisor = system.actorSelection("/user/supervisor")
-              val submissionId = SubmissionId(institutionId, period, seqNr)
-              val fHmdaFileValidator = (supervisor ? FindProcessingActor(HmdaFileValidator.name, submissionId)).mapTo[ActorRef]
-              val fValidationState = for {
-                a <- fHmdaFileValidator
-                j <- (a ? JustifyMacroEdit(justifyEdit.edit, justifyEdit.justification)).mapTo[MacroEditJustified]
-                state <- getValidationState(institutionId, period, seqNr)
-              } yield state
-              completeValidationState(editType, fValidationState, uri, "")
-            }
-          }
-        } else {
-          completeWithMethodNotAllowed(uri)
-        }
       }
     }
 
+  // institutions/<institutionId>/filings/<period>/submissions/<seqNr>/edits/macro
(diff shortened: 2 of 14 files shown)


schbetsy:
Known: There currently isn't logic to prevent users from filing if they haven't verified Quality edits (same for justifying Macro edits, I think).

Also, I think I mis-spoke in my demo today: `verified` always starts as `false`, even if the file didn't have any Quality edits.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
```diff
@@ -92,38 +94,68 @@ trait SubmissionEditPaths
...
+  def justifyMacroEditPath(institutionId: String)(implicit ec: ExecutionContext) =
+    path("filings" / Segment / "submissions" / IntNumber / "edits" / "macro") { (period, seqNr) =>
+      timedPost { uri =>
+        entity(as[MacroEditJustificationWithName]) { justifyEdit =>
+          completeVerified(institutionId, period, seqNr, uri) {
+            val fValidator = fHmdaFileValidator(SubmissionId(institutionId, period, seqNr))
+
+            val fValidationState = for {
+              v <- fValidator
+              j <- (v ? JustifyMacroEdit(justifyEdit.edit, justifyEdit.justification)).mapTo[MacroEditJustified]
```
nickgrippin: You probably don't need to convert this to anything if the variable isn't used later

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
```diff
@@ -92,38 +94,68 @@ trait SubmissionEditPaths
...
+
+            val fSubmissions = for {
+              va <- fValidator
+              v <- va ? VerifyQualityEdits(verified)
+              sa <- fSubmissionsActor
+              s <- (sa ? GetSubmissionById(subId)).mapTo[Submission]
+            } yield s
+
+            onComplete(fSubmissions) {
+              case Success(submission) =>
```
nickgrippin: You may want to check that the `VerifyQualityEdits` call returned successfully here.  Maybe return both `v` and `s` on the yield statement, and make sure they've both completed.

We did something similar [here](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionBasePaths.scala#L53)

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
```diff
@@ -211,27 +217,34 @@ class SubmissionEditPathsSpec extends InstitutionHttpApiSpec with LarGenerators
   }
 
   "Edit Type endpoint: return 405 when posting justification to syntactical endpoint" in {
-    postWithCfpbHeaders("/institutions/0/filings/2017/submissions/0/edits/syntactical") ~> institutionsRoutes ~> check {
+    postWithCfpbHeaders("/institutions/0/filings/2017/submissions/0/edits/syntactical") ~> Route.seal(institutionsRoutes) ~> check {
```
nickgrippin: Why this change?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
schbetsy: That makes sense. I'll make that change.

I didn't add this code originally, it just shows up in the diff because I changed the indentation level, but I'm happy to update it.

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
schbetsy: Since I changed how `SubmissionEditPaths` responds to the "Justify Macro Edits" endpoint, we no longer have to manually return a "Method Not Allowed" error (that's why I could remove the `completeWithMethodNotAllowed` method in `HmdaCustomDirectives`). Now, akka-http handles that rejection automatically, but only when the route is "sealed".

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
schbetsy: I updated the `for` comprehension so it maps the return message from `VerifyQualityEdits` to its expected type. In my tests, this catches the case where the call fails or returns the wrong type.

(I could be wrong--dealing with futures is complex--but) I don't think it's possible for this to fail silently. If line 132 never returns, the `for` comprehension just won't yield anything.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionEditPaths.scala
schbetsy: I changed my mind, based on the example below. Mapping the value this way ensures that the response was of the expected type. The code will throw an error if something is wrong, instead of failing silently.

Documents/API.md
```diff
@@ -527,7 +528,29 @@ All endpoints in the `/institutions` namespace require two headers (see "Authori
...
     ```
 
-    * `POST` - Provides verification for macro edits
+    * `POST` - Provides verification for Quality edits
+
+    Example payload, in `JSON` format:
+
+    ```json
+    {
+      "verified": true,
```
nickgrippin: This shouldn't have a comma

nickgrippin:
Aside from that one tiny last change, this LGTM

Documents/API.md
schbetsy: Good call. Done in 51edaf4.

wpears:
LGTM

PR #801, opened 2017-02-07
Clean up tables on load

### LINKED ISSUE #787: Clean up demo database upon restart

nickgrippin:
Closes #787


api/src/main/scala/hmda/api/HmdaPlatform.scala
```diff
@@ -39,8 +43,25 @@ object HmdaPlatform {
 
   }
 
+  private def cleanup(system: ActorSystem, supervisor: ActorRef, querySupervisor: ActorRef)(implicit ec: ExecutionContext): Unit = {
```
jmarin: Do you still need `supervisor` / `querySupervisor` with this approach?

jmarin:
Just a question about the `cleanup` function signature, otherwise LGTM

schbetsy:
I tried running the Panel CSV Parser and got this error:
```
[error] (run-main-0) org.postgresql.util.PSQLException: ERROR: relation "institutions" already exists
org.postgresql.util.PSQLException: ERROR: relation "institutions" already exists
```

Would it be out of scope for this PR to solve that failure too?

schbetsy:
How does this PR interact with the Panel CSV Parser PR? I tried loading panel data from a file then starting the project (like in the instructions in #797), but I just saw Demo data instead.

nickgrippin:
@schbetsy: make sure you have the `HMDA_IS_DEMO` environment variable set to `false`.  Let me know if that works for you.

schbetsy:
The environment variable did the trick. Thank you.

I think the only remaining question is [this one](https://github.com/cfpb/hmda-platform/pull/801#issuecomment-278774276), about the PSQL Exception in the Panel CSV Parser.

PR #816, opened 2017-02-15
Prevent one intermittent test failure

schbetsy:
Use `Await.result` to ensure submission status finishes updating before testing the status.


nickgrippin:
LGTM

PR #791, opened 2017-01-24
Modified lar

jmarin:
Closes #775

PR #844, opened 2017-02-24
Irs endpoint

### LINKED ISSUE #767: IRS report
Currently, the `/irs` endpoint is using mocked data. Now that it is a report of the only the data in the submission, and has no other dependencies, we can start to make this endpoint with real data.

The response format is the same as the mocked format in the [docs](https://github.com/cfpb/hmda-platform/blob/7f47895d0cecb5499736121da79c9b6d84d68e31/Documents/API.md), with the `status` removed.

Because there are no dependencies we'll have to update the [submission status](https://github.com/cfpb/hmda-platform/blob/7f47895d0cecb5499736121da79c9b6d84d68e31/Documents/submission-status.md) too. We shouldn't need the `10: IRS generated` status, we can remove it. The front-end will have to poll for the `/irs` endpoint to check if it's ready. We'll only start polling if the status is > `7: validating` (and if on the summary page). The endpoint should return an empty `msas: []` until the report is ready.

So the IRS report can start to be generated once status is > `7: validating`.

nickgrippin:
I've implemented an approach similar to Juan's streaming in #791.  However, I'm still unsure about [this method to read from the database](https://github.com/cfpb/hmda-platform/compare/master...nickgrippin:irs-endpoint?expand=1#diff-2aed2acf1e258d6e8fa56eb77a243b30R359).  Any feedback would be greatly appreciated 😄 

Closes #767


api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
```diff
@@ -28,16 +34,14 @@ trait SubmissionIrsPaths
...
-        val irsJson = "{\n  \"msas\": [\n    {\n      \"id\": \"123\",\n      \"name\": \"MSA 123\",\n      \"totalLARS\": 4,\n      \"totalAmount\": 123,\n      \"conv\": 4,\n      \"FHA\": 0,\n      \"VA\": 0,\n      \"FSA\": 0,\n      \"1to4Family\": 4,\n      \"MFD\": 0,\n      \"multiFamily\": 0,\n      \"homePurchase\": 0,\n      \"homeImprovement\": 0,\n      \"refinance\": 4\n    },\n    {\n      \"id\": \"456\",\n      \"name\": \"MSA 456\",\n      \"totalLARS\": 5,\n      \"totalAmount\": 456,\n      \"conv\": 5,\n      \"FHA\": 0,\n      \"VA\": 0,\n      \"FSA\": 0,\n      \"1to4Family\": 5,\n      \"MFD\": 0,\n      \"multiFamily\": 0,\n      \"homePurchase\": 0,\n      \"homeImprovement\": 0,\n      \"refinance\": 5\n    }\n  ],\n   \"status\": {\n       \"code\": 10,\n       \"message\": \"IRS report generated\"\n     }}"
+        val data = larTotalRepository.getMsaSource().map(x => ChunkStreamPart(x.toString))
 
-        val response = HttpResponse(StatusCodes.OK, entity = HttpEntity(ContentTypes.`application/json`, irsJson))
+        val response = HttpResponse(StatusCodes.OK, entity = HttpEntity.Chunked(ContentTypes.`application/json`, data))
```
jmarin: Is this working for you? When I run it I don't get any response back

nickgrippin: No, I don't get any response back either.  Not sure why it drops the connection so suddenly, I don't see any errors in the logs.

jmarin: I think the `lars_total` view is not being created (happened to me when I ran it, not sure if that behavior is consistent)

query/src/test/scala/hmda/query/repository/filing/FilingComponentSpec.scala
```diff
@@ -125,6 +123,24 @@ class FilingComponentSpec extends AsyncWordSpec with MustMatchers with FilingCom
 
     }
 
+    "Stream IRS" in {
```
schbetsy: Could we get another test that shows that the table is still correct when multiple LAR files get uploaded? (e.g. by multiple institutions or another from the same institution)

nickgrippin: That should be handled on the `UploadPaths`, which deletes the tables when a new file is uploaded.

See [here](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala#L64)

schbetsy: Could we get a test that proves that it works?

nickgrippin: Sure thing, but that should be in a separate ticket, since that's behavior that's not affected by this pull request

query/src/main/scala/hmda/query/repository/filing/FilingComponent.scala
```diff
@@ -296,28 +299,79 @@ trait FilingComponent { this: DbConfiguration =>
...
+    private def getTableStream()(implicit ec: ExecutionContext): DatabasePublisher[Msa] = {
+      val disableAutocommit = SimpleDBIO(_.connection.setAutoCommit(false))
+      val query = table.drop(0)
+      val action = query.result.withStatementParameters(fetchSize = queryFetchSize)
+
+      db.stream(disableAutocommit andThen action)
+    }
+
+    def getMsaSeq()(implicit ec: ExecutionContext): Future[Seq[Msa]] = {
+      Source.fromPublisher(getTableStream()).grouped(1000).runWith(Sink.head)
```
jmarin: It might make sense to put this value into the configuration file and make it a parameter

schbetsy:
It doesn't seem to be working correctly. Currently, it returns the same response regardless of which institution/period/submissionId I use in the request parameters.
It seems to be collecting totals for all of the LARs in **all** submitted files, instead of totaling each one separately.
Hitting `localhost:8080/institutions/0/filings/2017/submissions/1/irs` should just give me totals for submission `0-2017-1`, but instead it gives me totals for all submissions.

query/src/main/scala/hmda/query/repository/filing/FilingComponent.scala
```diff
@@ -296,28 +299,79 @@ trait FilingComponent { this: DbConfiguration =>
...
   }
 
-  class LarTotalTable(tag: Tag) extends Table[LoanApplicationRegisterTotal](tag, "lars_total") {
-    def id = column[Int]("id", O.PrimaryKey)
-    def respondentId = column[String]("respondent_id")
-    def period = column[String]("period")
-    def total = column[Int]("total")
-
-    override def * = (id, respondentId, period, total) <> (LoanApplicationRegisterTotal.tupled, LoanApplicationRegisterTotal.unapply)
+  class LarTotalTable(tag: Tag) extends Table[Msa](tag, "lars_total") {
```
jmarin: Not a major thing, but the naming is a bit off now imho, "lars_total" as a table name is mapped to `Msa`. I think the table name should reflect the fact that it mostly contains information by `msa`. 
Maybe `lars_total_msa`?

nickgrippin:
@schbetsy: I can't replicate that behavior on my end.  The lars table should be dropped when a new file is uploaded.  I tested this by uploading a file with 10 LARs to submission 1, then the same file to submission 2.  Hitting the IRS endpoint for submission 2 showed a total of 10 lines, as expected.

What steps did you follow to test this out?  Perhaps I can try retracing your steps and see where the problem is.

schbetsy:
Sure. Here's what I did:

1. Drop the tables in the psql console (to ensure a clean slate)
2. In SBT, `clean` , `compile`, `project api`, and `reStart`
3. Upload `passes.txt` to Institution 0. (submission id: `0-2017-3`)
4. Upload `FirstTestBankData_clean_407_2017.txt` to Institution 1. (submission id: `1-2017-1`)
5. Hit the IRS endpoint for each of those submissions.

I expect the IRS report for `passes.txt` (`/institutions/0/filings/2017/submissions/3/irs`) to show a total of 10 LARs, since that file contains 10 LARs total. Instead it contains thousands of lars.

I expect the IRS report for `FirstTestBankData_clean_407_2017.txt` (`/institutions/1/filings/2017/submissions/1/irs`) to have a different response from the other submission's IRS, but they are the same.

schbetsy:
A high-level question: This version of the IRS endpoint is accessed with the respondentID from the submitted HMDA file, not necessarily the Institution ID in the platform. 

For example, if we're using institution `0` and the test file `clean_164-lars.txt`, that means the request needs to be `/institutions/88-6600810/filings/2017/submissions/1/irs` since `88-6600810` is the respondent id from the file. 

Is this what we want? If so, the front-end will need to know the respondentID that was used in the file, so they can request the correct endpoint. 

Once we're using realistic data, how likely is this scenario, where the institutionID and the respondentID are different?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
```diff
@@ -28,18 +29,18 @@ trait SubmissionIrsPaths
   implicit val timeout: Timeout
 
   // institutions/<institutionId>/filings/<period>/submissions/<submissionId>/irs
-  // NOTE:  This is currently a mocked, static endpoint
   def submissionIrsPath(institutionId: String)(implicit ec: ExecutionContext) =
     path("filings" / Segment / "submissions" / IntNumber / "irs") { (period, submissionId) =>
```
schbetsy: I noticed that the `submissionId` parameter is ignored (which makes sense... the endpoint can only return data for the latest submission, because of how the underlying table works). Should we consider removing that parameter from the endpoint and shortening the path? It could become `institutions/<id>/filings/<period>/irs`.

One argument for NOT removing that parameter: if we kept it, you could use the `RequestVerificationUtils` to evaluate requests and return a 404 when the institution/period/submission is nonexistent. Currently it returns a 500 Internal Server error for any institution or period not in the `lars_total_msa` table.

nickgrippin: Just pushed an update to use the `RequestVerificationUtils`

schbetsy:
Ah, I think I found a bug with these steps:
1. Start the project with a clean slate (deleting tables in pg, cleaning and restarting the API)
2. Modify `clean_164-lars.txt` so its respondentID is 0, then upload it (submission id 1-2017-1).
3. Note the contents of the `lars_total_msa` table (so we can compare against it in the next step)
4. Upload `clean_10-lars.txt` (which also has respondentID 0) (submission id 1-2017-2).

I would expect the table to have only the entries for the latest upload. Instead, it has totals for both of the submissions.

jmarin:
@nickgrippin It seems that some tests are still failing

nickgrippin:
@schbetsy: In response to those steps that you followed, we delete the previous submission based on the institution ID in the URL, not in the database (see [here](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala#L65)).  The institution ID is based of the id_rssd field in panel (see [here](https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/main/scala/hmda/parser/fi/InstitutionParser.scala#L18)), which currently matches the respondent ID sometimes, but not perfectly.  It's the best solution we have right now, but the whole institution ID obviously needs more work.

nickgrippin:
@jmarin: The tests for the database seem to have sporadic failures due to the tests not waiting for the data to be inserted.  I've tried to fix some of these errors, but they still aren't behaving properly.  Are there any changes you would suggest?

schbetsy:
Sounds related to #861. My strategy in fixing that one will probably be to use `Await` to ensure the data are loaded before the test starts, which requires some refactoring.

schbetsy:
The IRS either needs to always use the respondent ID or always use the Institution ID. As long as it's inconsistent, this feature won't work in a way that we can count on.

nickgrippin:
@schbetsy: I can understand the confusion, but it's the best solution we have at the moment.  We can't retrieve LARs from the database by institution ID, as it's not something that we currently store.  And we don't know the respondent ID of an institution's submitted LARs when making the request from the API.  

The [modified LAR endpoint](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/public/PublicLarHttpApi.scala#L26) uses the exact same approach that I'm using here.  

Currently, we have to expect test files to have the same respondent ID as institution ID (see my changes to the `clean-10-lars.txt` file).  Sometime in the future, we will have to clear up the concept of an institution ID.

jmarin:
Merging. Issues found here will be ticketed and worked on as separate issues

PR #851, opened 2017-03-02
WIP: Q011

### LINKED ISSUE #224: Validation (2017): Q011 (Total Number of Loan Applications)
**Edit #:** Q011
**Type:** Macro
**Column:** Total Number of Loan Applications

If the current or previous year's total number of applications is ≥ 500, then the current year should be within (+ or -) 20% of the previous year's total.

Additional resources:

jmarin:
Closes #224

This is one possible approach. Feedback welcome


schbetsy:
@jmarin I've looked at the approach from a high level, and I think it makes sense, assuming we can have the previous year's data into the `lars` table. Do we have a plan for getting 2016 data and loading it?

I'll be adding more specific low-level comments on the code soon.

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -45,31 +48,35 @@ class InstitutionPersistence extends HmdaPersistentActor {
 
   override def receiveCommand: Receive = {
     case CreateInstitution(i) =>
-      if (!state.institutionIds.contains(i.id)) {
+      if (!state.institutions.contains(i)) {
```
jmarin: Good question. I don't think it should be created if it has the same id. I'm going to go back to checking creation by making sure the id is not already present

persistence/src/main/scala/hmda/persistence/institutions/InstitutionPersistence.scala
```diff
@@ -45,31 +48,35 @@ class InstitutionPersistence extends HmdaPersistentActor {
...
         sender() ! None
         log.warning(s"Institution does not exist. Could not update $i")
       }
 
     case GetState =>
-      sender() ! state.institutionIds
+      sender() ! state.institutions
+
+    case GetInstitution(id) =>
+      val i = state.institutions.find(x => x.id == id)
```
jmarin: find returns `Option`, so in that case you will get a `None`

build.sbt
```diff
@@ -135,6 +136,7 @@ lazy val persistence = (project in file("persistence"))
 lazy val query = (project in file("query"))
   .settings(hmdaBuildSettings:_*)
   .settings(
+    parallelExecution in Test := false,
```
jmarin: Yeah. This isn't related to the implementation of edits per se, but I noticed that we have been running tests sequentially by default, so I've [changed](https://github.com/cfpb/hmda-platform/pull/851/files#diff-c6575342216356c040d3b53a22f877c4R21) that. There are some exceptions though, since some tests expect to be run in a certain sequence (they usually are the ones where we're dealing with a database of some sort). This change should make your tests run much faster overall (though not in this particular project).

PR #857, opened 2017-03-03
Cleanup sample files

### LINKED ISSUE #809: Reorganize sample HMDA files
Currently, our collection of HMDA files (which lives in `parser/jvm/src/test/resources/txt/`) has several issues:

- Some files have a `|` character at the end of each line, making them fail in the parser when they shouldn't.
- They don't adhere to a naming scheme, so it's hard to tell which file to use for which purpose.
- Some files _shouldn't_ fail any edits (i.e. Debbie found that they passed all edits in previous years), but they do fail some edits in our system. This either indicates an issue with the file or with our code, and we need to figure out which is the case.

kgudel:
Closes #809 

This PR organizes the test files into folders that describe how or if they fail, the structure is as follows:

- parseErrors
- edits
  - syntax_validity
  - quality_macro
- clean

It also removes the ending pipes from a number of files.

I have also looked at the errors in the QualityMacroPasses files and the errors look valid.


schbetsy:
This is already really helpful! A few small comments/requests:

1. Could we standardize the filenames so they're no longer misleading and/or inconsistent? One idea: maybe the filename for each could include the `institutionId` and the number of lines in the file. For example: `QualityMacroPasses_Test1.txt` (which has syntax and validity failures, so the current name is misleading) could become `Springfield_8899000007_129-lars.txt`

2. I have a slight preference for using shallower nesting of directories. So I'd like to see the directories be `clean`, `quality_macro_edits`, `syntax_validity_edits`, `parse_errors` at the top level. (But I'm open to other opinions)

awolfe76:
I second the shallower nesting, or maybe even none at all. It's easy to see that we could _**(and should)**_ have files that have edits of one, a few, and all types. So, I think a good file naming scheme is more important than the directory structure.

`failsSyntactical.txt`
`failsValidity.txt`
`failstSyntacticalValidity.txt`
`fails[whateverAndThisToo].txt`

At some point I think we might need to get really specific and have a file that only fails a specific edit, so:

`fails[thisEdit].txt`

We could also add @schbetsy idea and add other information to the file names as needed.

schbetsy:
Sure, I think descriptive names like those would work. Maybe something like these?:

- `ParserErrors_127-lars.txt`
- `ValidityMacroEdits_338-lars.txt`
- `SyntacticalOnly_40-lars.txt`

Eventually it will be important to know what HMDA year a file is written for (once we start working on 2018 edits), but those new files could just live in a separate folder.

One concern is the amount of maintenance that will require. Since we're still adding edits to the system and might fix bugs in existing edits, we'll need someone (everyone?) to be responsible for keeping the naming and organization up-to-date.

nickgrippin:
Looks great!

PR #860, opened 2017-03-06
Replace sample.get with sample.getOrElse

### LINKED ISSUE #825: Replace sample.get with sample.getOrElse in tests
Some tests are using `sample.get` from `ScalaCheck` generators, which could potentially return a `None` (in most cases this is not the intended behavior). Replace these with `sample.getOrElse` and provide a default value instead of `None`. 

Example of usage [here](https://github.com/cfpb/hmda-platform/blob/master/api/src/test/scala/hmda/api/http/public/InstitutionSearchPathSpec.scala#L32)

nickgrippin:
Closes #825

### DIFF
--- a/query/src/test/scala/hmda/query/repository/filing/FilingComponentSpec.scala
+++ b/query/src/test/scala/hmda/query/repository/filing/FilingComponentSpec.scala
@@ -48,8 +48,8 @@ class FilingComponentSpec extends AsyncWordSpec with MustMatchers with FilingCom
 
   "LAR Repository" must {
     "insert new records" in {
-      val lar1 = larGen.sample.getOrElse(LoanApplicationRegister()).copy(respondentId = "resp1")
-      val lar2 = larGen.sample.getOrElse(LoanApplicationRegister()).copy(respondentId = "resp1")
+      val lar1 = getOneLar.copy(respondentId = "resp1")
+      val lar2 = getOneLar.copy(respondentId = "resp1")
       repository.insertOrUpdate(lar1).map(x => x mustBe 1)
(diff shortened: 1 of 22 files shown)


jmarin:
Travis is reporting a compilation failure

query/src/test/scala/hmda/query/repository/filing/FilingComponentSpec.scala
```diff
@@ -99,10 +99,10 @@ class FilingComponentSpec extends AsyncWordSpec with MustMatchers with FilingCom
     "Stream rows for a specific respondent id" in {
       val respId = "resp2"
       val p = ""
-      val lar1 = larGen.sample.get.copy(respondentId = respId)
-      val lar2 = larGen.sample.get.copy(respondentId = respId)
-      val lar3 = larGen.sample.get.copy(respondentId = respId)
-      val lar4 = larGen.sample.get.copy(respondentId = "resp3")
+      val lar1 = larGen.sample.getOrElse(LoanApplicationRegister()).copy(respondentId = respId)
```
schbetsy: Code like this is repeated lots of times: `larGen.sample.getOrElse(LoanApplicationRegister())`, `InstitutionGenerators.institutionGen.sample.getOrElse(Institution.empty)`, etc.

Could it work to abstract that into a method? Like this: `LarGenerators.getOne` and `InstitutionGenerators.getOne`

schbetsy:
Looking again, I think a more intuitive name for those methods would be `sampleLar` and `sampleInstitution`. 

Otherwise, this looks good to me. 👍

jmarin:
LGTM

PR #863, opened 2017-03-07
Persistence backpressure

### LINKED ISSUE #701: Add backpressure to persistent actors

jmarin:
Closes #701 

For an explanation of the approach implemented here, [this](http://doc.akka.io/docs/akka/2.4/scala/stream/stream-integrations.html) might be useful

**Note**: backpressure has been added to the file upload, raw messages and parsed and validated `TS` and `LAR`. Messages that contain validation errors are still being sent in a _"fire and forget"_ fashion and are not backpressured. Changing this calls for a refactoring that is quite invasive at this point in time and one that will most likely also affected by the clustering work. I've created this [ticket](https://github.com/cfpb/hmda-platform/issues/862) to track the remainder of this work


persistence/src/main/scala/hmda/persistence/processing/ProcessingMessages.scala
```diff
@@ -25,5 +25,6 @@ object ProcessingMessages {
   case class ParsingCompletedWithErrors(submissionId: SubmissionId) extends Event
   case class ValidationCompleted(submissionId: SubmissionId) extends Event
   case class ValidationCompletedWithErrors(submissionId: SubmissionId) extends Event
+  case class Persisted() extends Event
```
schbetsy: I'm not clear on what empty parens do. Could `Persisted` work, rather than `Persisted()`? If there's a benefit to one or the other, what is it?

persistence/src/test/scala/hmda/persistence/processing/HmdaFileParserSpec.scala
```diff
@@ -50,12 +50,14 @@ class HmdaFileParserSpec extends ActorSpec with BeforeAndAfterEach with HmdaFile
 
     "persist parsed LARs" in {
       parseLars(hmdaFileParser, probe, lines)
+      (1 to lines.length - 1).foreach(_ => probe.expectMsg(Persisted()))
```
schbetsy: Could this line live inside of the `parseLars` method?

api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala
```diff
@@ -81,7 +84,9 @@ trait UploadPaths extends InstitutionProtocol with ApiErrorProtocol with Submiss
         val uploadedF = byteSource
           .via(splitLines)
           .map(_.utf8String)
-          .runForeach(line => processingActor ! AddLine(uploadTimestamp, line))
+          //.runForeach(line => processingActor ! AddLine(uploadTimestamp, line))
+          .mapAsync(parallelism = flowParallelism)(line => (processingActor ? AddLine(uploadTimestamp, line)).mapTo[Persisted])
```
nickgrippin: Would it be possible to use `mapAsyncUnordered` here, since we don't care about the order of the lars?

persistence/src/main/resources/application.conf
```diff
@@ -16,4 +16,5 @@ akka.persistence.snapshot-store.local.dir = "target/snapshots"
 hmda {
   actor-lookup-timeout = 5
   persistent-actor-timeout = 60
+  actor-flow-parallelism = 4
```
nickgrippin: How was this number picked?

persistence/src/main/scala/hmda/persistence/processing/ProcessingMessages.scala
jmarin: Case classes need to have parenthesis by default, this didn't use to be the case but with current `Scala` versions it is mandatory. If you try to do something like this

```scala
case class Person
```

You will get a compiler error as follows:

```shell
error: case classes without a parameter list are not allowed;
use either case objects or case classes with an explicit `()' as a parameter list.
```
Case objects (singletons), on the other hand, do not need parenthesis. I modeled this as a case class initially so that some additional data could be passed back to the sender (i.e. the time when something was persisted). Since I ended up not using that, I'm going to revert this to a `case object`

api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala
jmarin: I think we **do** care though. If we are ever asked to **exactly** reproduce the file that was sent to us (this has come up in the past), having things in order is going to be mandatory. In addition to this, `mapAsyncUnordered` is a nice optimization, but we don't know at this point if / how much it really helps

persistence/src/main/resources/application.conf
jmarin: From the Akka Documentation:

`There is a performance advantage of using parallelism > 1 even though the actor will only process one message at a time because then there is already a message in the mailbox when the actor has completed previous message.`

Most examples I've seen use values from 2 to 5. What's important is that it's greater than 1. In any case I've made this configurable should we ever go down to this level of tuning

api/src/main/scala/hmda/api/http/institutions/UploadPaths.scala
nickgrippin: That makes sense, I wasn't sure if order was going to be needed.

persistence/src/main/scala/hmda/persistence/processing/ProcessingMessages.scala
schbetsy: Thank you for this explanation; that makes a lot of sense. 👍

PR #865, opened 2017-03-07
add data file from panel process

awolfe76:
This adds the data file, containing real data and "dummy" data, from the panel process.

See [GHE]/HMDA-Operations/hmda-panel-dev/pull/172


nickgrippin:
The file's format is good and it parses correctly.  Out of the 213 lines in this file, 182 were pushed to the backend database, which means there are 31 lines with duplicate IDs (see [here](https://github.com/cfpb/hmda-platform/blob/74307f93df04097248b46325a937eedc04ee167a/inst_data_2015_20170306_dummy.csv#L30-L33), the ID in this case is `3025406`).  This isn't necessarily a blocker, you'll just get a bunch of duplication errors when you run panel.

The only other comment is the location of the file.  Maybe have this located in the panel project, instead of in the main directory?

awolfe76:
> The only other comment is the location of the file. Maybe have this located in the panel project, instead of in the main directory?

Thought maybe this would come up. That's ok with me.

awolfe76:
@nickgrippin any specific path in panel?

nickgrippin:
Maybe `hmda-platform/panel/src/main/resources`?

PR #866, opened 2017-03-09
Default placeholder

### LINKED ISSUE #823: Replace unused matched variables with placeholder
The codebase is full of matching structures that hold a variable which is not used in the scope in which they are defined. For instance, in many places expressions as the following:

```scala
case SomeObject(somevalue) =>
...
```
can be replaced by the following if we don't care about the value of `somevalue`

```scala
case SomeObject(_) =>
...
```

This change is to apply consistency across the codebase, and avoid annoying editing warnings in some IDEs (i.e. IntelliJ IDEA, `Declaration is never used`)

jmarin:
Closes #823 
Replace unused variable names with placeholders; fix minor stylistic issues


schbetsy:
I wonder if we could use an automated linter for catching style fixes like this.

PR #868, opened 2017-03-09
Drop database before running panel

### LINKED ISSUE #807: PSQL Exception in Panel Parser
Running the Panel CSV Parser gets this error, unless I manually drop the `institutions` table in `psql`, which is tedious:
```
[error] (run-main-0) org.postgresql.util.PSQLException: ERROR: relation "institutions" already exists
org.postgresql.util.PSQLException: ERROR: relation "institutions" already exists
```

Update the Panel CSV Parser so it doesn't throw that error. We'll have to decide whether that means it should drop the `institutions` table first, or if it should allow adding additional institutions to the ones that are currently in the database.

nickgrippin:
Closes #807

### DIFF
--- a/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
+++ b/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
@@ -45,11 +45,15 @@ object PanelCsvParser extends InstitutionComponent with DbConfiguration {
 
     val timeout = 5.seconds
 
+    println("Cleaning DB...")
+    Await.result(repository.dropSchema(), timeout)
+    println("Creating new schema...")
     Await.result(repository.createSchema(), timeout)
     val institutionPersistence = Await.result(institutionPersistenceF, timeout)
 
     val source = FileIO.fromPath(new File(args(0)).toPath)
 
+    println("Reading file...")
     source
       .via(Framing.delimiter(ByteString("\n"), maximumFrameLength = 1024, allowTruncation = true))
       .drop(1)


wpears:
https://xkcd.com/327/

panel/src/main/scala/hmda/panel/PanelCsvParser.scala
```diff
@@ -45,11 +45,15 @@ object PanelCsvParser extends InstitutionComponent with DbConfiguration {
 
     val timeout = 5.seconds
 
+    println("Cleaning DB...")
+    Await.result(repository.dropSchema(), timeout)
```
jmarin: Can we do the same without using `Await.result`, which is a blocking operation?. Not terribly concerned with this since this is its own application that will just run once

schbetsy:
I had to go back to the original Panel PR for testing instructions, so I thought I'd copy them here for anyone who wants to test:
* No need to drop tables in postgres.
* In the hmda-platform project, run: `sbt`, `clean`, `project panel`.
* Run the panel project with `run /path/to/panelData.csv` (an example test file is on Google Drive, under `HMDA/Panel/inst_data_2016_1_trim.csv`)
* You currently have to force-stop the program when it's finished with control+C (this might get fixed later)
* In the same terminal, run `export HMDA_IS_DEMO=false`
* Restart `sbt`, the select `project api` and `reStart`. (Don't use `clean`.)
* Make sure your authorization header is updated with a few real IDs (which you can get from the log output, it's the first field argument for `InstitutionQuery`)
* Try out the endpoint `localhost:8080/institutions`, and you should get a response with real data!

jmarin:
I'm going to approve this. Hopefully we can get rid of blocking operations on futures soon

Issue #849, opened 2017-02-24
Update states in the application


jmarin:

schbetsy:
Could it make sense to combine Syntactical + Validity, since they have the same effect for the user? In my example below, I use "blocking edits" as the term for combined S+V edits, but we could negotiate on the name. 

I think we could also combine Quality + Macro, since they also have the same action required. I've called them "verifiable edits", but we could maybe come up with a better name.


- 1: created - The filing period is now open and available to accept HMDA data.
- 2: uploading - The data is currently being uploaded to the HMDA Platform.
- 3: uploaded - The data has finished uploading and is ready to be analyzed.
- 4: parsing - The uploaded file is being analyzed to ensure that it follows the requirements specified in the Filing Instructions Guide.
- 5: parsed with errors - The uploaded file is not formatted according to the requirements specified in the Filing Instructions Guide. The filing process may not proceed until the file has been corrected and reuploaded.
- 6: parsed - The data conforms to the requirements specified in the Filing Instructions Guide and is ready to proceed with the filing process.
- 7: validating - The data is currently being validated.
- 8: validated with blocking edits - The data validation process is complete but there are syntactical and/or validity edits that need to be addressed. The filing process may not proceed until the file has been corrected and reuploaded.
- 9: validated with verifiable edits - The data validation process is complete. However, there are quality and/or macro edits that need to be verified before proceeding with the filing process.
- 10: validated - The data validation process is complete and is ready to be submitted.
- 11: signed - The financial institution has certified that the data is correct. This completes the HMDA filing process for this year.
- -1: Failed - An error occurred during the process of submitting the data. Please select the Submit button again.

jmarin:
We decided not to work on this at this stage. Closing

PR #869, opened 2017-03-09
Fix SubmissionManagerSpec

### LINKED ISSUE #810: Fix SubmissionManagerSpec
As a result of [this PR](https://github.com/cfpb/hmda-platform/pull/752), the test no longer works.

schbetsy:
Closes #810


jmarin:
The problem that I see is that while this fixes the test, this change renders the `Failed` state useless, we will never reach that unless I am missing something. While we are thinking about [changing](https://github.com/cfpb/hmda-platform/issues/849) the states we need to keep a `Failed` state in the application

schbetsy:
Yeah, that's kind of what I was trying to get at with questions 1 and 2 in my original comment. 

Under what circumstances _should_ the FSM receive `None`?

schbetsy:
Marking as WIP so I can investigate more thoroughly.

nickgrippin:
LGTM!

PR #872, opened 2017-03-09
Update demo data

### LINKED ISSUE #817: Update demo data
We have 4 demo institutions. Can we update the demo data to match a few use cases?

For example:
- bank 0 - no filing, no submissions
- bank 1 - has a filing that's not started, with 0 submissions
- bank 2 - has a filing that's in progress, with n submissions in various states
- bank 3 - has a filing that's completed, with n submissions in various states but one that is completed

And add an institution:
- bank 4 - has a filing that's canceled, with n submissions in various states

kgudel:
Closes #817


awolfe76:
From the docs this looks good.
@nickgrippin, can you take a look too? You know these formats better than me.

persistence/src/main/resources/demoSubmissions.csv
```diff
@@ -1,4 +1,6 @@
-0|2017|1|created|1484842271000|0
-0|2017|2|created|1484842271000|0
-0|2017|3|created|1484842271000|0
-1|2017|1|created|1484842271000|0
+2|2017|1|parsed-with-errors|1484842271000|0
```
nickgrippin: Can we have a submission here with a status of created?  When testing any endpoints that require a file to be uploaded, it would be easier to have a submission ready to go, rather than having to create a new one every time we re-run the API.

PR #874, opened 2017-03-09
WIP: Database access as singleton

### LINKED ISSUE #873: Access relational database from singleton object
There are several classes extending this [trait](https://github.com/cfpb/hmda-platform/blob/master/query/src/main/scala/hmda/query/DbConfiguration.scala#L6). While the `config` object is lazy that doesn't mean that there is a single instance of the database running.

jmarin:
Closes #873


query/src/test/scala/hmda/query/projections/filing/HmdaFilingDBProjectionSpec.scala
```diff
@@ -1,22 +1,21 @@
...
 
-class HmdaFilingDBProjectionSpec extends ActorSpec with DbConfiguration with BeforeAndAfterEach with BeforeAndAfterAll with LarGenerators {
+class HmdaFilingDBProjectionSpec extends ActorSpec with BeforeAndAfterAll with LarGenerators {
 
   implicit val timeout = 5.seconds
 
   override def afterAll(): Unit = {
     super.afterAll()
-    larRepository.config.db.close()
+    //larRepository.config.db.close()
```
nickgrippin: Why did you comment out this line?

jmarin: I don't think it's needed. Tests run locally with these changes. Waiting for `TravisCI` to confirm

PR #879, opened 2017-03-13
Access relational DB from singleton

### LINKED ISSUE #873: Access relational database from singleton object

jmarin:
Closes #873


nickgrippin:
LGTM

PR #880, opened 2017-03-13
WIP: Add formatted names to external ID

### LINKED ISSUE #806: Add human-readable names to ExternalIdType
As mentioned on https://github.com/cfpb/hmda-platform/pull/798#discussion_r99916755, it would be convenient and add consistency across webapps if we added a human readable name for each of the `ExternalIdType`s.  This _could_ be represented something like the following in JSON.

```json
{ "typeCode": "fdic-certificate-number", "typeName": "FDIC Certificate ID", "value": "1234567"}
```

We could also use this as an opportunity to formalize how each ID is represented to the user.  Some of the IDs, like "Federal Tax ID" [have many variants](https://forums.digitalpoint.com/threads/is-an-ein-the-same-as-tin.682962/#post-17483185):
* Federal Tax ID Number
* Employee Identification Number (EIN)
* Federal Employee Identification Number (FEIN)

There is no _immediate_ need for this.  I've added a [mapping on the Keycloak side](https://github.com/cfpb/hmda-platform-auth/pull/75/files#diff-0d19d4c8d9acc4f8c775f35627a24d52R59) to take care of this for now.  However, I suspect we'll add something similar to the filing app at some point, so we could reevaluate this idea then.

nickgrippin:
For testing, the only response that includes external ID is the institution search path on the public API.

Closes #806

### DIFF
--- a/api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
+++ b/api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
@@ -51,7 +51,7 @@ trait WriteInstitutionProtocol extends InstitutionProtocol {
     override def write(obj: ExternalIdType): JsValue =
       JsObject(
         "entryName" -> JsString(obj.entryName),
-        "formattedName" -> JsString(obj.name)
+        "formattedName" -> JsString(obj.formattedName)
       )
 
     override def read(json: JsValue): ExternalIdType = json match {
--- a/api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
+++ b/api/src/main/scala/hmda/api/protocol/admin/WriteInstitutionProtocol.scala
@@ -50,17 +50,17 @@ trait WriteInstitutionProtocol extends InstitutionProtocol {
 
     override def write(obj: ExternalIdType): JsValue =
       JsObject(
-        "entryName" -> JsString(obj.entryName),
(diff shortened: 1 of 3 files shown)


schbetsy:
The code here LGTM. 

@hkeeler, how does this look from an Auth perspective?

hkeeler:
Oh yeah, also, let's not merge this until we have the corresponding change in the -auth repo ready to go.  If we merge now, it'll break the registration process in Keycloak.

jmarin:
Added `WIP` so this doesn't get merged until auth work is done

nickgrippin:
> {
>   "type": {
>     "code": "rssd-id",
>     "name": "RSSD ID"
>   },
>   "value": "1234567"
> }

"type" is a keyword in Scala and would be a little frustrating to deal with (I'd have to escape every reference of it in the code).  Is there another word we can use?  I can switch to "code" and "name" pretty easily inside that object.

hkeeler:
Right.  Forgot about that.  We could go more explicit with `externalIdType`?

hkeeler:
Looks good.  Will merge once https://github.com/cfpb/hmda-platform-auth/issues/97 is ready.

hkeeler:
The corresponding change for the registration page is now up at https://github.com/cfpb/hmda-platform-auth/pull/100.  Once that has the green light, this is good-to-go for mergification.

hkeeler:
Pulled down the latest from all the projects, and confirmed all is happy.

PR #883, opened 2017-03-14
Update to Institution ID

### LINKED ISSUE #877: Figure out Respondent ID/Institution ID
Our current implementation of Institution ID uses the [`id_rssd` field](https://github.com/cfpb/hmda-platform/blob/master/parser/shared/src/main/scala/hmda/parser/fi/InstitutionParser.scala#L18) from the panel file.

However, when storing LARs in the PSQL database, we only store the exact LAR itself, absent any information about who submitted it.  This means that when retrieving the LAR for either [the modified LAR endpoint](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/public/PublicLarHttpApi.scala#L26) or [the IRS endpoin](https://github.com/cfpb/hmda-platform/blob/master/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala#L38)t, we filter by respondent_id, since we have no other information about the Institution with which to look up/filter the database.

In reality, id_rssd and respondent_id will rarely match.  We need to address this by providing a way for us to look up LARs by `id_rssd`.

nickgrippin:
Closes #877


nickgrippin:
WIP, just checking to see what fails.  Feel free to comment!

nickgrippin:
One big concern with this PR: I un-commented the Submission IRS Path test, as it seemed to be actually passing.  Please check this out locally and run it a couple times to make sure it's not going to cause trouble.  If it is, I can just add the comments back in.

schbetsy:
Hmm. I did a smoke test where I:
- uploaded a 90-line file to Institution 2
- uploaded a 10-line file to Institution 2
- checked IRS and found that institution 2 had 100 total lars. 😕 

I haven't read the code, so I can't pinpoint exactly where it is, but it seems there's still a disconnect where the db doesn't handle old LARs properly when a new file is uploaded.

We definitely need a spec that will test for correct behavior when multiple files are uploaded. Repeatedly smoke testing is a super inefficient way to check.

nickgrippin:
> We definitely need a spec that will test for correct behavior when multiple files are uploaded. Repeatedly smoke testing is a super inefficient way to check this.

I've added a [test](https://github.com/cfpb/hmda-platform/pull/883/files#diff-5cc7d7f6a0cfdfc525e6d12fce11b2e6R80) to the `UploadPathsSpec` that currently fails when you try and re-upload a file.  Working on a fix right now.

schbetsy:
Commenting to document the errors we're seeing in specs that query the lars table.
```
java.util.concurrent.RejectedExecutionException: Task slick.basic.BasicBackend$DatabaseDef$$anon$2@1c56ba24 rejected from slick.util.AsyncExecutor$$anon$2$$anon$1@5826a92b[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 28]
```

@jmarin When we move to Cassandra, will the tests be on Cassandra too? Do you think this error is Postgres-specific, or are we likely to keep seeing it after we migrate the new read storage?

jmarin:
This error is specific to the H2 database. The initial work on `Cassandra` is not going to address the read side for now, so I can't offer a solution from that angle. 

From the looks of it, it sounds like two tests are concurrently trying to write to the H2 engine (which unfortunately, in embedded mode, is single thread).

api/src/main/resources/application.conf
```diff
@@ -4,6 +4,7 @@ akka {
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
   http.parsing.max-content-length = 1G
   http.server.default-host-header = "cfpb.gov"
+  test.timefactor=30.0
```
jmarin: Does it need to be this high? This means that `expectMsg` assertions will take this value times `akka.test.single-expect-default` (3 seconds by default)

query/src/main/scala/hmda/query/view/filing/HmdaFilingView.scala
```diff
@@ -54,7 +54,7 @@ class HmdaFilingView(period: String) extends HmdaPersistentActor {
         saveSnapshot(state)
       }
       event match {
-        case LarValidated(lar) =>
+        case LarValidated(lar, institutionId) =>
```
jmarin: Can probably do `Larvalidated(lar, _)`

PR #886, opened 2017-03-16
Created panel documentation

### LINKED ISSUE #805: Documentation for PanelCsvParser
There were instructions in #797 for how to run and use the `PanelCsvParser`. Let's capture those in documentation so people will know how to use it.

Also note that the `HMDA_IS_DEMO` environment variable needs to be set. (via https://github.com/cfpb/hmda-platform/pull/801#issuecomment-278775104)

nickgrippin:
Closes #805

### DIFF
--- a/Documents/panel.md
+++ b/Documents/panel.md
@@ -1,0 +1,32 @@
+# Panel CSV Parser
+
+## Overview
+The panel parser is designed to read a CSV file and load the data onto the HMDA-Platform.  The CSV file should use the `|` (pipe) delimiter, and should include a header row as the first line.
+
+## Running the parser
+An example panel file is located at `hmda-platform/panel/src/main/resources/inst_data_2015_20170308_dummy.csv`
+
+While in the `hmda-platform` project, start up `sbt`, then execute the following commands:
+```shell
+> project panel
+> clean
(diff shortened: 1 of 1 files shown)


schbetsy:
Should this document include specifics about the format of the file? For instance, users might want to know
- what data each column should contain
- whether any fields will cause failures if they're blank or formatted incorrectly (i.e. what if a field contains a string where there should be an integer?)
- whether it assumes there's a header row in the file

nickgrippin:
> Should this document include specifics about the format of the file?

That's a great question, and I could easily add that.  The only question is, who would that be useful for?  David is the only one who generates the panel files for us, and we have a spreadsheet in the shared drive to keep track of the expected field types and locations.

schbetsy:
Good point. I guess the only other potential audience is people who aren't on our team. Maybe an FI wants to clone and understand our code? <-- I'm not sure how likely that is. Maybe we can ask @feomike or @Kibrael

schbetsy:
Documenting what we discussed in standup today: Mike confirmed that we are the audience for this documentation. Since David has documented his panel generation process in other venues, we don't need to have a complete file specification here.

Issue #845, opened 2017-02-24
Pagination for edits


awolfe76:
_Pulls the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<type>/<edit>` endpoint out of #719. Details coming._

awolfe76:
To be able to add pagination to edits we need to paginate on the specific edits, eg S020, instead of the edit types. This will require a new endpoint for each edit, `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<type>/<edit>`, and will cause updates to the [`/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits`](https://github.com/cfpb/hmda-platform/blob/23a1dff0463e84db9c55972bb73ec2dfac9b67dc/Documents/API.md#edits) endpoint as well.

_**Note**: the UI is currently NOT using the [`/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<type>`](https://github.com/cfpb/hmda-platform/blob/23a1dff0463e84db9c55972bb73ec2dfac9b67dc/Documents/API.md#edits-by-type) endpoints, but changes would be required there as well._

#### Update the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits`

Update this endpoint to return a list of edits. For example:

```json
{
  "syntactical": {
    "edits": [
      {
        "edit": "S020",
        "description": "Agency code must = 1, 2, 3, 5, 7, 9. The agency that submits the data must be the same as the reported agency code."
      },
      {
        "edit": "S010",
        "description": "The first record identifier in the file must = 1 (TS). The second and all subsequent record identifiers must = 2 (LAR)."
      }
    ]
  },
  "validity": {
    "edits": []
  },
  "quality": {
    "verified": false,
    "edits": []
  },
  "macro": {
    "edits": []
  }
}
```

This would allow someone to get the full list of edits and rows and then, in the case of sorting by edits, make API calls for the details of each specific edit.

#### Create the `/institutions/<institution>/filings/<period>/submissions/<submissionId>/edits/<type>/<edit>` endpoint

The would return the details for an edit and include the pagination information. For example:

```json
{
  "edit": "S020",
  "rows": [
    {
      "row": {
        "rowId": "Transmittal Sheet"
      },
      "fields": {
        "Agency Code": 10
      }
    },
    {
      "row": {
        "rowId": "8299422144"
      },
      "fields": {
        "Agency Code": 10
      }
    }
  ],
[...]

PR #889, opened 2017-03-17
Verify Macro Edits (and Remove Justification of Macros)

### LINKED ISSUE #846: Modify the approval process for macro quality edits
Instead of having the user provide a justification for macro quality edits, the process will now be the same as with quality macro edits. The user will confirm through a check box that the data are correct.

schbetsy:
Closes #846


persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -59,37 +57,23 @@ object HmdaFileValidator {
...
-        HmdaFileValidationState(ts, lars, tsSyntactical, tsValidity, tsQuality, larSyntactical, larValidity, larQuality, v, larMacro)
-      case MacroEditJustified(e, j) =>
-        val elem = larMacro.find(x => x.ruleName == e)
-        elem match {
-          case Some(v) =>
-            val macroUpdated: Seq[ValidationError] = MacroValidationError.updateJustifications(larMacro, j, v)
-            HmdaFileValidationState(ts, lars, tsSyntactical, tsValidity, tsQuality, larSyntactical, larValidity, larQuality, qualityVerified, macroUpdated)
-          case None => this
-        }
+      case tsValidated @ TsValidated(newTs) => this.copy(ts = Some(newTs))
```
nickgrippin: Good cleanup here! 👍

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -224,17 +208,13 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor
...
-        sender() ! QualityEditsVerified(v)
-      }
-
-    case JustifyMacroEdit(error, j) =>
-      persist(MacroEditJustified(error, j)) { e =>
-        updateState(e)
-        sender() ! MacroEditJustified(error, j)
-      }
+    case VerifyEdits(editType, v) =>
+      if (editType == Quality || editType == Macro) {
```
nickgrippin: I definitely understand the check for `editType`, but I think this `if` statement might just be clutter.  From my understanding, it's impossible for this statement to fail.

api/src/main/scala/hmda/api/model/EditResults.scala
```diff
@@ -20,15 +19,15 @@ trait EditResultsCollection {
...
 
-case class MacroResult(edit: String, justifications: Set[MacroEditJustification])
-case class MacroResults(edits: Seq[MacroResult]) {
+case class MacroResult(edit: String)
+case class MacroResults(verified: Boolean, edits: Seq[MacroResult]) {
   def toCsv = edits.map(e => "macro, " + e.edit + "\n").mkString
 }
 case object MacroResults {
-  def empty: MacroResults = MacroResults(Nil)
+  def empty: MacroResults = MacroResults(true, Nil)
```
nickgrippin: I don't think this is used anywhere...  Did you default the empty result set to true because there wouldn't be any edits to verify?  Is that how the endpoint would work?

Also, if this is kept, could it be changed to `verified = true`?

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
schbetsy: You're right that the current code will never send an edit type other than Quality or Macro. But I think safeguards like this are valuable to future-proof the code for future development. This way, the purpose of the code is clear, and it enforces the business logic, so no developers will misuse this method in the future.

api/src/main/scala/hmda/api/model/EditResults.scala
schbetsy: Good call! That object is indeed unused, so I deleted it in 599078734ec9170adfc5cb1d3cbc5c3361fa9aa2.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
schbetsy: If we like that cleanup, we can do the same in a few other actors, too.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
nickgrippin: Totally valid reasoning 👍 I'm ok with leaving this as is

schbetsy:
@jmarin Could you review/merge this PR? It's causing conflict with #896 and #845.

PR #903, opened 2017-03-30
update demo data and a few test files for demo institutions

### LINKED ISSUE #904: Update readme to show new changes to bank data
We list the available institutions in the readme, within the ["to run the entire platform"](https://github.com/cfpb/hmda-platform#to-run-the-entire-platform) section. We should update that with the new institution data.

awolfe76:
Can you check that I'm right that the LAR data should be using `id_rssd` in the 2nd column?

Also, if you run it locally you should be able to register and start filing for both `bank0.com` and `bank1.com` but after the upload (use `bank0.txt` and `bank1.txt` to test), you'll get stuck on the syntactical and validity page. It looks like this is because of the macro updates so it's expected. So it doesn't appear to be related to the data.

Also closes #904


parser/jvm/src/test/resources/txt/bank0.txt
```diff
@@ -0,0 +1,201 @@
+1|Bank0_RID|7|201803071050|2017|99-1234567|200|bank-0 National Association|1650 Biscayne Blvd |Miami|FL|33132     ||||||Jane Smith|111-111-1111|222-222-2222|jane_smith@bank0.com
+2|Bank0_RID|7|8299422144|20170220|1|1|1|2|78|3|1|20170226|16974|17|063|0001.03|2|5|5| | | | |8| | | | |2|5|173|6| | | |NA   |2|1
```
nickgrippin: Until [this PR](https://github.com/cfpb/hmda-platform/pull/883) is merged in, you'll want to respondent ID of the LAR to match the `id_rssd` of the bank you're testing.  In this case, `6999998`.

Currently, this will cause a 500 error on the IRS endpoint and a blank response on the modified LAR endpoint.

The same thing will need to be done for `bank1.txt`.  Or wait until the other PR is merged in, after which any LAR file will work.

awolfe76:
@nickgrippin - updated the test files.

schbetsy:
Tell me about the files in the test folder called `bank0` and `bank1`. Are they just copies of another test file, but with the respondentID updated to match the institutions?  I'm worried because their names don't match the name scheme we have for that directory of test files, so their uses aren't clear.

awolfe76:
> Tell me about the files you added in the test folder called bank0 and bank1. Are they just copies of another test file, but with the respondentID updated to match the institutions? I'm worried because their names don't match the name scheme we have for that directory of test files, so their uses aren't clear.

That's exactly it. I _think_ later we can remove them completely but for now, to see the whole process, you need to have an account with bank0 or bank1 and use the corresponding file. But yes, they are copies of another file with the ID changed to make the whole work flow work.

PR #921, opened 2017-04-06
Q011

### LINKED ISSUE #224: Validation (2017): Q011 (Total Number of Loan Applications)

### LINKED ISSUE #900: Include Institution in ValidationContext
The Hmda File Validator uses the `ValidationContext` to provide extra information for validating the HMDA file. It's set up to send Institution info to individual edits, but that institution isn't actually being supplied. This means that all edits that require info about the institution aren't being run ([S025](https://github.com/cfpb/hmda-platform/issues/94) is an example). 

[This line](https://github.com/cfpb/hmda-platform/blob/7783ce7ceb6b9154c2124d79a18de3b870b66213/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala#L122) is where the `ValidationContext` is set up. Currently it's getting `None` as its institution value, instead of the actual institution submitting the file.

jmarin:
Closes #224
Closes #900


validation/src/main/scala/hmda/validation/rules/lar/macro/Q011.scala
```diff
@@ -0,0 +1,62 @@
...
+    val currentLarCount: Future[Int] = count(lars)
+
+    val validationStats = system.actorSelection("/user/validation-stats")
+
+    val lastYearCount = (validationStats ? FindTotalLars(institution.id, lastYear.toString)).mapTo[Int]
+
+    for {
+      c <- currentLarCount
+      l <- lastYearCount
+      lower = 1 * (1 - multiplier)
```
schbetsy: I think it would be less confusing to calculate `upperValue` and `lowerValue` in the `for` comprehension, rather than splitting the logic between here and lines 55-56.

validation/src/main/scala/hmda/validation/rules/lar/macro/Q011.scala
```diff
@@ -0,0 +1,62 @@
...
+      upper = 1 * (1 + multiplier)
+    } yield {
+      when(c is greaterThanOrEqual(larSize) or (l is greaterThanOrEqual(larSize))) {
+        val lowerValue = lower * l
+        val upperValue = upper * l
+        c.toDouble.toString is numericallyBetween(lowerValue.toString, upperValue.toString)
```
schbetsy: I think you can use the `between` predicate rather than `numericallyBetween`, so then you won't have to convert the values to strings.

validation/src/main/scala/hmda/validation/ValidationStats.scala
```diff
@@ -0,0 +1,61 @@
...
+        updateState(e)
+      }
+    case FindTotalLars(id, period) =>
+      val submissionStats = state.stats
+      val filtered = submissionStats.filter(s => s.id.institutionId == id && s.id.period == period)
+      if (filtered.nonEmpty) {
+        val submission = filtered.sortWith(_.id.sequenceNumber > _.id.sequenceNumber).head
+        sender() ! submission.totalLars
+      } else {
+        sender() ! 0
```
nickgrippin: Do we want to default to 0 here?  If an Institution doesn't have a submission from the previous year, it will cause Q011 to fail no matter how many LARs are in their current submission.  (This might also be a question for @lesgou)

For example, all of the current `clean_X-lars.txt` files will trigger Q011 when running the API in demo mode.

validation/src/main/scala/hmda/validation/rules/lar/macro/Q011.scala
schbetsy: `lowerBound = lastYear * (1 - multiplier)`
`upperBound = lastYear * (1 + multiplier)`

nickgrippin:
This seems to make the validation process noticeably slow down... I'm having issues (locally) where the validation process doesn't complete before the actors start timing out and shutting down.  This tends to happen in files with a LAR count of 5000 or greater.

validation/src/main/scala/hmda/validation/rules/IfInstitutionPresentInAggregate.scala
```diff
@@ -0,0 +1,13 @@
+package hmda.validation.rules
+
+import hmda.model.institution.Institution
+import hmda.validation.context.ValidationContext
+
+object IfInstitutionPresentInAggregate {
```
schbetsy: This name is misleading, since it's actually checking for both the institution and the year are present. Maybe we could call it `IfContextPresentInAggregate`?

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -82,18 +89,35 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor
...
+  def ctx: ValidationContext = ValidationContext(institution, Try(Some(submissionId.period.toInt)).getOrElse(None))
+
   var state = HmdaFileValidationState()
 
   val supervisor = system.actorSelection("/user/supervisor")
   val fHmdaFiling = (supervisor ? FindHmdaFiling(submissionId.period)).mapTo[ActorRef]
 
+  val submissionLarStats = context.actorOf(SubmissionLarStats.props(submissionId))
+
+  override def preStart(): Unit = {
```
schbetsy: We also need to use `ctx` in the `BeginValidation` method. Currently is uses the one defined on line 140, which doesn't supply the institution.

validation/src/main/scala/hmda/validation/ValidationStats.scala
jmarin: @lesgou I'm going to leave this at _confusing_ for now. The edit is complicated enough to also require a lookup in past year's panel (which we don't have yet) to determine if it's a new filer or now

lesgou: :+1:

schbetsy:
In smoke testing this branch, it seems that the validation process never finishes. The submission stays in "Validating" status indefinitely. From my rudimentary debugging, it looks like the `for`-comprehension in `Q011` never yields.

validation/src/main/scala/hmda/validation/SubmissionLarStats.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  case class SubmissionLarStatsState(totalLars: Int = 0) {
+    def updated(event: Event): SubmissionLarStatsState = event match {
+      case SubmissionStatsUpdated(total) =>
+        SubmissionLarStatsState(total)
+    }
+  }
+}
+
+class SubmissionLarStats(submissionId: SubmissionId) extends HmdaPersistentActor {
```
schbetsy: I'm having trouble understanding this actor. The `HmdaFileValidationState` already holds every lar that was validated for the submission, so it seems like this actor's job could just be done in the `HmdaFileValidator`. 

Is this to prevent overloading that actor with too many responsibilities? Or other purpose I'm missing?

validation/src/main/scala/hmda/validation/ValidationStats.scala
```diff
@@ -0,0 +1,65 @@
...
+      case SubmissionStatsAdded(s) =>
+        ValidationStatsState(stats :+ s)
+    }
+  }
+}
+
+class ValidationStats extends HmdaPersistentActor {
+  import ValidationStats._
+
+  override def preStart(): Unit = {
```
schbetsy: Do you want to do `super.preStart()` here? (It's something I've seen in other actors)

schbetsy:
It now seems to work perfectly in smoke tests now, so well done fixing that bug 👍 . Just a few more code comments.

validation/src/main/scala/hmda/validation/SubmissionLarStats.scala
jmarin: Yes, that is exactly the case. `HmdaFileValidator` is already a beast, we should probably break it down into smaller pieces (not part of this PR though).

validation/src/main/scala/hmda/validation/ValidationStats.scala
jmarin: No, I left that out on purpose. The bug we were seeing was related to `super.preStart()`, where the `HmdaPersistentActor` uses a `receive timeout` to shut itself down. This 'ValidationStats' actor is a top level actor that is supposed to be running all the time, so there is no need to shut it down after a period of inactivity

validation/src/main/scala/hmda/validation/SubmissionLarStats.scala
schbetsy: Okay, that makes sense--since other macro edits will need various stats too, I see how this addition makes that simpler. Thanks.

jmarin:
@nickgrippin @schbetsy I was finally able to fix the original PR. This should include everything necessary. As you can see there is a change in `HmdaFileValidatorSpec` in one of the tests, the changes in this PR trigger `S025` for those LARs. Previously, `S025` was not being triggered because no institution was being passed in `ValidationContext` when running the `HmdaFileValidator`. This is changed now. This could be "fixed" with changes to demo data and more code in the test, but I believe it complicates things.

PR #923, opened 2017-04-07
Fix "field name mismatch" error

schbetsy:
Fixes this issue:

screen shot 2017-04-07 at 2 59 53 pm

Activity Year for this edit is always 2017, It isn't a property of an individual LAR.


schbetsy:
With this change, it now looks like this:

screen shot 2017-04-07 at 4 47 23 pm

lesgou:
Just to clarify, activity year is not in a lar row, but it is a data field provided in the transmittal sheet, and only for this specific edit and year, it is 2017.

PR #924, opened 2017-04-10
Add name to MSA response

### LINKED ISSUE #892: Add MSA Title to IRS response
The MSA title will be the "CBSA Title" field in the [`jul_2015_cbsa.csv`](https://github.com/cfpb/hmda-platform/blob/master/census/src/main/resources/jul_2015_cbsa.csv) file.  The field `name` should be added to the [MSA object](https://github.com/cfpb/hmda-platform/blob/master/query/src/main/scala/hmda/query/model/filing/Msa.scala), and can be looked up based on the `id` field.

nickgrippin:
Closes #892

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -536,6 +536,7 @@ Example response:
   "msas": [
     {
       "id": "123",
+      "name": "Some, Place",
       "totalLARS": 4,
       "totalAmount": 123,
       "conv": 4,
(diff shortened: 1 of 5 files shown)


schbetsy:
The models for MSA and IRS will not exist in the same way once #926 is merged, so this change will unfortunately get clobbered. Maybe you could use #926 as the base branch and make your changes from there? Otherwise, I can try to add this functionality to it.

schbetsy:
Looks good. If no one objects, I'm going to go ahead and merge, so I can start incorporating this change into #926.

PR #926, opened 2017-04-10
Pagination for IRS Endpoint

### LINKED ISSUE #864: Add pagination to IRS
Using the same idea/format from #821, add pagination to the IRS endpoint.

schbetsy:
Closes #864 

Known issue: This PR doesn't update the `SubmissionIrsPathsSpec`. It is still commented out to avoid the `slick.util.AsyncExecutor` issue from previous sprints.

### DIFF
--- a/api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
+++ b/api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
@@ -22,11 +22,11 @@ trait MsaProtocol extends DefaultJsonProtocol with ParserResultsProtocol {
     override def read(json: JsValue): IrsResponse = {
       json.asJsObject.getFields("msas", "summary", "total", "_links") match {
         case Seq(JsArray(msas), JsObject(summ), JsNumber(tot), JsObject(links)) =>
-          val msaCollection: Seq[MsaWithName] = msas.map(_.convertTo[MsaWithName])
-          val path: String = PaginatedResponse.staticPath(links("href").convertTo[String])
-          val currentPage: Int = PaginatedResponse.currentPage(links("self").convertTo[String])
-          val summary: MsaSummary = JsObject(summ).convertTo[MsaSummary]
-          val total: Int = tot.intValue
+          val msaCollection = msas.map(_.convertTo[MsaWithName])
+          val path = PaginatedResponse.staticPath(links("href").convertTo[String])
+          val currentPage = PaginatedResponse.currentPage(links("self").convertTo[String])
+          val summary = JsObject(summ).convertTo[MsaSummary]
+          val total = tot.intValue
           IrsResponse(msaCollection.toList, summary, path, currentPage, total)
 
         case _ => throw DeserializationException("IRS Summary expected")
--- a/api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
+++ b/api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
@@ -1,6 +1,6 @@
 package hmda.api.protocol.processing
 
-import hmda.api.model.{ IrsResponse, PaginatedResponse, PaginationLinks }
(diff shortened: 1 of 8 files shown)


census/src/main/scala/hmda/census/model/CbsaLookup.scala
```diff
@@ -34,8 +34,10 @@ object CbsaLookup extends CbsaResourceUtils {
     }
   }.toSeq
 
-  val codeMap: Map[String, String] = {
-    values.map(x => (x.cbsa, x.cbsaTitle)).toMap
+  def nameFor(cbsaId: String): String = {
+    values.find(_.cbsa == cbsaId)
```
nickgrippin: This new approach is interesting, is there a performance improvement gained by switching to `find`?

query/src/main/scala/hmda/query/model/filing/Msa.scala
```diff
@@ -33,6 +16,40 @@ case class MsaWithName(
     homePurchase: Int,
     homeImprovement: Int,
     refinance: Int
+) {
+  def addName: MsaWithName = MsaWithName(
```
nickgrippin: The return type on this method is pretty clear, I don't think it needs to be specified in the declaration.

census/src/main/scala/hmda/census/model/CbsaLookup.scala
schbetsy: Just an improvement in modularity of code and separation of concerns. Don't know about performance.

query/src/main/scala/hmda/query/model/filing/Msa.scala
schbetsy: Is it important to remove it?

census/src/main/scala/hmda/census/model/CbsaLookup.scala
schbetsy: I can add the map back in if we're concerned about performance.

query/src/main/scala/hmda/query/model/filing/Msa.scala
jmarin: In general, adding return types for methods is encouraged in typical `Scala` code.

api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
```diff
@@ -1,11 +1,36 @@
...
+
+    override def read(json: JsValue): IrsResponse = {
+      json.asJsObject.getFields("msas", "summary", "total", "_links") match {
+        case Seq(JsArray(msas), JsObject(summ), JsNumber(tot), JsObject(links)) =>
+          val msaCollection: Seq[MsaWithName] = msas.map(_.convertTo[MsaWithName])
+          val path: String = PaginatedResponse.staticPath(links("href").convertTo[String])
+          val currentPage: Int = PaginatedResponse.currentPage(links("self").convertTo[String])
+          val summary: MsaSummary = JsObject(summ).convertTo[MsaSummary]
+          val total: Int = tot.intValue
+          IrsResponse(msaCollection.toList, summary, path, currentPage, total)
```
jmarin: I would remove the types for these `val` declarations if the compiler is happy with it. In general, `val` or `var` declarations that can use type inference should do so.

census/src/main/scala/hmda/census/model/CbsaLookup.scala
jmarin: I have a strong opinion about "performance". In the absence of an objective measurement that proves a performance issue, there isn't a performance issue. So I'm ok with this change based on what I see in the code

api/src/main/scala/hmda/api/protocol/processing/MsaProtocol.scala
schbetsy: Interesting. I'm fine with that (and the compiler seems to be too). Done in 8b4a628a46e6915f35ac7df23278501a30d0d983.

PR #930, opened 2017-04-13
Update submission status when macro and quality edits are verified

### LINKED ISSUE #906: Submission status updating on edit verification
Currently, if all macro/quality edits are verified, the submission status will remain at 8 (validated with errors).  The status needs to be updated with the following two behaviors:

- If all edits are verified, the submission status needs to be updated to 9 (validated)
- If an edit is un-verified, the submission status needs to be updated to 8 (validated with errors)

Note: these numbers may change with #849

jmarin:
Closes #906

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
@@ -234,8 +234,7 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor 
       self ! CompleteValidation(replyTo)
 
     case CompleteValidation(replyTo) =>
-      if (state.larSyntactical.isEmpty && state.larValidity.isEmpty && state.qualityVerified && state.macroVerified
-        && state.tsSyntactical.isEmpty && state.tsValidity.isEmpty && state.tsQuality.isEmpty) {
+      if (state.syntacticalErrors.isEmpty && state.validityErrors.isEmpty && state.qualityVerified && state.macroVerified) {
         log.debug(s"Validation completed for $submissionId")
         replyTo ! ValidationCompleted(submissionId)
       } else {
--- a/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
@@ -32,7 +32,7 @@ object HmdaFileValidator {
   case class ValidationStarted(submissionId: SubmissionId) extends Event
   case class ValidateMacro(source: LoanApplicationRegisterSource, replyTo: ActorRef) extends Command
   case class CompleteMacroValidation(errors: LarValidationErrors, replyTo: ActorRef) extends Command
-  case class VerifyEdits(editType: ValidationErrorType, verified: Boolean) extends Command
+  case class VerifyEdits(editType: ValidationErrorType, verified: Boolean, replyTo: ActorRef) extends Command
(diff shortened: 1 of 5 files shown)


persistence/src/main/scala/hmda/persistence/processing/SubmissionFSM.scala
```diff
@@ -230,6 +230,15 @@ class SubmissionFSM(submissionId: SubmissionId)(implicit val domainEventClassTag
       val status = hmda.model.fi.Signed
```
schbetsy: This block means that the FSM allows a submission to go from `ValidatedWithErrors` to `Signed` with the `Sign` event. (I added this back when we misunderstood how statuses worked.) Should this block be removed?

schbetsy:
This also needs to ensure that if the edits become un-verified (like if the verify-edits endpoint receives `verified: false` in the payload), the status goes back to `validated with errors`.

persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -209,7 +209,7 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor
       self ! CompleteValidation(replyTo)
 
     case CompleteValidation(replyTo) =>
-      if (state.larSyntactical.isEmpty && state.larValidity.isEmpty && state.larQuality.isEmpty && state.larMacro.isEmpty
+      if (state.larSyntactical.isEmpty && state.larValidity.isEmpty && state.qualityVerified && state.macroVerified
         && state.tsSyntactical.isEmpty && state.tsValidity.isEmpty && state.tsQuality.isEmpty) {
```
schbetsy: Looks like `&& state.tsQuality.isEmpty` was left in by accident. That shouldn't be a check in this conditional.

Also, we now have `validityErrors` and `syntacticalErrors` convenience methods on the `HmdaFileValidationState`, so you could use those to simplify this expression if that would help readability.

persistence/src/main/scala/hmda/persistence/processing/SubmissionFSM.scala
jmarin: I don't know, it's a good question. I believe the frontend might prevent this scenario from happening, but not sure. In any case, if this change is needed I would like to do it in a different PR.

schbetsy:
The status update seems to work when verifying and un-verifying. 👍 

One thing I noticed: the response from the verify-edits endpoint doesn't have the most current status (When I POST-ed to verify edits, the response showed `validated with errors` even though it was fully validated. I had to request the status again to see that it had updated to `validated`). 

We should probably fix that (in this PR or elsewhere), and we should make sure the front-end knows this is happening, in case they were planning to use this response for logic.

awolfe76:
> When I POST-ed to verify edits, the response showed validated with errors even though it was fully validated. I had to request the status again to see that it had updated to validated.

Yeah, that's basically what is stopping the signature now. That response isn't coming back as validated so we can't update the app state correctly. I would think we would need to fix that ... if I'm understanding this correctly.

_We could make an additional call but that seems weird to me._

schbetsy:
Maybe this endpoint could work like the signature endpoint, where it gets the submission status after the sign action is done (or in this case, when the "verify edits" action is done).

jmarin:
@schbetsy @awolfe76 Can we move this additional step to another issue / PR? I don't have time to continue with this and am going to have to delegate that part

schbetsy:
Works for me. 👍

awolfe76:
Yep, sounds good to me too.

PR #931, opened 2017-04-13
Create Filings when running Panel

### LINKED ISSUE #915: Create filing when institution is created
A filing for the current year should be created for every institution loaded from panel (add this to `PanelCsvParser`)

nickgrippin:
This is one possible approach, using `BroadcastHub` to attach two consumers to the parsing `Source`, while still utilizing the advantages of Akka Streams.  I'm having issues when running panel on master, so this is still a work in progress.

Another approach is commented out, as it blocks the creation of institutions until the filing is created.

Closes #915

### DIFF
--- a/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
+++ b/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
@@ -69,7 +69,7 @@ object PanelCsvParser extends InstitutionComponent {
 
     fromGraph
       .runForeach(i => {
-        val f = Filing(i.activityYear.toString, i.id, NotStarted, filingRequired = false, 0, 0)
+        val f = Filing(i.activityYear.toString, i.id)
         for {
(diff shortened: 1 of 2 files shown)


panel/src/main/scala/hmda/panel/PanelCsvParser.scala
```diff
@@ -43,25 +44,36 @@ object PanelCsvParser extends InstitutionComponent {
       sys.exit(1)
     }
 
-    val timeout = 5.seconds
-
     println("Cleaning DB...")
-    Await.result(repository.dropSchema(), timeout)
+    Await.result(repository.dropSchema(), 5.seconds)
```
nickgrippin: There's an `implicit val timeout` declared on line 31, it was causing some compilation issues.

schbetsy:
Since the conflict was simple, I fixed it in GitHub's UI. 

I haven't seen this work locally yet, but if y'all have, I guess I'm okay with merging this.

PR #939, opened 2017-04-26
Q012

### LINKED ISSUE #180: Validation (2017): Q012 (Tax Identification (ID) Number)
**Edit #:** Q012
**Type:** Quality
**Category:** Transmittal Sheet (only)
**Column:** Tax Identification (ID) Number

The reported Tax ID number on the transmittal sheet of your HMDA data file does not match the Tax ID number reported in the previous calendar year submission.

Additional resources:

nickgrippin:
Closes #180


validation/src/main/scala/hmda/validation/engine/ts/quality/TsQualityEngine.scala
```diff
@@ -1,21 +1,34 @@
...
 import hmda.validation.engine.Quality
 import hmda.validation.engine.ts.TsCommonEngine
 import hmda.validation.rules.ts.quality._
 
+import scala.concurrent.{ ExecutionContext, Future }
+
 trait TsQualityEngine extends TsCommonEngine with ValidationApi {
 
-  def checkQuality(ts: TransmittalSheet, ctx: ValidationContext): TsValidation = {
+  def checkQuality(ts: TransmittalSheet, ctx: ValidationContext)(implicit system: ActorSystem, materializer: ActorMaterializer, ec: ExecutionContext): Future[TsValidation] = {
```
jmarin: I'd suggest you use context bounds as you have done in Q012 to condense this code with implicits.

validation/src/main/scala/hmda/validation/engine/ts/quality/TsQualityEngine.scala
```diff
@@ -1,21 +1,34 @@
...
-    ).map(check(_, ts, tsId, Quality, true))
+    ).map(x => Future(check(x, ts, tsId, Quality, ts = true)))
+
+    val allChecks = Future.sequence(checks :+ q012(ts, ctx))
+
+    allChecks.map(c => validateAll(c, ts))
+  }
 
-    validateAll(checks, ts)
+  private def q012(ts: TransmittalSheet, ctx: ValidationContext)(implicit system: ActorSystem, materializer: ActorMaterializer, ec: ExecutionContext) = {
```
jmarin: Same comment about context bounds for implicit resolution in this method

validation/src/main/scala/hmda/validation/ValidationStats.scala
```diff
@@ -22,7 +23,14 @@ object ValidationStats {
   case class ValidationStatsState(stats: Seq[SubmissionStats] = Nil) {
     def updated(event: Event): ValidationStatsState = event match {
       case SubmissionStatsAdded(s) =>
-        ValidationStatsState(stats :+ s)
+        val matchingSubs = stats.filter(stat => stat.id == s.id)
+        if (matchingSubs.isEmpty) {
+          ValidationStatsState(stats :+ s)
+        } else {
+          val oldSub = matchingSubs.head
+          val newSub = oldSub.copy(id = s.id, totalLars = oldSub.totalLars + s.totalLars, taxId = oldSub.taxId + s.taxId)
```
schbetsy: Why does this line add the old taxId and new taxId together?

validation/src/main/scala/hmda/validation/ValidationStats.scala
```diff
@@ -22,7 +23,14 @@ object ValidationStats {
...
+          ValidationStatsState(stats.filterNot(s => s == oldSub) :+ newSub)
```
schbetsy: This `s` is confusing because that's the name of a different variable earlier in this method. Could you change one of them to something clearer?

validation/src/main/scala/hmda/validation/ValidationStats.scala
nickgrippin: This was my approach to modifying a currently existing `SubmissionStats` object.  When the Tax ID is found, [here](https://github.com/cfpb/hmda-platform/pull/939/files#diff-bf281a4df82187ea62138ba687b8007eR152), any unneeded fields are filled with an empty string or zero (in this case, a zero for the LAR count).  Later, once the total LAR count has been calculated [here](https://github.com/cfpb/hmda-platform/pull/939/files#diff-6ca062a943a4a1d03f3bddb0618aa3e9R51), an empty string is sent in place of the tax ID.  If this is confusing or unwanted, I can modify the approach.

schbetsy: Okay, I see what it's doing. It seems like this `SubmissionStatsAdded` event is doing two different jobs. `totalLars` and `taxId` will never get added at the same time, so we don't need this method to do both. I think if we had something like `SubmissionTotalsAdded` and `SubmissionTaxIdAdded` separately, the purpose of the code would be much clearer.

nickgrippin: Good call!  That change should be up soon., just need to tweak some tests.

validation/src/test/scala/hmda/validation/engine/ts/quality/TsQualityEngineSpec.scala
```diff
@@ -24,7 +28,7 @@ class TsQualityEngineSpec
     val ts = line.map(l => TsCsvParser(l))
 
     ts.foreach { ts =>
-      checkQuality(ts.right.get, ctx).isSuccess mustBe true
+      checkQuality(ts.right.get, ctx).map(validation => validation.isSuccess mustBe true)
```
schbetsy: Since this assertion is in the result of a future, it never gets tested. I think you'll have to change this spec to use `AsyncWordSpec`

PR #940, opened 2017-04-27
Prevent ValidationStats Timeout

### LINKED ISSUE #936: Validation process not completing
The validation process will occasionally stop on `Validation started for submission`, and never complete.  This behavior has been observed happening infrequently when running just hmda-platform and happening regularly when running the entire project.

nickgrippin:
The ValidationStats actor was receiving a timeout and shutting itself down, which would cause the yield statements on both Q011 and Q012 to get stuck.  This PR overrides the `preStart` method so the actor doesn't stop itself.

Closes #936

### DIFF
--- a/validation/src/main/scala/hmda/validation/ValidationStats.scala
+++ b/validation/src/main/scala/hmda/validation/ValidationStats.scala
@@ -32,6 +32,10 @@ class ValidationStats extends HmdaPersistentActor {
 
   override def persistenceId: String = s"$name"
 
+  override def preStart(): Unit = {
+    log.info(s"Actor started at ${self.path}")
+  }
+
   var state = ValidationStatsState()
 
   override def updateState(event: Event): Unit = {


schbetsy:
Does this change mean the ValidationStats actor just stays alive forever?

I'm curious; is it worth exploring (slash did you already look into) an alternative solution where `Q011` uses a more fault-tolerant strategy for getting a reference to the ValidationStats actor? Like maybe instead of using `system.actorSelection("/user/validation-stats")` to get it, we could use the pattern that exists in the `HmdaSupervisor`. Like, it could have a `FindValidationStats` message, which could have a "find or create" behavior.

(Not sure if I've explained myself clearly--let me know if I can clarify)

jmarin:
The `ValidationStats` stays up and running, yes, we are treating it as a top level actor. The mechanism that I implemented with the `Supervisor` is actually less robust than what is ideal, and I would like to remove it eventually. Doing `system.actorSelction` is brittle and not very fault tolerant, especially in a clustered environment.

PR #941, opened 2017-04-28
Q130

### LINKED ISSUE #181: Validation (2017): Q130 (# of Loan Applications)
**Edit #:** Q130
**Type:** Quality
**Category:** Transmittal Sheet (only)
**Column:** # of Loan Applications

The number of loan/application records received in this transmission file per respondent does not = the total number of loan/application records reported in this respondent’s transmission or the total number of loan/application records in this submission is missing from the transmittal sheet.

Additional resources:

nickgrippin:
Depends on #939 

Closes #181


persistence/src/main/scala/hmda/persistence/processing/HmdaFileParser.scala
```diff
@@ -71,6 +73,8 @@ class HmdaFileParser(submissionId: SubmissionId) extends HmdaPersistentActor {
   val config = ConfigFactory.load()
   val flowParallelism = config.getInt("hmda.actor-flow-parallelism")
 
+  val submissionLarStats = context.actorOf(SubmissionLarStats.props(submissionId))
```
jmarin: This is a problem. This is a second instance of `SubmissionLarStats`, different from the one initially implemented in `HmdaFileValidator`. As a persistent actor, it should be unique. One way to solve this might be to extract this actor and start it somewhere else and pass it as a reference in the constructor to both `HmdaFileValidator` and `HmdaFileParser`

nickgrippin: Working on a fix, moving the declaration into `SubmissionManager` and passing references.  I'll have it up shortly.

validation/src/main/scala/hmda/validation/SubmissionLarStats.scala
```diff
@@ -40,15 +43,26 @@ class SubmissionLarStats(submissionId: SubmissionId) extends HmdaPersistentActor
...
 
   override def receiveCommand: Receive = super.receiveCommand orElse {
+    case s: String =>
+      totalSubmittedLars = totalSubmittedLars + 1
+
     case LarValidated(_, _) =>
-      totalLars = totalLars + 1
+      totalValidatedLars = totalValidatedLars + 1
+
+    case CountSubmittedLarsInSubmission =>
```
schbetsy: I could use help understanding why we need two different counts--one for "submitted" lars, and one for "validated" lars. Under what circumstances would those counts be different for a single submission?

schbetsy: Oh, I think I figured it out. Is it because Q130 is a Quality edit (as opposed to Macro), and the "validated" count doesn't get calculated soon enough?

nickgrippin: Exactly 👍

persistence/src/main/scala/hmda/persistence/processing/SubmissionManager.scala
```diff
@@ -45,6 +46,7 @@ class SubmissionManager(submissionId: SubmissionId) extends HmdaActor {
   val supervisor = context.parent
   val hmdaFilingF = (supervisor ? FindHmdaFiling(period)).mapTo[ActorRef]
 
+  val submissionLarStats: ActorRef = context.actorOf(SubmissionLarStats.props(submissionId), s"submission-lar-stats-${submissionId.toString}")
```
schbetsy: It looks like this `val` is never used in this file. If that's the case, what is this line for? Does it ensure that the `SubmissionLarStats` actor gets started?

schbetsy: Oh, I might have figured out this one too. Is it to assign the name of the actor (`submission-lar-stats-${submissionId.toString}`) so that the `HmdaFileValidator` adn `HmdaFileParser` can find it?

PR #942, opened 2017-04-28
Allow IRS to show Metropolitan Division names

### LINKED ISSUE #933: MSA Response
The MSA field in a LAR can also be an MD.  Update the MSA name lookup to also check for any MD matches.  The ideal approach would be to add a mapping of MD Code -> MD Title on the `codeMap` in `CbsaLookup`.

Also, sort the IRS response by the ID.

schbetsy:
Closes #933

### DIFF
--- a/census/src/main/scala/hmda/census/model/CbsaLookup.scala
+++ b/census/src/main/scala/hmda/census/model/CbsaLookup.scala
@@ -38,8 +38,17 @@ object CbsaLookup extends CbsaResourceUtils {
     values.map(x => (x.cbsa, x.cbsaTitle)).toMap
   }
 
-  def nameFor(cbsaId: String): String = {
-    cbsaIdToNameMap.getOrElse(cbsaId, "NA")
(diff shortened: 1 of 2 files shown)


census/src/main/scala/hmda/census/model/CbsaLookup.scala
```diff
@@ -38,8 +38,17 @@ object CbsaLookup extends CbsaResourceUtils {
...
+    values.map(c => (c.metroDiv, c.metroDivTitle)).toMap
+  }
+
+  // Returns CBSA Title if `id` is a CBSA Code, or
+  //   Metropolitan Division Title if `id` is a MD Code.
+  def nameFor(id: String): String = {
+    val cbsa = cbsaIdToNameMap.get(id)
+    val md = mdIdToNameMap.get(id)
+    val na = Some("NA")
+    Seq(cbsa, md, na).flatten.head
```
nickgrippin: Nice coding here!  Really good use of `flatten` and options 👍

PR #951, opened 2017-05-02
Institution serializer

### LINKED ISSUE #830: InstitutionPersistence serializer

jmarin:
Closes #830

### DIFF
--- a/persistence-model/src/main/resources/application-dev.conf
+++ b/persistence-model/src/main/resources/application-dev.conf
@@ -2,17 +2,27 @@ akka {
   loglevel = "INFO"
   actor.warn-about-java-serializer-usage = off
   persistence {
-    journal.plugin = "cassandra-journal"
-    snapshot-store.plugin = "cassandra-snapshot-store"
+    journal {
+      plugin = "akka.persistence.journal.leveldb"
+      leveldb {
+        dir = "target/journal"
+        native = false
+      }
+    }
+    snapshot-store {
+      plugin = "akka.persistence.snapshot-store.local"
+      local {
+        dir = "target/snapshots"
+      }
+    }
     query {
-      journal.id = "cassandra-query-journal"
+      journal {
+        id = "akka.persistence.query.journal.leveldb"
+      }
     }
   }
 }
 
-akka.persistence.journal.plugin = "inmemory-journal"
-akka.persistence.snapshot-store.plugin = "inmemory-snapshot-store"
-akka.persistence.query.journal.id = "inmemory-read-journal"
 
 hmda {
   actor-lookup-timeout = 5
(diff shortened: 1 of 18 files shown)


persistence-model/src/main/scala/hmda/persistence/serialization/institutions/InstitutionProtobufSerializer.scala
```diff
@@ -0,0 +1,29 @@
...
+
+class InstitutionProtobufSerializer extends SerializerWithStringManifest {
+  override def identifier: Int = 1000
+
+  override def manifest(o: AnyRef): String = o.getClass.getName
+
+  final val InstitutionCreatedManifest = classOf[InstitutionCreated].getName
+  final val InstitutionModifiedManifest = classOf[InstitutionModified].getName
+
+  override def toBinary(o: AnyRef): Array[Byte] = o match {
```
schbetsy: According to IntelliJ, these methods (`toBinary` and `fromBinary`) are currently only used in specs. Do they have uses planned? (or do they already have uses I didn't see?)

jmarin: These are used by `Akka Persistence` when storing / reading objects to / from the journal. They are not called directly from our code. It doesn't surprise me that `IntelliJ` gets confused.

persistence-model/src/main/resources/application-dev.conf
```diff
@@ -1,18 +1,37 @@
...
   loglevel = "INFO"
-  actor.warn-about-java-serializer-usage = off
+  actor {
+    warn-about-java-serializer-usage = off
+    serializers {
+      institution = "hmda.persistence.serialization.institutions.InstitutionProtobufSerializer"
+    }
+    serialization-bindings {
+      "hmda.persistence.messages.events.institutions.InstitutionEvents$InstitutionCreated" = institution
+      "hmda.persistence.messages.events.institutions.InstitutionEvents$InstitutionModified" = institution
```
schbetsy: Testing my knowledge: These two events (`InstitutionCreated` and `InstitutionModified`) are the only ones that need to be listed here because they will be part of the Event Stream. All of the lower-level structures (like `Institution`, `Parent`, `InstitutionType`, etc.) are only needed to serialize the two top-level messages. I'm guessing that's also why `.toBinary` and `.fromBinary` only handle those two message types.

Is that about right? (And if I haven't been clear, let me know; I can try to articulate better)

jmarin: Yes, that is right. These two events are what get persisted as part of `InstitutionPersistence`.

PR #953, opened 2017-05-02
Return latest status after verifying edits

### LINKED ISSUE #947: Make sure correct status is returned when verifying edits
See comments in #930

schbetsy:
Closes #947

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
@@ -251,21 +251,20 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor 
       self ! LarValidationErrors(e.errors)
       self ! CompleteValidation(replyTo)
 
-    case CompleteValidation(replyTo) =>
+    case CompleteValidation(replyTo, originalSender) =>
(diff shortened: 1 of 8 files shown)


persistence/src/main/scala/hmda/persistence/processing/HmdaFileValidator.scala
```diff
@@ -233,21 +233,20 @@ class HmdaFileValidator(submissionId: SubmissionId) extends HmdaPersistentActor
...
+        replyTo ! ValidationCompletedWithErrors(originalSender)
       }
 
     case VerifyEdits(editType, v, replyTo) =>
       if (editType == Quality || editType == Macro) {
         persist(EditsVerified(editType, v)) { e =>
           updateState(e)
-          self ! CompleteValidation(replyTo)
-          sender() ! EditsVerified(editType, v)
+          self ! CompleteValidation(replyTo, Some(sender()))
```
jmarin: This could be a problem. The `sender()` might get created in a different thread than the original message processing due to the async nature of `persist`. To always be safe, the best thing to do in the first line after `VerifyEdits` is to "fix" the reference to the original sender, like so:

```scala
val client = sender()
```
And then replace `Some(sender())` with `Some(client)` (or whatever you want to call that `val`)

schbetsy: Good to know. Fixed in 10a22722af82814f3572499db3d92503b01b2c19.

PR #954, opened 2017-05-03
Load Panel through Admin API

### LINKED ISSUE #938: Load panel data through Admin API
The current panel data is being loaded through a separate `ActorSystem`. This causes problems with the serialization version of messages in the journal.

nickgrippin:
Note for testing: the API must be running in order for the panel data to load.  I had one terminal with the API running (`HMDA_IS_DEMO=false`), and ran panel in a separate window.  The panel process won't output any logs after "Creating schema", but you should see output from the API window.

Closes #938

### DIFF
--- a/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
+++ b/panel/src/main/scala/hmda/panel/PanelCsvParser.scala
@@ -3,39 +3,33 @@ package hmda.panel
 import java.io.File
 
 import akka.NotUsed
-import akka.actor.{ ActorRef, ActorSystem }
-import akka.util.{ ByteString, Timeout }
-import akka.pattern.ask
+import akka.actor.{ActorRef, ActorSystem}
+import akka.util.{ByteString, Timeout}
 import akka.stream.ActorMaterializer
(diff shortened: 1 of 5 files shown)


panel/src/main/scala/hmda/panel/PanelCsvParser.scala
```diff
@@ -3,39 +3,35 @@ package hmda.panel
...
   implicit val timeout: Timeout = Timeout(5.second)
   implicit val ec = system.dispatcher
   val log = LoggerFactory.getLogger("hmda")
 
-  val supervisor = createSupervisor(system)
-  val institutionPersistenceF = (supervisor ? FindActorByName(InstitutionPersistence.name)).mapTo[ActorRef]
+  val repository = new InstitutionRepository(hmda.query.DbConfiguration.config)
 
-  val repository = new InstitutionRepository(config)
+  val config = ConfigFactory.parseFile(new File("api/src/main/resources/application.conf")).resolve()
```
jmarin: I think you can replace this line with `ConfigFactory.load()`

PR #965, opened 2017-05-09
Fix TS parsing errors in API response

### LINKED ISSUE #960: Parser Errors API response: Pagination should only affect LAR results
This will solve https://github.com/cfpb/hmda-platform-ui/issues/523

We can always return the TS results regardless of which page of LAR results we're on.

schbetsy:
Closes #960

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -276,10 +276,10 @@ Example response, with HTTP code 400:
 | page | Integer. If blank, will default to page 1. Page size is 20 lines of errors. |
 
 
-This endpoint is paginated. The response contains 3 fields of pagination metadata:
+The `larErrors` array in this endpoint is paginated. The response contains 3 fields of pagination metadata:
 
- - `total`: total number of parser errors for this file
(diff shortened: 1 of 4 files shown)


awolfe76:
@schbetsy, a couple questions:
- Will this change the format of the response at all (doc update)?
- Does this mean that the pagination for `parseErrors` only applies to the LAR errors, never the TS errors? What would happen if there were +20 errors on the TS _(TS should only have 21 possible rows anyway so maybe we just never apply pagination there)_? And, if this does mean that only LAR errors are paginated maybe we should note that in the doc too.

schbetsy:
>Will this change the format of the response at all (doc update)?
Nope, format stays the same. Let me know if you need it changed.

>Does this mean that the pagination for parseErrors only applies to the LAR errors, never the TS errors? 
For 2017, the largest number of parser errors the TS can have is 5. So pagination won't be necessary for TS.

>And, if this does mean that only LAR errors are paginated maybe we should note that in the doc too.
You're right. Updates in c8a234dd. Let me know if there are other edits that would make it clearer.

PR #967, opened 2017-05-10
Q071

### LINKED ISSUE #238: Validation (2017): Q071 (Type of Purchaser)
**Edit #:** Q071
**Type:** Macro
**Column:** Type of Purchaser

If action taken type = 1 or 6, purpose of loan = 1 or 3, property type = 1 or 2, loan type = 2; and if the percentage of these loans that are sold to Ginnie Mae in the current year is less than the percentage of the same type of loans sold to Ginnie Mae in the prior year, then the percentage difference between the two years should be less than 10%; or if the number of these loans for the current year is >= 2, 500, then the percentage of these loans sold to Ginnie Mae for the current year should be > 30%.

Additional resources:

schbetsy:
Closes #238


validation/src/main/scala/hmda/validation/SubmissionLarStats.scala
```diff
@@ -2,28 +2,37 @@ package hmda.validation
...
   def createSubmissionStats(system: ActorSystem, submissionId: SubmissionId): ActorRef = {
     system.actorOf(SubmissionLarStats.props(submissionId))
   }
 
-  case class SubmissionLarStatsState(totalSubmitted: Int = 0, totalVerified: Int = 0) {
+  case class SubmissionLarStatsState(
+      totalSubmitted: Int = 0,
+      totalValidated: Int = 0,
+      q071Total: Int = 0,
+      q071SoldTotal: Int = 0
```
nickgrippin: Just thinking ahead here to the other edits: are we OK with the state having an additional 8 fields added?  We're going to also have `q070Total`, `q072Total`, `q073Total`, etc... along with all the related changes in this file and the `ValidationStats` file.

I don't have a solution here, and I'm not critiquing the code, but I just wanted to draw attention to the potential future of this path.

schbetsy: Yeah... that's definitely the direction this is headed. I'm okay with it for now. I bet, after we add a few more edits, the right kind of refactor/optimization will be more obvious.

validation/src/main/resources/application.conf
```diff
@@ -158,6 +158,11 @@ hmda.validation.macro {
   Q065 {
     hoepaLoanLimit = 200
   }
+  Q071 {
+    currentYearThreshold = 50
```
nickgrippin: In reference to Q071, how are these values used in the calculations?  When reading the ticket, I see a threshold of 2500, a current year ratio of 30%, and a 10% difference between the two years.

schbetsy: Those are test values. I'm planning to change them before I PR. Hence the "WIP:".

validation/src/test/scala/hmda/validation/rules/lar/macro/Q071Spec.scala
```diff
@@ -0,0 +1,169 @@
...
+      val ctx1 = ValidationContext(None, Some(currentYear))
+      Q071.inContext(ctx1).name mustBe "empty"
+
+      val ctx2 = ValidationContext(Some(Institution.empty), None)
+      Q071.inContext(ctx2).name mustBe "empty"
+    }
+
+  }
+
+  private def ctx(institutionId: String, currentYear: Int = 2017): ValidationContext = {
```
nickgrippin: Looks like a lot of these helper methods are going to be shared between several other tests... Would it be possible to move them into a separate file?

schbetsy: Good idea. I'm on it.

PR #968, opened 2017-05-11
Filing serializer

### LINKED ISSUE #829: FilingPersistence serializer

jmarin:
Closes #829


model/shared/src/main/scala/hmda/model/fi/Filing.scala
```diff
@@ -17,7 +17,7 @@ case object Completed extends FilingStatus {
   override def message: String = "completed"
 }
 case object Cancelled extends FilingStatus {
-  override def code: Int = 1
+  override def code: Int = -1
```
schbetsy: In `filing-status.md`, this status's code is `4`. Let's update the documentation to reflect this change.

persistence-model/src/main/scala/hmda/persistence/messages/events/institutions/FilingEvents.scala
```diff
@@ -0,0 +1,10 @@
+package hmda.persistence.messages.events.institutions
+
+import hmda.model.fi.Filing
+import hmda.persistence.messages.CommonMessages.Event
+
+object FilingEvents {
+  sealed trait SubmissionEvent extends Event
```
schbetsy: `SubmissionEvents.scala` also declares a trait called `SubmissionEvent`. Why does it need to be declared twice (or is this a typo)? What is it for?

PR #985, opened 2017-05-24
Panel data

### LINKED ISSUE #982: Load panel file
Load panel file (version as of May 22nd 2017, release 1.5) into Cassandra.

nickgrippin:
Addresses #982


schbetsy:
I'm confused. How does this close #982?

nickgrippin:
@schbetsy: the ticket has two parts, one of them is adding the most updated panel file to the hmda-platform, which is what this PR addresses.  The second part is actually loading the file into our running application.  I'll be working with Juan on the second part.

panel/src/main/scala/hmda/panel/PanelCsvParser.scala
```diff
@@ -71,4 +70,15 @@ object PanelCsvParser extends InstitutionComponent with WriteInstitutionProtocol
 
   private def byte2StringFlow: Flow[ByteString, String, NotUsed] =
     Flow[ByteString].map(bs => bs.utf8String)
+
+  private def createInstitutionDB() {
+    println("Creating new schema...")
```
jmarin: I know this is a console app, but could we add proper logging here as opposed to `println`?

jmarin:
I agree with @schbetsy in that this doesn't cover the whole work to be done. Can we split it in two tickets, or say that this ticket addresses #982 but doesn't close it? I don't want to lose track of where we are with this process.

nickgrippin:
This PR now only adds the panel file, without any changes to the parser.

PR #989, opened 2017-05-26
SubmissionLarStats serializer

### LINKED ISSUE #963: SubmissionLarStats serializer

nickgrippin:
Closes #963


persistence-model/src/test/scala/hmda/persistence/serialization/validation/SubmissionLarStatsProtobufConverterSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+    forAll(Gen.choose(0, 50000)) { num =>
+      val macroStatsUpdated = MacroStatsUpdated(
+        totalValidated = num + difGen.sample.get,
+        q070Total = num + difGen.sample.get,
+        q070Sold = num + difGen.sample.get,
+        q071Total = num + difGen.sample.get,
+        q071Sold = num + difGen.sample.get,
+        q072Total = num + difGen.sample.get,
+        q072Sold = num + difGen.sample.get,
+        q075Ratio = num + difGen.sample.get,
```
schbetsy: Looks like the values in this test for `q075Ratio` and `q076Ratio` will be `Int`s instead of `double`s. If we wanted to test with non-integer values, maybe divide these values by 100 or something?

persistence-model/src/test/scala/hmda/persistence/serialization/validation/SubmissionLarStatsProtobufSerializerSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+      val bytes = serializer.toBinary(msg)
+      serializer.fromBinary(bytes, serializer.SubmittedLarsUpdatedManifest) mustBe msg
+    }
+  }
+
+  property("MacroStatsUpdated messages must be serialized to binary and back") {
+    val difGen = Gen.choose(-50000, 50000)
+    forAll(Gen.choose(0, 50000)) { num =>
+      val msg = MacroStatsUpdated(
+        totalValidated = num + difGen.sample.get,
```
schbetsy: I'm not sure I see the value in using `num` as the base for each value. What's the difference between this strategy and just sampling `Gen.choose(0,50000)` for all of the different values in the object?

persistence-model/src/test/scala/hmda/persistence/serialization/validation/SubmissionLarStatsProtobufConverterSpec.scala
nickgrippin: The numbers will be converted to a `double` when they're assigned, there isn't a need to perform the extra arithmetic.

schbetsy: Yeah, I'm not worried about it failing to compile or anything. I'm just talking from a test integrity standpoint. It's better to use realistic values in specs, so we can ensure the code works the way it will in production.

schbetsy: Anyway, I'd vote to make the same change here as you did in `SubmissionLarStatsProtobufSerializerSpec.scala`

PR #993, opened 2017-05-31
Update docs for Zookeeper variables

nickgrippin:


schbetsy:
I get errors like this when I try to run the application locally with these instructions:

```
[WARN ][hmda-akka.actor.default-dispatcher-29] Couldn't join seed nodes after [2] attmpts, will try again. seed-nodes=[akka.tcp://hmda@127.0.0.1:49282]
[WARN ][hmda-akka.actor.default-dispatcher-29] Association with remote system [akka.tcp://hmda@127.0.0.1:49282] has failed, address is now gated for [5000] ms. Reason: [Association failed with [akka.tcp://hmda@127.0.0.1:49282]] Caused by: [Connection refused: /127.0.0.1:49282]
[ERROR][hmda-akka.actor.default-dispatcher-31] Timeout in Joining!
[ERROR][hmda-akka.actor.default-dispatcher-31] Terminating system, because constructr-machine terminated!
```

Anyone else having these issues?

schbetsy:
Update: It seems that the error I pasted above happens when attempting to stop and re-start the API and connect to the same Zookeeper container. If I create a new zookeeper instance every time I restart the API, this works correctly.

PR #994, opened 2017-05-31
InstitutionView Serializer

### LINKED ISSUE #838: InstitutionView serializer

nickgrippin:
Closes #838

### DIFF
--- a/query/src/test/scala/hmda/query/serialization/InstitutionViewProtobufConverterSpec.scala
+++ b/query/src/test/scala/hmda/query/serialization/InstitutionViewProtobufConverterSpec.scala
@@ -1,0 +1,22 @@
+package hmda.query.serialization
+
+import hmda.model.institution.InstitutionGenerators._
+import hmda.persistence.model.serialization.InstitutionViewEvents.InstitutionViewStateMessage
+import hmda.query.view.institutions.InstitutionView.InstitutionViewState
+import hmda.query.serialization.InstitutionViewProtobufConverter._
+import org.scalacheck.Gen
+import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.prop.PropertyChecks
+
+class InstitutionViewProtobufConverterSpec extends PropSpec with PropertyChecks with MustMatchers {
+  val institutionList = Gen.listOf(institutionGen)
+
+  property("InstitutionViewState must serialize to protobuf and back") {
+    forAll(institutionList, Gen.choose(0l, 100000l)) { (i, seq) =>
+      val state = InstitutionViewState(i.toSet, seq)
+      val protobuf = institutionViewStateToProtobuf(state).toByteArray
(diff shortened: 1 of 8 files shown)


query/src/test/scala/hmda/query/serialization/InstitutionViewProtobufConverterSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+import hmda.query.serialization.InstitutionViewProtobufConverter._
+import org.scalacheck.Gen
+import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.prop.PropertyChecks
+
+class InstitutionViewProtobufConverterSpec extends PropSpec with PropertyChecks with MustMatchers {
+  val institutionList = Gen.listOf(institutionGen)
+
+  property("InstitutionViewState must serialize to protobuf and back") {
+    forAll(institutionList, Gen.choose(0l, 100000l)) { (i, seq) =>
```
jmarin: I think it reads clearer if we use capital L for `Long` literals. This change would make it consistent with the rest of the codebase too

query/src/test/scala/hmda/query/serialization/InstitutionViewProtobufSerializerSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.scalacheck.Gen
+import org.scalatest.{ MustMatchers, PropSpec }
+import org.scalatest.prop.PropertyChecks
+
+class InstitutionViewProtobufSerializerSpec extends PropSpec with PropertyChecks with MustMatchers {
+  val serializer = new InstitutionViewProtobufSerializer()
+
+  val institutionList = Gen.listOf(institutionGen)
+  property("InstitutionViewState must serialize to protobuf and back") {
+    forAll(institutionList, Gen.choose(0l, 100000l)) { (i, seq) =>
```
jmarin: Same comment as above about `Long` literal code style

query/src/test/scala/hmda/query/serialization/InstitutionViewProtobufSerializerSpec.scala
```diff
@@ -11,7 +11,7 @@ class InstitutionViewProtobufSerializerSpec extends PropSpec with PropertyChecks
 
   val institutionList = Gen.listOf(institutionGen)
   property("InstitutionViewState must serialize to protobuf and back") {
-    forAll(institutionList, Gen.choose(0l, 100000l)) { (i, seq) =>
+    forAll(institutionList, Gen.choose(0: Long, 100000: Long)) { (i, seq) =>
```
jmarin: This is fine, for for next time, `0L` and `100000L` would suffice. The problem with using lowercase is that it is harder to read as a letter if you are not using an IDE

Issue #979, opened 2017-05-21
IRS returns incorrect values in smoke tests


schbetsy:
In this screenshot, the IRS shows 92 total LARs for this submission, but the actual submission only has 10 LARs. I've seen this frequently when uploading multiple files for an institution.

<img width="648" alt="screen shot 2017-05-21 at 11 00 05 am" src="https://cloud.githubusercontent.com/assets/4295388/26284999/b4156072-3e14-11e7-9872-474ca3a8da21.png">

schbetsy:
Here's the steps for one way I have generated this bug. Notes: I was using institution `6999998` for all submissions. `HMDA_IS_DEMO` was set to `true`.

1. Clean slate in Postgres: in `psql hmda`, `drop table lars cascade;`, `\q`
1. Start the API in the usual way: In sbt, `clean`, `compile`, `project api`, `reStart` (repeat `reStop`-`reStart` cycle as necessary until the demo data gets populated 😞 )
2. Create submission. upload 90-line file. check irs... total lars = 90. ✅ 
3. Create another submission. upload 400-line file. check irs... total lars = 490. ❌ 
4. Just for fun, do it again: upload 10-line file. irs says total lines = 500. ❌ 

I've also seen it when running the whole docker setup.

nickgrippin:
Confirmed, the issue seems to be an intermittent failure to delete the previous LAR submissions from the repository, affecting the IRS report and modified LAR endpoint.

nickgrippin:
IRS reporting incorrect values will be addressed in #984

This ticket will stay open in order to address the bug with the modified LAR endpoint.

schbetsy:
This seems to be resolved by #1000.

nickgrippin:
I agree, that's what I'm seeing as well.  Closing this ticket.

PR #1000, opened 2017-06-07
Move IRS to Actor model

### LINKED ISSUE #984: Use actors to persist IRS data
Remove dependency on the LAR repository and use the `SubmissionLarStats` to persist the information for the IRS report.

nickgrippin:
Closes #984


nickgrippin:
Millennium PR!

💯0

nickgrippin:
I'm not seeing the #979 error occur on this branch.  Can someone else test/confirm this?

To clarify: when viewing the Postgres `lars` table, I can see the correct count of lars after every file upload event.

census/src/main/scala/hmda/census/model/Msa.scala
```diff
@@ -0,0 +1,93 @@
...
+      homeImprovement = homeImprovement + (lar.loan.purpose == 2),
+      refinance = refinance + (lar.loan.purpose == 3)
+    )
+  }
+}
+
+case class MsaMap(
+    msas: ListMap[String, Msa] = ListMap[String, Msa]()
+) {
+  def +(lar: LoanApplicationRegister): MsaMap = {
```
jmarin: How many LAR records do we usually see here? `ListMap` is a collection that is only suitable for small number of records. Do you need the extra properties of a `ListMap` or would an immutable `Map` be sufficient?

lesgou: For the IRS, since the rows are by MSA/MD (plus a total row), there should be no more than 1000 rows. Not sure if that is considered small.

nickgrippin: My original concern was related to ordering and pagination.  Before I switched to storing a `Seq` in the ValidationStats, I was worried about relying on the consistency of the iteration order through a `Map`.  `ListMap` guarantees a sorted order based on insertion, but that property is no longer necessary.  I'll make the switch back to an immutable `Map`.

schbetsy:
@nickgrippin I'll check on #979 and try to confirm that it's fixed.

PR #1004, opened 2017-06-09
IRS CSV endpoint

### LINKED ISSUE #950: IRS download
Depends on https://github.com/cfpb/hmda-platform-ui/issues/517

We'll need to add a link to download the CSV and have it make the appropriate API call.

nickgrippin:
Closes #950

### DIFF
--- a/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
+++ b/api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
@@ -64,8 +64,6 @@ trait SubmissionIrsPaths
   private def getMsa(instId: String, period: String, seqNr: Int)(implicit ec: ExecutionContext): Future[Seq[Msa]] = {
     val validationStats = system.actorSelection("/user/validation-stats")
     val submissionId = SubmissionId(instId, period, seqNr)
-    for {
-      data <- (validationStats ? FindIrsStats(submissionId)).mapTo[Seq[Msa]]
-    } yield data
+    (validationStats ? FindIrsStats(submissionId)).mapTo[Seq[Msa]]
   }
 }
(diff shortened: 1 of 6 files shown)


api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
```diff
@@ -49,4 +49,23 @@ trait SubmissionIrsPaths
         }
       }
     }
+
+  // institutions/<institutionId>/filings/<period>/submissions/<submissionId>/irs/csv
+  def submissionIrsCsvPath(institutionId: String)(implicit ec: ExecutionContext) =
+    path("filings" / Segment / "submissions" / IntNumber / "irs" / "csv") { (period, seqNr) =>
+      timedGet { uri =>
+        completeVerified(institutionId, period, seqNr, uri) {
+          val validationStats = system.actorSelection("/user/validation-stats")
```
schbetsy: Looks like these three lines are duplicated from the other IRS endpoint. I think we should extract these into a method.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionIrsPaths.scala
```diff
@@ -49,4 +45,27 @@ trait SubmissionIrsPaths
...
+            case Failure(e) => completeWithInternalError(uri, e)
+          }
+        }
+      }
+    }
+
+  private def getMsa(instId: String, period: String, seqNr: Int)(implicit ec: ExecutionContext): Future[Seq[Msa]] = {
+    val validationStats = system.actorSelection("/user/validation-stats")
+    val submissionId = SubmissionId(instId, period, seqNr)
+    for {
```
schbetsy: You can simplify this to 
```scala
  private def getMsa(instId: String, period: String, seqNr: Int)(implicit ec: ExecutionContext): Future[Seq[Msa]] = {
    val validationStats = system.actorSelection("/user/validation-stats")
    val submissionId = SubmissionId(instId, period, seqNr)
    (validationStats ? FindIrsStats(submissionId)).mapTo[Seq[Msa]]
  }
```

PR #1013, opened 2017-06-13
Institutions cassandra

### LINKED ISSUE #1012: Cassandra based institutions projection
Project institution data to Cassandra based backend

jmarin:
Closes #1012 
Doesn't remove `PostgreSQL` implementation yet


schbetsy:
Smoke tests for this PR look good, both locally and on Docker-Compose. I still need to take a closer look at the code, but I wanted to say 👍  so far.

I did notice a long `keycloak` stack trace in the docker-compose logs that started with the following line. I'm not sure it's relevant to this PR, but I wanted to call it out.
```
keycloak_1     | 14:23:36,222 ERROR [org.keycloak.services] (default task-14) KC-SERVICES0067: 
    failed to parse RestartLoginCookie: java.lang.RuntimeException: 
    java.lang.RuntimeException: java.security.InvalidKeyException: key is null
```

jmarin:
Thanks @schbetsy. I would be surprised if that error is related to this PR, maybe it requires a bit more investigation.

query/src/main/scala/hmda/query/repository/institutions/InstitutionCassandraRepository.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  override def dropTable(): ResultSet = {
+    val query = s"""
+      |DROP TABLE IF EXISTS $keyspace.institutions;
+    """.stripMargin
+
+    session.execute(query)
+  }
+
+  override def insertData(source: Source[InstitutionQuery, NotUsed]): Future[Done] = {
```
schbetsy: Looks like the `insertData` and `readData` methods are currently used in a spec (`InstitutionCassandraRepositorySpec`), but they aren't called explicitly from anywhere else in the code. Is this another case where they're called implicitly somewhere?

api/src/main/resources/application.conf
```diff
@@ -20,6 +20,12 @@ akka {
   cluster {
     failure-detector.threshold = 12 //Increase value for AWS environments
     metrics.enabled = off
+
+    retry-unsuccessful-join-after = 20s
+
+    //DON'T USE IN PRODUCTION!!!!!
```
jmarin: Auto downing of nodes is not recommended in production systems due to the possibility of network failures that form a partition in the cluster. It's useful in development so that the cluster lifecycle (joining, unreachable, leaving, etc. ) can be tested

query/src/main/resources/application.conf
```diff
@@ -31,3 +31,12 @@ db {
   }
 }
 
+cassandra {
+  host = "127.0.0.1"
+  host = ${CASSANDRA_CLUSTER_HOSTS}
+  port = 9142
```
jmarin: This is a mistake from a previous try. `CassandraUnit` uses 9142, but we're not using it when running the application. Will change back to default port

query/src/main/scala/hmda/query/repository/institutions/InstitutionCassandraRepository.scala
jmarin: Not right now. Both of those are convenience methods. Reading will happen when we introduce publication (down the road) and the write method is useful if we need to populate a panel file directly (i.e. for historical data)

query/src/main/scala/hmda/query/projections/institutions/InstitutionCassandraProjection.scala
```diff
@@ -0,0 +1,41 @@
...
+import hmda.persistence.processing.HmdaQuery._
+import hmda.query.model.institutions.InstitutionQuery
+import hmda.query.repository.institutions.InstitutionCassandraRepository
+import hmda.query.repository.institutions.InstitutionConverter._
+
+import scala.concurrent.ExecutionContext
+
+class InstitutionCassandraProjection extends InstitutionCassandraRepository {
+
+  def startUp(): Unit = {
```
jmarin: This starts up a stream of events from the journal, yes. The difference from previous implementations (i.e. what happens in validation) is that this stream is running continuously, listening to changes in the journal to push those events to the read model. One part that we haven't done yet (because it's not really 100% clear yet what the requirements are) is storing snapshots as an optimization, so not every event has to be read back to the read model.

api/src/test/resources/application.conf
```diff
@@ -39,3 +39,9 @@ db {
     keepAliveConnection = true
   }
 }
+
+cassandra {
+  host = "127.0.0.1"
+  port = 9142
```
schbetsy: Should this port also change from `9142` to `9042`?

query/src/test/resources/application.conf
```diff
@@ -25,3 +25,13 @@ hmda {
   isDemo = true
 }
 
+akka.persistence.journal.plugin = "inmemory-journal"
+akka.persistence.query.journal.id = "inmemory-read-journal"
+akka.persistence.snapshot-store.plugin = "inmemory-snapshot-store"
+
+cassandra {
+  host = "127.0.0.1"
+  port = 9142
```
schbetsy: Same question as above about port `9142`

jmarin: No, because `EmbeddedCassandraServerHelper` uses port `9142` for tests

api/src/main/resources/application.conf
schbetsy: Okay. Let's make sure there's an open issue in GH to remind us to update that value before we go to production.

schbetsy:
I've seen this in the Travis output during the specs, but it doesn't come up when I run the specs locally. 

```
[error] Error during tests:
[error]   Running java with options -classpath /home/travis/build/cfpb/hmda-platform/query/target/scala-2.12/test-classes:/home/travis/build/cfpb/hmda-platform/query/target/scala-2.12/classes:
           [...  I truncated this really long classpath]
           2.10.6/org.scala-sbt/sbt/0.13.13/test-interface-1.0.jar sbt.ForkMain 38993 failed with exit code 137
```
This is new as of this PR, right? Any idea where it might be coming from?

api/src/main/resources/application.conf
schbetsy: Created. #1017

schbetsy:
We'll merge this around midday tomorrow, even if Travis doesn't pass. Tests pass locally and smoke tests look good, so we're calling this safe to merge.

build.sbt
```diff
@@ -162,7 +162,8 @@ lazy val query = (project in file("query"))
         oldStrategy(x)
     },
     parallelExecution in Test := false,
-    libraryDependencies ++= configDeps ++ akkaPersistenceDeps ++ slickDeps
+    fork in Test := true,
```
schbetsy: It looks like this line is causing the `java.io.EOFException` that causes the "sbt.ForkMain failed with exit code 137" error in Travis.

Is forking necessary in the `query` project's specs? If so, why?

schbetsy: Source: I learned more about this error by reading [this GH issue from SBT](https://github.com/sbt/sbt/issues/653).

PR #1016, opened 2017-06-14
Panel refactor

### LINKED ISSUE #1011: Refactor panel loader
The current panel loader needs to be refactored / upgraded with the following requirements: 

* Rename to _loader_ as opposed to _parser_ to better reflect its function
* Make the project an independent package (jar file) so that it can be run independently
* Remove dependencies on any other project that has runtime implications (i.e. loader joining the `hmda` application cluster, when it shouldn't). This means that it should not depend on `api` or `query` for instance. This might require extraction of common code into separate projects, or code duplication of model objects

nickgrippin:
Closes #1011


schbetsy:
I used the instructions in the [Panel CSV Parser](https://github.com/cfpb/hmda-platform/blob/master/Documents/panel.md) documentation to smoke test this PR. 

Looks like some of it wants to be updated with this change. Some things I noticed:
- Zookeeper and Cassandra are required to run the `api` project
- The path to the example file isn't correct anymore. Where does it live now?

panel/src/main/scala/hmda/panel/PanelCsvLoader.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def main(args: Array[String]): Unit = {
+
+    if (args.length < 1) {
+      exitSys("No file argument provided", 1)
+    }
+
+    val file = new File(args(0))
+    if (!file.exists() || !file.isFile) {
+      exitSys("File does not exist", 2)
```
schbetsy: FYI (not sure if this is an issue, but) when I run the panel with a nonexistent file, I don't see "File does not exist" in the log. This is what I see:

```
> run wrong_filename.txt
[info] Running hmda.panel.PanelCsvLoader /Users/lortone/dev/hmda-project/hmda-platform-2/panel/src/main/resources/inst_datajjj
[info] [DEBUG] [06/22/2017 15:24:43.446] [main] [EventStream] StandardOutLogger started
[info] [DEBUG] [06/22/2017 15:24:43.666] [main] [EventStream(akka://hmda)] logger log1-Logging$DefaultLogger started
[info] [DEBUG] [06/22/2017 15:24:43.667] [main] [EventStream(akka://hmda)] Default Loggers started
java.lang.RuntimeException: Nonzero exit code returned from runner: 2
	at scala.sys.package$.error(package.scala:27)
[trace] Stack trace suppressed: run last panel/compile:run for the full output.
[error] (panel/compile:run) Nonzero exit code returned from runner: 2
[error] Total time: 2 s, completed Jun 22, 2017 3:24:44 PM
```

panel/src/main/scala/hmda/panel/PanelCsvLoader.scala
```diff
@@ -0,0 +1,101 @@
...
+            .via(Framing.delimiter(ByteString("\n"), maximumFrameLength = 1024, allowTruncation = true))
+            .drop(1)
+            .via(byte2StringFlow)
+            .via(stringToHttpFlow)
+            .via(connectionFlow)
+            .runWith(Sink.foreach[HttpResponse](elem => log.info(elem.entity.toString)))
+
+          completedF.onComplete(result => {
+            if (result.isSuccess)
+              sys.exit(0)
```
schbetsy: 🎉 Yay for exiting when it completes!

build.sbt
```diff
@@ -105,18 +105,23 @@ lazy val validation = (project in file("validation"))
...
+      mainClass in assembly := Some("hmda.panel.PanelCsvLoader"),
+      assemblyMergeStrategy in assembly := {
+        case "application.conf" => MergeStrategy.concat
+        case "META-INF/io.netty.versions.properties" => MergeStrategy.concat
+        case x =>
+          val oldStrategy = (assemblyMergeStrategy in assembly).value
+          oldStrategy(x)
+        },
+      libraryDependencies ++= httpDeps,
+      fork in run := true
```
schbetsy: Why is `fork in run` necessary?

nickgrippin: It forks the JVM in which the panel loader runs, allowing it to exit and return to the `sbt`, rather than exiting to the terminal.  This is only for usability reasons; it seemed like killing `sbt` wasn't good practice.

panel/src/main/scala/hmda/panel/PanelCsvLoader.scala
nickgrippin: I'm not entirely sure why the log doesn't print, but your logs do show the program exiting with the correct code (`2`).

build.sbt
```diff
@@ -105,18 +105,23 @@ lazy val validation = (project in file("validation"))
...
-    assemblyMergeStrategy in assembly := {
-      case "application.conf" => MergeStrategy.concat
-      case "META-INF/io.netty.versions.properties" => MergeStrategy.concat
-      case x =>
-        val oldStrategy = (assemblyMergeStrategy in assembly).value
-        oldStrategy(x)
-    },
-    libraryDependencies ++= akkaPersistenceDeps ++ httpDeps
+    Seq(
+      assemblyJarName in assembly := {s"${name.value}.jar"},
```
schbetsy: What does the .jar name resolve to, and where will it be once it gets generated?

I ran `sbt clean assembly` in the project, then looked in `target/scala-2.12/` for the compiled .jar files. I only saw `hmda.jar`. I'm not sure if that means the panel loader .jar will live in a different target directory, or if it didn't get created correctly.

build.sbt
schbetsy: Ooh, good call.

panel/src/main/scala/hmda/panel/PanelCsvLoader.scala
schbetsy: Works for me ¯\_(ツ)_/¯

build.sbt
nickgrippin: Are you sure you were in the panel project when running `assembly`?  This is what I saw:
```
[info] Packaging /Users/grippinn/HMDA/hmda-platform/panel/target/scala-2.12/panel.jar ...
[info] Done packaging.
[success] Total time: 122 s, completed Jun 22, 2017 3:50:32 PM
```

schbetsy: No, I ran `sbt clean assembly`, like I said. Maybe add that it has to be packaged separately to the documentation?

schbetsy: You're right, though. This is what worked for me:
```
$ sbt
> project panel
> assembly
```
... which saves the .jar to `panel/target/scala-2.12/panel.jar`.

Then, with the API (and Cassandra and Zookeeper) running, this command runs the loader successfully:

```
java -jar  panel/target/scala-2.12/panel.jar [path-to-project]/panel/src/main/resources/inst_data_2017_dummy.csv
```

schbetsy:
One last documentation update, then LGTM

PR #1018, opened 2017-06-21
Lar cassandra

### LINKED ISSUE #1014: Cassandra based filing projection
Project `LAR` data to Cassandra based backend

jmarin:
Closes #1014

### DIFF
--- a/query/src/main/scala/hmda/query/projections/filing/FilingCassandraProjection.scala
+++ b/query/src/main/scala/hmda/query/projections/filing/FilingCassandraProjection.scala
@@ -21,7 +21,10 @@ class FilingCassandraProjection extends FilingCassandraRepository {
     }
     val sink = CassandraSink[LoanApplicationRegisterQuery](parallelism = 2, preparedStatement, statementBinder)
     source
-      .map(lar => toLoanApplicationRegisterQuery(lar))
+      .map { lar =>
+        val queryLar = toLoanApplicationRegisterQuery(lar)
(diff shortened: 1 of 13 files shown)


query/src/main/scala/hmda/query/projections/filing/FilingCassandraProjection.scala
```diff
@@ -0,0 +1,31 @@
...
+import hmda.query.model.filing.LoanApplicationRegisterQuery
+import hmda.query.repository.filing.LarConverter._
+
+class FilingCassandraProjection extends FilingCassandraRepository {
+
+  def startUp(): Unit = {
+    createKeyspace()
+    createTable()
+
+    val source: Source[LoanApplicationRegister, NotUsed] = liveEvents("HmdaFiling-2017").map {
```
nickgrippin: I'm seeing the 2017 year a lot throughout this PR,  are we OK with having this value hard-coded?

jmarin: `2017` in general refers to data model version, not year. In this case, we are only going to project data from 2017, which is what we are collecting. Past data will follow the same "2017" data model, but will be inserted differently (through the `insertData` method in the LAR Cassandra repository object)

schbetsy: That makes sense about the data model versioning.

Though there are several instances of `$keyspace.lars2017`.  Would it be prudent to tokenize the table name `lars2017` so we can't introduce typos?

PR #1022, opened 2017-06-23
Update IRS CSV formatting

### LINKED ISSUE #1021: IRS CSV
The MSA/MD name can have a `,` in it, eg "Terre Haute, IN". This causes the downloaded CSVs columns to be misaligned.

## Example file
[2-irs-report.txt](https://github.com/cfpb/hmda-platform/files/1098432/2-irs-report.txt) (update the extension to .csv)

schbetsy:
Closes #1021 

According to [this StackOverflow](https://stackoverflow.com/questions/12473480/how-should-i-escape-commas-and-speech-marks-in-csv-files-so-they-work-in-excel), the values with commas need to be surrounded by quotation marks AND not have a leading space.

So, 
```
this works:
"12345,"Indianapolis,IN", 1, 2, 3"

but this doesn't:
"12345, "Indianapolis, IN", 1, 2, 3"
       ^ ... because of this space
```

### DIFF
--- a/Documents/API.md
+++ b/Documents/API.md
@@ -627,8 +627,8 @@ Example response:
 Example response:
 ```csv
 MSA/MD, MSA/MD Name, Total LARs, Total Amt. (in thousands), CONV, FHA, VA, FSA/RHS, 1-4 Family, MFD, Multi-Family, Home Purchase, Home Improvement, Refinance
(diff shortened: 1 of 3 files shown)


nickgrippin:
The API doc needs to be updated, otherwise this LGTM

PR #1033, opened 2017-07-10
Add income per MSA

### LINKED ISSUE #1032: Add income per MSA

jmarin:
Closes #1032

**Note**: this data will have to be updated shortly, see #1034


schbetsy:
How will this be used? (i.e. Why are we adding this?)

nickgrippin:
@schbetsy: this is used in a bunch of the A&D reports

PR #1037, opened 2017-07-11
Update panel docs and error messages

### LINKED ISSUE #1036: Update Panel Loader Documentation

nickgrippin:
Closes #1036

### DIFF
--- a/Documents/panel.md
+++ b/Documents/panel.md
@@ -8,7 +8,7 @@ There are two environment variables used by the panel loader.  Both must be set 
 
 For testing locally, no changes need to be made.  The defaults for both of these variables will point to the correct local admin API.
 
-For loading panel data into the cluster, navigate to a running `hmda-platform` service in DC/OS.  Under the "Endpoints" section you'll find the IP address and three ports.  As of now, the middle port will always be the admin API.  In your terminal, set both of the environment variables to the DC/OS values.
+For loading panel data into a remote system, you'll need to set the following environment variables:
 ```shell
 > export HMDA_HTTP_ADMIN_HOST={ip address}
 > export HMDA_HTTP_ADMIN_PORT={port #}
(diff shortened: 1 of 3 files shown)


Documents/panel.md
```diff
@@ -1,17 +1,29 @@
...
 ## Overview
-The panel parser is designed to read a CSV file and load the data onto the HMDA-Platform.  The CSV file should use the `|` (pipe) delimiter, and should include a header row as the first line.
+The panel loader is designed to read a CSV file and load the data onto the HMDA-Platform.  The CSV file should use the `|` (pipe) delimiter, and should include a header row as the first line.
+
+## Environment Variables
+There are two environment variables used by the panel loader.  Both must be set correctly in order for the data to be sent to the admin API.
+
+For testing locally, no changes need to be made.  The defaults for both of these variables will point to the correct local admin API.
+
+For loading panel data into the cluster, navigate to a running `hmda-platform` service in DC/OS.  Under the "Endpoints" section you'll find the IP address and three ports.  As of now, the middle port will always be the admin API.  In your terminal, set both of the environment variables to the DC/OS values.
```
schbetsy: I'm trying to follow along with these instructions, but I can't find what you mean here. I don't see the "endpoints" section. 

Also, why will the middle port always be the admin API? What's the likelihood that that will change?

Documents/panel.md
```diff
@@ -21,6 +33,14 @@ sbt> clean
 sbt> assembly
 ```
 Then the panel loader can be run with `java -jar  panel/target/scala-2.12/panel.jar path/to/institution_file.csv`
+
+## Error codes
+There are four ways the panel loader can fail.  The exit code and error message should tell you what happened.
+1) There were no command line arguments passed to the loader
+2) The path passed to the loader didn't point to a file
+3) The call to `institutions/create` didn't return the correct response.  This can indicate that you don't have the correct environment variables set, or that something is wrong with the hmda-platform.
+4) The loader didn't finish processing all the institutions.  This will happen when running the real panel file, but unsure as to why this happens.
```
schbetsy: Please create a GH ticket for this bug, if there isn't one.

nickgrippin: I'm still trying to figure out where the bug is (so I can create the ticket in the appropriate repo)

Documents/panel.md
nickgrippin: The ports are declared in the `application.conf` file for the api project.  I don't know how likely this is to change, but as of now, the middle port (8081, when running locally) is the admin api.

When you navigate to a running service in DC/OS, you'll find a details page that lists configuration settings for that service.  One section is called "endpoints".

schbetsy: ah, okay. I was looking at the parent page. I see the "endpoints" section now. Looks fine to me.

Documents/panel.md
schbetsy: When you do create that ticket, include a reminder to update this documentation. Once the bug is fixed, we don't want to have this "it's broken and we don't know why" message in here.

Documents/panel.md
hkeeler: I don't think we should be putting DC/OS-specific bits here in the public repo.  If there are tips/tricks to getting it running there, those should probably go in our internal repo's docs.  I think this could just refer to a _remote system_.  Something like:

>When loading panel data into a remote instance of the Admin API, you'll need to set the following environment variables.

Documents/panel.md
```diff
@@ -21,6 +33,14 @@ sbt> clean
 sbt> assembly
 ```
 Then the panel loader can be run with `java -jar  panel/target/scala-2.12/panel.jar path/to/institution_file.csv`
+
+## Error codes
+There are four ways the panel loader can fail.  The exit code and error message should tell you what happened.
+1) There were no command line arguments passed to the loader
```
hkeeler: Could you use standard Markdown numbering? :)

hkeeler:
Also, just to confirm, is everything loader-related still accurate in https://github.com/cfpb/hmda-platform#interactive?

nickgrippin:
@hkeeler: looks right, I just need to update the file path.

PR #1038, opened 2017-07-11
Use separate EDITS_DEMO_MODE env var to control S025

### LINKED ISSUE #1035: Toggle S025 and Demo Data load with separate environment variables
Currently, the `HMDA_IS_DEMO` environment variable has two uses. However, we want to control its two functions separately, so they should be changed to use specific environment variables.

1. `S025` checks the the `HMDA_IS_DEMO` environment variable to see whether it should run. If `true`, the check always passes. If `false`, the check runs as usual. Update `S025` to use a separate, specifically named env var instead. Maybe `S025_DEMO_MODE`?
2. The `HMDA_IS_DEMO` var also controls whether the system loads demo data (vs production data) into the db. Consider renaming that var to be more descriptive of how it's actually used, e.g. `LOAD_DEMO_DATA` or similar.

schbetsy:
Closes #1035

### DIFF
--- a/validation/src/main/resources/application.conf
+++ b/validation/src/main/resources/application.conf
@@ -10,6 +10,8 @@ hmda {
     timeout = 5
   }
   persistent-actor-timeout = 60
+  edits.demoMode = false
+  edits.demoMode = ${?EDITS_DEMO_MODE}
 }
 
 hmda.validation.quality {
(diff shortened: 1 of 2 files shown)


schbetsy:
This will mean that any demo/testing environments will need to include the `EDITS_DEMO_MODE=true` environment variable, to ensure that S025 won't block testing.
@hkeeler Should I create a devops ticket for including that env var?

hkeeler:
>@hkeeler Should I create a devops ticket for including that env var?

Yes, please. :)

PR #1050, opened 2017-07-14
Env variables default to prod values

### LINKED ISSUE #1042: Make env vars default to Production values
... then document their values and create a ticket in the devops repo

schbetsy:
Closes #1042 

~~Don't merge this yet. Needs testing in multiple environments first.~~


schbetsy:
@hkeeler Does this look reasonable to you? If so, should we test it on the cluster? (at a strategic time, so it can't potentially disrupt user testing)

PR #1087, opened 2017-08-04
Update README w/ latest Compose details

### LINKED ISSUE #1065: Update README with URLs for all services
The URLs changed a bit with #1049.  Let's update the docs to reflect those changes.

hkeeler:
Closes #1065

I also took the opportunity to update the Compose section as a whole.

If you want to see the rendered version, head over to:
* https://github.com/hkeeler/hmda-platform/tree/readme-compose-lb#to-run-the-entire-platform

~~Marking this as _WIP_ as I forgot https://github.com/cfpb/hmda-platform/issues/1065#issuecomment-318171245 was part of the requested change.  I'll have that up shortly as well.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,34 @@ The Public API will run on `$(docker-machine ip):8082`
 
 #### To run the entire platform
 
+1. Dedicate appropriate resources to your Docker environment.  We've found
+    that for the full stack to run efficiently, you need approximately:
+
+    * 4 CPUs
+    * 6 GB RAM
+    * 80 GB Disk space
+
+    Assuming you are using Docker Machine to provision your Docker
+    environment, you can check you current settings with (ignore the 
+    second `Memory`):
+
+        $ docker-machine inspect | grep 'CPU\|Memory\|DiskSize'
+            "CPU": 4,
+            "Memory": 6144,
+            "DiskSize": 81920,
+            "Memory": 0,
+
+    If your settings are _below_ these suggestions, you should create a new
+    Docker VM. The following will create a _machine_ named `hmda-platform`
+    with the appropriate resources:
+
+        $ docker-machine create \
+        --driver virtualbox \
+        --virtualbox-disk-size 81920 \
+        --virtualbox-cpu-count 4 \
+        --virtualbox-memory 6144 \
+        hmda-platform
+
 1. Clone [hmda-platform-ui](https://github.com/cfpb/hmda-platform-ui) and 
     [hmda-platform-auth](https://github.com/cfpb/hmda-platform-auth) into the same
     directory as hmda-platform.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -208,51 +208,96 @@ The Filing API will run on `$(docker-machine ip):8080`
...
-docker-machine ip dev
-```
+1. Build hmda-platform
+
+        cd hmda-platform && \
+        sbt clean assembly
+
+1. Launch the stack with Docker Compose
+
+        docker-compose up
```
schbetsy: So this doesn't need the `--build` flag because you're assuming the reader is doing this for the first time? (I'm not clear on all of the docker commands)

README.md
```diff
@@ -208,51 +208,96 @@ The Filing API will run on `$(docker-machine ip):8080`
...
-- https://192.168.99.100:8443/
-- https://192.168.99.100:4443/
-- https://192.168.99.100:9443/
+If you've updated any of the hmda-platform services, and would like to see those 
+changes reflected in the Docker Compose setup, the simplest way to do this is to
+rebuild everything from scratch.  The following command should be executed from
+within the `hmda-platform` directory.
 
-Visit the app at http://192.168.99.100, click the "Login" button, and click "Register" when redirected to the keycloak login screen.
+    docker-compose stop -t0 && \
```
schbetsy: If they stop the docker-compose setup with `docker-compose down`, then these steps (`stop` and `rm -vf`) aren't necessary. Should we include that anywhere?

hkeeler: Good question.  I don't ever use `docker-compose down`.  I'll look into it, and if they are indeed equivalent, I'm all for changing it.

README.md
hkeeler: Correct.  When you `up` for the first time, or really start any service that doesn't have a cached image, it has to do the equivalent of `--build`.

README.md
```diff
@@ -208,51 +208,96 @@ The Filing API will run on `$(docker-machine ip):8080`
...
+        cd hmda-platform && \
+        sbt clean assembly
+
+1. Launch the stack with Docker Compose
+
+        docker-compose up
+
+    This will bring up all the HMDA Platform services. The first run may take several minutes.
+
+1. Discover you Docker host's IP
```
schbetsy: Typo: "your"

README.md
hkeeler: I haven't had time to experiment with this.  The only thing I can _see_ that is different is the lack of the `stop -t0`, which stops the containers without waiting for them to stop gracefully.  Personally, I'm impatient, so I like this, but I also like the simplicity of `docker-compose down`.

I say let's keep the instructions as-is for now, and I will play with this a bit in the future, and update this doc if it works out like we think it will.

PR #1088, opened 2017-08-07
Configure publication

### LINKED ISSUE #1067: Separate publication .jar?
Should the `publication` project be aggregated into the hmda.jar file or packaged into it's own .jar?

_From Nick's comment here: https://github.com/cfpb/hmda-platform/pull/1064#discussion_r129877947_

jmarin:
Closes #1067

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,8 @@ lazy val hmda = (project in file("."))
     query,
     platformTest,
     validation,
-    census)
+    census,
+    publication)
 
 lazy val model = (crossProject in file("model"))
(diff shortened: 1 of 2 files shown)


schbetsy:
Okay, looks like the answer to Nick's original question is: We won't be packaging `publication` as a separate `.jar`. This will be part of the normal platform deploy.

If so, LGTM.

jmarin:
Yes. The `HMDA Platform` is entirely packaged as a single `JAR` file. The only exception is the panel loader, which is currently run from a desktop environment.

PR #1089, opened 2017-08-07
Dependencies docker

jmarin:
Configures the necessary dependencies for development of the HMDA Platform, so that `Cassandra`, `Zookeeper` and `PostgreSQL` can be brought up with one `docker-compose` command. Provides defaults in configuration files so that there is no need to configure environment variables for development.

### DIFF
--- a/query/src/main/resources/application.conf
+++ b/query/src/main/resources/application.conf
@@ -49,7 +49,7 @@ db {
   driver = "slick.driver.PostgresDriver$"
 
   db {
-    url = "jdbc:postgresql://localhost/hmda?user=postgres&password=postgres"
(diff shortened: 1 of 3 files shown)


query/src/main/resources/application.conf
```diff
@@ -49,7 +49,7 @@ db {
   driver = "slick.driver.PostgresDriver$"
 
   db {
-    url = "jdbc:postgresql://localhost/hmda?user=postgres&password=postgres"
+    url = "jdbc:postgresql://192.168.99.100/hmda?user=postgres&password=postgres"
```
schbetsy: For this to work for me, I needed this to have the port specified, like this:
```
url = "jdbc:postgresql://192.168.99.100:54321/hmda?user=postgres&password=postgres"
```

PR #1091, opened 2017-08-07
Dispatchers

### LINKED ISSUE #1090: Implement bulkhead pattern with dispatchers

jmarin:
Closes #1090

### DIFF
--- a/api/src/main/scala/hmda/api/HmdaPlatform.scala
+++ b/api/src/main/scala/hmda/api/HmdaPlatform.scala
@@ -34,7 +34,7 @@ object HmdaPlatform {
     val system = ActorSystem(configuration.getString("clustering.name"), configuration)
     val supervisor = createSupervisor(system)
     val querySupervisor = createQuerySupervisor(system)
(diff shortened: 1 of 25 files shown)


api/src/main/scala/hmda/api/HmdaPlatform.scala
```diff
@@ -34,7 +34,7 @@ object HmdaPlatform {
     val system = ActorSystem(configuration.getString("clustering.name"), configuration)
     val supervisor = createSupervisor(system)
     val querySupervisor = createQuerySupervisor(system)
-    implicit val ec = system.dispatcher
+    implicit val ec = system.dispatchers.lookup("api-dispatcher")
```
schbetsy: (For anyone who's googling around like me, trying to understand this code better: )  [This documentation on dispatchers](http://doc.akka.io/docs/akka/current/scala/dispatchers.html) is a good resource. If I'm reading this correctly, this line ensures that any actor that is started with this `ec` in scope will have the `api-dispatcher` as its dispatcher.

PR #1092, opened 2017-08-07
Fix Q061

### LINKED ISSUE #1076: Q061: incorrect field reference
Q061 is coded to check the proportion of HOEPA status but should instead be checking property type.

nickgrippin:
Closes #1076

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/macro/Q061.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/macro/Q061.scala
@@ -22,7 +22,7 @@ object Q061 extends AggregateEditCheck[LoanApplicationRegisterSource, LoanApplic
 
     val firstLienPropertyLoans =
       count(lars.filter(lar => lar.loan.propertyType == 1 && lar.actionTakenType == 1 && lar.lienStatus == 1 && lar.rateSpread != "NA")
-        .filter(lar => lar.rateSpread.toInt >= 5))
+        .filter(lar => lar.rateSpread.toDouble > 5.0))
 
     val total = count(lars.filter(lar => lar.actionTakenType == 1))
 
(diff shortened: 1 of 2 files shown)


validation/src/main/scala/hmda/validation/rules/lar/macro/Q061.scala
```diff
@@ -20,14 +20,14 @@ object Q061 extends AggregateEditCheck[LoanApplicationRegisterSource, LoanApplic
 
   override def apply[as: AS, mat: MAT, ec: EC](lars: LoanApplicationRegisterSource): Future[Result] = {
 
-    val firstLienHoepaLoans =
-      count(lars.filter(lar => lar.hoepaStatus == 1 && lar.actionTakenType == 1 && lar.lienStatus == 1 && lar.rateSpread != "NA")
+    val firstLienPropertyLoans =
+      count(lars.filter(lar => lar.loan.propertyType == 1 && lar.actionTakenType == 1 && lar.lienStatus == 1 && lar.rateSpread != "NA")
         .filter(lar => lar.rateSpread.toInt >= 5))
```
schbetsy: (I know this isn't the stated issue, but it's worth looking into while we're here) Should this rateSpread comparison really use `toInt`? Will that cause rounding issues?

nickgrippin:
@schbetsy: Good catch, definitely could cause problems.  Also noticed that it should be `>`, not `>=`.  Done in 117f80e903e2a5a7768920be04006f5b601bd9b8

schbetsy:
I submitted the Q061 test file (in the shared drive at HMDA > Platform > HMDA Edits > Validation Testing > Test Files > Q061.txt) to this updated branch and `Q061` did fail. Is that the desired outcome, @Kibrael?

schbetsy:
@nickgrippin I saw this test failure in Travis:
```
[info] Q061Spec:
[info] - be invalid if first lien property type loans > 0.01 * total *** FAILED ***
[info]   Success() was not an instance of hmda.validation.dsl.Failure, but an instance of hmda.validation.dsl.Success (SummaryEditCheckSpec.scala:26)
```

Maybe there's an edge case not accounted for in the specs?

nickgrippin:
@schbetsy: It wasn't an edge case, I forgot to update the tests with the new "greater than" logic.  Should be fixed now.

schbetsy:
Excellent. And @Kibrael confirmed that the test file should indeed fail, so this is showing the correct behavior now. LGTM, once Travis is green.

Kibrael:
+1

PR #1098, opened 2017-08-09
Q016 - field reference fix

### LINKED ISSUE #1077: Q016: incorrect field reference
Q016 checks that income should be < 10k for <= 20% of total applications.

Implementation is currently checking loan amount instead of applicant income.

nickgrippin:
Closes #1077

### DIFF
--- a/validation/src/test/scala/hmda/validation/rules/lar/macro/Q016Spec.scala
+++ b/validation/src/test/scala/hmda/validation/rules/lar/macro/Q016Spec.scala
@@ -17,7 +17,7 @@ class Q016Spec extends MacroSpec {
     lar.copy(applicant = underCapApplicant)
   }
   def overCap(lar: LoanApplicationRegister) = {
-    val overCapApplicant = lar.applicant.copy(income = (incomeCap + 1).toString)
+    val overCapApplicant = lar.applicant.copy(income = incomeCap.toString)
     lar.copy(applicant = overCapApplicant)
   }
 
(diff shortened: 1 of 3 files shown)


validation/src/test/scala/hmda/validation/rules/lar/macro/Q016Spec.scala
```diff
@@ -13,12 +13,17 @@ class Q016Spec extends MacroSpec {
...
   def underCap(lar: LoanApplicationRegister) = {
-    val underCapLoan = lar.loan.copy(amount = incomeCap - 1)
-    lar.copy(loan = underCapLoan)
+    val underCapApplicant = lar.applicant.copy(income = (incomeCap - 1).toString)
+    lar.copy(applicant = underCapApplicant)
   }
   def overCap(lar: LoanApplicationRegister) = {
-    val overCapLoan = lar.loan.copy(amount = incomeCap + 1)
-    lar.copy(loan = overCapLoan)
+    val overCapApplicant = lar.applicant.copy(income = (incomeCap + 1).toString)
```
schbetsy: We should use `incomeCap` here, not `incomeCap + 1`, so that the spec covers the edge case.

validation/src/test/scala/hmda/validation/rules/lar/macro/Q016Spec.scala
```diff
@@ -13,12 +13,17 @@ class Q016Spec extends MacroSpec {
...
+    lar.copy(applicant = overCapApplicant)
+  }
+
+  def naIncome(lar: LoanApplicationRegister) = {
+    val naApplicant = lar.applicant.copy(income = "NA")
+    lar.copy(applicant = naApplicant)
   }
 
   property(s"be valid if under cap loans < $multiplier * over cap loans") {
```
schbetsy: These property names aren't quite accurate. They all say `$multiplier * over cap loans`, but `$multiplier * total loans` is more truthful.

schbetsy:
I confirmed that the test file for Q016 (in the shared drive at HMDA > Platform > HMDA Edits > Validation Testing > Test Files > Quality (Macro) > Q016.txt) now does fail `Q016`, which is the desired outcome. 👍

PR #1104, opened 2017-08-14
Update README

### LINKED ISSUE #1105: Improve README with new dev dependencies strategy
Now that #464 is merged, the section of the README about running the dev dependencies can be much simpler. Only include the docker-compose strategy in the readme; move background info and alternative ways of running dependencies into a separate document.

Take this opportunity to make any other cleanup edits necessary for clarity in the documentation.

schbetsy:
The project setup has updated and the README (and other documentation) was no longer very clear.

Closes #1105


nickgrippin:
Now that we have specific memory instructions, maybe we could include those under [this section](https://github.com/schbetsy/hmda-platform/blob/e1f57e1e35445616cb9fd1a42ad6c19a0da65c76/README.md#docker) of the README, where we give instructions on creating a docker machine.

schbetsy:
Good idea @nickgrippin. Your comment helped me notice that we had two different sets of instructions for how to create a Docker Machine, so I consolidated them.

Documents/local-dependencies.md
```diff
@@ -0,0 +1,99 @@
...
+docker run -it --link cassandra:cassandra --rm cassandra cqlsh cassandra
+```
+
+Once the `Cassandra` server is running, set the following environment variable to the appropriate Cassandra host (in this example, the default local docker host for a machine running MacOs X):
+
+```shell
+export CASSANDRA_CLUSTER_HOSTS=192.168.99.100
+```
+
+To load data into `Cassandra`, you can run the following (the Cassandra server needs to be running and correct environment variables configured as per the previous instructions):
```
nickgrippin: I would just link the panel docs for the instructions here.  If a user is following this doc, they'll miss the environment variable they need to set for the panel loader to run properly.

schbetsy: Yeah.... I was trying to avoid having to rewrite all of the Cassandra and panel-related documentation (by just using the documentation that was already written), but I guess it just won't work. The Cassandra section in the "local-dependencies" file and the "panel" file just don't make sense together 😕 . 

I'll rewrite those and update this PR.

PR #1109, opened 2017-08-16
Cluster roles

### LINKED ISSUE #841: Implement cluster roles
Add cluster roles so that a node spinning up the application starts up the necessary components associated with defined role (i.e. filing api, public api, seed node, etc.)

jmarin:
Closes #841


schbetsy:
Please update the README (specifically the "[Running the API](https://github.com/cfpb/hmda-platform/#running-the-api)" section) with the new instructions for running the project locally. Maybe it just needs to change to `project cluster`, but you can check for any other changes we need.

persistence/src/test/scala/hmda/persistence/processing/HmdaFileValidatorSpec.scala
```diff
@@ -104,20 +107,13 @@ class HmdaFileValidatorSpec extends ActorSpec with BeforeAndAfterEach with HmdaF
       probe.expectMsg(HmdaFileValidationState(
         Some(ts),
         lars,
-        List(
-          SyntacticalValidationError("38800009923", "S025", true),
-          SyntacticalValidationError("38800009923", "S011", true)
-        ),
+        List(),
```
nickgrippin: The LARs being sent to the `HmdaFileValidator` haven't changed, correct?  Why are there fewer errors being returned in this test?

schbetsy:
Quick update on smoke testing: The app works well running locally--I'm able to perform all of the basic actions (upload, validate, and sign a HMDA file) 🎉 . I've had success when the cluster roles are divided between different sessions (I've tried `api`, `persistence`, and `publication` in one session, with `query` in another).

We're still having an issue with persistent actor timeouts. After some persistent actors time out, the validation process no longer fully works. In debug mode, I see this message repeatedly:
```
cluster [2017-08-31 12:30:00,796][DEBUG][hmda-akka.actor.default-dispatcher-48] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,147][DEBUG][hmda-akka.actor.default-dispatcher-135] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,756][DEBUG][hmda-akka.actor.default-dispatcher-48] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,786][DEBUG][hmda-akka.actor.default-dispatcher-133] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,786][DEBUG][hmda-akka.actor.default-dispatcher-133] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,796][DEBUG][hmda-akka.actor.default-dispatcher-134] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
cluster [2017-08-31 12:30:01,796][DEBUG][hmda-akka.actor.default-dispatcher-134] Trying to identify singleton at [akka.tcp://hmda@127.0.0.1:56423/user/validation-stats/singleton]
```

jmarin:
This actor probably doesn't need to have the "stop after a while" logic, especially since it's a singleton. It does by virtue of being a `HmdaPersistentActor`, but I'm going to remove that logic from its `preStart` function. This should solve the validation timeouts that we have been seeing in the application (this problem is not specific to clustering; but I'm happy to fix it here)

persistence/src/test/scala/hmda/persistence/processing/HmdaFileValidatorSpec.scala
jmarin: Right. When I introduced this PR I mentioned that I had to change a test to make it work. This is it. I'm not sure why this change happens, though I think the API is doing the right thing when running the project.

nickgrippin:
I don't know how much of a priority this is, but I've been having issues getting demo data to load.  Since demo data loads with the `query` role, there is some dependency on the role starting order. 
 I think when all of the roles start at once, the data doesn't always load properly.  Is this something that should be fixed, or is it not a big concern?

jmarin:
I guess it could be a problem if the `query` role is started before `persistence`. Is this happening when running it on separate instances, or also when running the whole project together? This is not an issue in production

schbetsy:
I can confirm: demo data loads successfully when running all the roles in the same session. I only saw demo data fail when it was separated into different sessions.

schbetsy:
Okay, I guess I was wrong about the `demoMode` config. LGTM, assuming Travis starts looking green.

Issue #1071, opened 2017-08-01
Test A&D reports actors


schbetsy:
- [ ] `AggregateAndDisclosureReports.scala`
- [ ] `DisclosureReports.scala`

... others?

nickgrippin:
Also should include tests for all the Util objects

PR #1111, opened 2017-08-18
Test race, ethnicity, and minority status utilities

schbetsy:
Addresses #1071

Something important I found out while writing these specs: `AsyncWordSpec`, which we frequently use for testing the results of `Future`s, can only handle ONE `Future` per spec. If it has more than one `Future.map` with assertions in the `map` statement, all but the last one will be ignored.

I've updated all of the specs in this PR to only have one `.map` per spec, but other places in the code still need to be updated.


nickgrippin:
This includes a lot of changes to the logic of the utils... Are we sure these are the correct changes?

schbetsy:
I discovered lots of errors in the filtering logic by writing the tests. The tests are designed to follow the [ethnicity](https://github.com/cfpb/hmda-platform/wiki/Ethnicity-Categorization), [race](https://github.com/cfpb/hmda-platform/wiki/Race-Categorization), and [minority status](https://github.com/cfpb/hmda-platform/wiki/Minority-Status-Categorization) categorization wikis. So... yes, I'm confident these are much closer to being correct now.

But you don't have to take my word for it. You could check the logic too.

publication/src/main/scala/hmda/publication/reports/util/RaceUtil.scala
```diff
@@ -49,7 +49,7 @@ object RaceUtil {
 
       case Joint =>
         larSource.filter { lar =>
-          (applicantTwoOrMoreMinorities(lar) || coApplicantTwoOrMoreMinorities(lar)) &&
+          (applicantOneOrMoreMinorities(lar) || coApplicantOneOrMoreMinorities(lar)) &&
```
nickgrippin: I'm not sure if this follows the wiki exactly... the `applicantOneOrMoreMinorities` should only check for the race being between 1 and 4 (and is that in any race field, not just `race1`?).

publication/src/main/scala/hmda/publication/reports/util/RaceUtil.scala
```diff
@@ -49,7 +49,7 @@ object RaceUtil {
...
             (lar.race1 == 5 || coApplicantWhite(lar))
```
nickgrippin: Shouldn't the `lar.race1 == 5` also make sure the other fields are blank, similar to the `coApplicantWhite` function?

publication/src/main/scala/hmda/publication/reports/util/RaceUtil.scala
schbetsy: For `race2` through `race5`, the only options are 1, 2, 3, 4, and 5. So for those fields, checking that they equal 1,2,3 or 4 is the same as checking `!= 5`

publication/src/main/scala/hmda/publication/reports/util/RaceUtil.scala
schbetsy: Ah, you're probably right. That sentence in the wiki is ambiguous 
```
Applicant Race: 1 or Co-applicant Race: 1 equals White (5) and all else are blank
```
but I think your interpretation is more likely to be the correct one. I'll update this.

PR #1113, opened 2017-08-21
Add authentication for cassandra journal and snapshot stores

jmarin:


persistence-model/src/main/resources/application.conf
```diff
@@ -77,6 +77,10 @@ cassandra-journal {
   connect-retries = 10
   connect-retry-delay = 5s
   reconnect-max-delay = 60s
+  authentication.username = ""
```
schbetsy: The code never appears to refer to these `cassandra-journal.authentication.username` and `cassandra-snapshot-store.authentication.username`. Should it?

(I see that the code _does_ refer to the settings in the query project's `application.conf`, so I'm not worried about those)

schbetsy:
As we discussed in the meeting, this change should merge into the `1.11.x` branch. Do you want to change the base branch for this PR?

kgudel:
Other than the comments by @schbetsy LGTM

persistence-model/src/main/resources/application.conf
jmarin: The journal and snapshot implementations take default configuration from [here](https://github.com/akka/akka-persistence-cassandra/blob/master/core/src/main/resources/reference.conf). The `Akka` code reads these properties and uses them, all I've done here is override them. This is how we have been able to use the host, port, etc. all along. The `CassandraRepository` implementation uses the `Java` driver directly and that is why the implementation is different.

jmarin:
These changes should go into `master`. Once that is done, I will cut the 1.11.1 release

PR #1118, opened 2017-08-25
Fix for Filing Status Bugs

### LINKED ISSUE #1115: Filing status resets to 'in-progress' after second resubmission
Once a filing has been signed, the filing status should *always* remain as
`{ code: 3, message: 'completed' }`

This status persists for **one** resubmission (`POST`ing to `/hmda/institutions/6999999/filings/2017/submissions`), but after the second resubmission, reverts to an in-progress status.

Steps to repro:

 - File and sign a submission
 - Go to the upload page and select a new file to upload
 - Go to the institutions page and note the warning about resubmitting a signed filing (you may need to refresh due to https://github.com/cfpb/hmda-platform-ui/issues/732)
 - Return to the upload page and select another new file to upload
 - Return to the institutions page (may need to refresh) and note the lack of warning

(this can also be observed in the API responses after the same process of POSTing twice to a previously signed submission.

schbetsy:
Closes #1115 

Also, fixes two bugs:
1. `CreateFiling` could previously persist multiple filings per filing period. Now it checks for duplicate filing periods before persisting new filings.
2. `UpdateFilingStatus` could previously overwrite all fields of a `Filing` object, not just the `status` and timestamps. Now it only updates the fields it should.

### DIFF
--- a/persistence/src/test/scala/hmda/persistence/institutions/FilingPersistenceSpec.scala
+++ b/persistence/src/test/scala/hmda/persistence/institutions/FilingPersistenceSpec.scala
@@ -3,7 +3,6 @@ package hmda.persistence.institutions
 import akka.testkit.TestProbe
 import hmda.model.fi._
 import hmda.persistence.messages.CommonMessages.GetState
-import hmda.persistence.demo.DemoData
 import hmda.persistence.institutions.FilingPersistence._
 import hmda.persistence.model.ActorSpec
 
@@ -26,15 +25,6 @@ class FilingPersistenceSpec extends ActorSpec {
       probe.send(filingsActor, GetState)
       probe.expectMsg(filings.reverse)
     }
-    "be able to change their status" in {
-      val filing = DemoData.testFilings.head
-      val modified = filing.copy(status = Cancelled)
-      probe.send(filingsActor, UpdateFilingStatus(modified))
-      probe.expectMsg(Some(modified))
-      probe.send(filingsActor, GetFilingByPeriod(filing.period))
-      probe.expectMsg(modified)
-    }
-
     "return None when persisting a filing period that already exists" in {
       val f1 = Filing("2018", "12345", Completed, true, 1483287071000L, 1514736671000L)
       probe.send(filingsActor, CreateFiling(f1))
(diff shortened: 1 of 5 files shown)


persistence/src/test/scala/hmda/persistence/institutions/FilingPersistenceSpec.scala
```diff
@@ -3,48 +3,67 @@ package hmda.persistence.institutions
...
-      probe.send(filingsActor, UpdateFilingStatus(modified))
-      probe.expectMsg(Some(modified))
-      probe.send(filingsActor, GetFilingByPeriod(filing.period))
-      probe.expectMsg(modified)
-    }
-
-    "return None when persisting a filing that already exists" in {
-      val f1 = Filing("2016", "12345", Completed, true, 1483287071000L, 1514736671000L)
+    "return None when persisting a filing period that already exists" in {
+      val f1 = Filing("2018", "12345", Completed, true, 1483287071000L, 1514736671000L)
```
nickgrippin: We shouldn't need this first section of the test, right?  It's only supposed to make sure `None` is returned for the 2016 filing.

validation/src/test/resources/application.conf
```diff
@@ -15,5 +15,6 @@ akka {
 
 hmda {
   isDemo = false
+  edits.demoMode = false
```
nickgrippin: Why is this config being added to the validation and persistence project?

schbetsy: It's useful to me when I'm developing on a branch. Sometimes in that process I'll go into a specific project in sbt (like `project persistence`) and run the specs there. When I did that recently, I found out that I needed this config for the specs to pass. 

(Note that both places where I added `edits.demoMode = false` are in `test` config files, not `main`)

persistence/src/test/scala/hmda/persistence/institutions/FilingPersistenceSpec.scala
schbetsy: 👍  Gave that spec a more accurate name in fb10cde

PR #1125, opened 2017-09-06
Util tests

schbetsy:
Addresses #1071 

This adds tests for a few of the utilities in `ReportUtil`. Some of the remaining methods posed a bigger challenge and will take more work (e.g. adding the [Akka Streams Testkit](http://doc.akka.io/docs/akka/current/scala/stream/stream-testkit.html#streams-testkit), which has a steeper learning curve than I expected.), so I'll be adding them in a future PR.

### DIFF
--- a/publication/src/test/scala/hmda/publication/reports/util/ReportUtilSpec.scala
+++ b/publication/src/test/scala/hmda/publication/reports/util/ReportUtilSpec.scala
@@ -1,0 +1,21 @@
+package hmda.publication.reports.util
+
+import hmda.model.publication.reports.MSAReport
+import hmda.publication.reports.util.ReportUtil._
+import org.scalatest.{ AsyncWordSpec, MustMatchers }
+
+class ReportUtilSpec extends AsyncWordSpec with MustMatchers {
+
+  "msaReport" must {
+    "get MSA info for given fips code and return MSAReport object" in {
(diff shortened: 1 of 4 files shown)


publication/src/test/scala/hmda/publication/reports/util/ReportUtilSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  "calculateMedianIncomeIntervals" must {
+    "get median income info for given fips, convert to thousands" in {
+      calculateMedianIncomeIntervals(10380)(2) mustBe 18.267
+      calculateMedianIncomeIntervals(26980)(2) mustBe 81.027
+      calculateMedianIncomeIntervals(41100)(2) mustBe 58.145
+    }
+    "return array of 50%, 80%, 100%, and 120% levels of median income (in thousands)" in {
+      def roundTo3(v: Double) = BigDecimal(v).setScale(3, BigDecimal.RoundingMode.HALF_UP).toDouble
```
nickgrippin: Is this just used to avoid a really long decimal in the tests?

schbetsy: Yep! And I figured any precision after 3 decimal places does not matter, so if there are minor differences beyond the thousandths place, we may as well ignore them.

PR #1129, opened 2017-09-11
Include file name in Submissions API response

### LINKED ISSUE #1119: Include filename in the /submissions endpoint
The filename for a submission is ingested via the upload's Content Disposition and is being saved and used for the summary. See [here](https://github.com/cfpb/hmda-platform/blob/e682f7d6c95b870beb6d084cb17ccf2d9e6a6be1/persistence/src/main/scala/hmda/persistence/processing/HmdaRawFile.scala) and [here](https://github.com/cfpb/hmda-platform/blob/88afb4962acdc642eac9e14995706456511a4821/api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionSummaryPathsSpec.scala).

We display the filename at different times, across refreshes, to different users of the same institution, and different filings. Using localStorage doesn't cover all these use cases and disrupts our state modeling.

If the filename is available as a key in the [submission](https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md#latest-submission), this would clean up some nasty code on the front-end and allow us to display the correct filename in all cases for a given submission.

schbetsy:
Closes #1119

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
+++ b/persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
@@ -13,6 +13,7 @@ object SubmissionPersistence {
 
   case object CreateSubmission extends Command
   case class UpdateSubmissionStatus(id: SubmissionId, status: SubmissionStatus) extends Command
+  case class AddSubmissionFileName(id: SubmissionId, fileName: String) extends Command
   case class GetSubmissionById(id: SubmissionId) extends Command
   case object GetLatestSubmission extends Command
 
@@ -40,6 +41,12 @@ object SubmissionPersistence {
           }
 
           SubmissionState(submissions.updated(i, updatedSub))
+
+        case SubmissionFileNameAdded(id, name) =>
+          val sub = submissions.find(_.id == id).getOrElse(Submission())
+          val index = submissions.indexOf(sub)
+          val updated = sub.copy(fileName = name)
+          SubmissionState(submissions.updated(index, updated))
       }
     }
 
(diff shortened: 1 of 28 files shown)


persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
```diff
@@ -40,6 +41,12 @@ object SubmissionPersistence {
           }
 
           SubmissionState(submissions.updated(i, updatedSub))
+
+        case SubmissionFileNameAdded(id, name) =>
```
nickgrippin: We already have a `name` value defined in this scope, maybe change this to something else?

persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
```diff
@@ -83,6 +90,17 @@ class SubmissionPersistence(institutionId: String, period: String) extends HmdaP
         sender() ! None
       }
 
+    case AddSubmissionFileName(id, name) =>
```
nickgrippin: Same comment here.

persistence/src/main/scala/hmda/persistence/institutions/SubmissionPersistence.scala
schbetsy: Good call. Fixed in e2965f7.

schbetsy:
Woohoo! Travis is finally passing. Ready to merge?

nickgrippin:
I'm getting an unused import warning for `SubmissionSummaryPathsSpec` with the most recent change.

PR #1140, opened 2017-09-22
Reports Metadata

### LINKED ISSUE #1066: A&D reports metadata file
Rather than storing strings and other descriptive material in an A&D report object, move that information into a separate file. Similar to `edit-metadata.txt`.

schbetsy:
Closes #1066 

There are definitely ways to take this idea farther than what I've done here--to automate more parts of report construction from the metadata file. I offer this as a start, and we can continue to reevaluate as we make more reports. Definitely (as always) open to feedback on design.

### DIFF
--- a/publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
+++ b/publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
@@ -1,0 +1,49 @@
+package hmda.publication.reports.util
+
+import com.github.tototoshi.csv.CSVParser.parse
+import hmda.model.ResourceUtils
+import hmda.model.publication.reports.ReportTypeEnum
+import hmda.model.publication.reports.ReportTypeEnum.Disclosure
+import hmda.publication.reports.util.DispositionType.{ DispositionType, ReceivedDisp }
+
+object ReportsMetaDataLookup extends ResourceUtils {
+
+  private val lines = resourceLines("/reports-metadata.txt")
+
+  val values: Map[String, ReportMetaData] = lines.drop(1).map { line =>
+    val data = ReportMetaData.fromString(line)
+    data.reportId -> data
+  }.toMap
+
+}
+
+case class ReportMetaData(
+  reportId: String,
+  reportType: ReportTypeEnum,
+  reportTable: String,
+  dispositions: List[DispositionType],
+  description: String
+)
+
+case object ReportMetaData {
+
+  def fromString(line: String): ReportMetaData = {
+    val values = parse(line, '\\', ',', '"').getOrElse(List())
+    val reportId = values.head
+    val reportType = ReportTypeEnum.byName.getOrElse(values(1).toLowerCase, Disclosure)
+    val reportNumber = values(2)
+    val description = values(4)
+    val dispositions =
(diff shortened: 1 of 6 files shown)


publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
```diff
@@ -0,0 +1,49 @@
...
+  dispositions: List[DispositionType],
+  description: String
+)
+
+case object ReportMetaData {
+
+  def fromString(line: String): ReportMetaData = {
+    val values = parse(line, '\\', ',', '"').getOrElse(List())
+    val reportId = values.head
+    val reportType = ReportTypeEnum.byName.getOrElse(values(1).toLowerCase, Disclosure)
```
nickgrippin: I always prefer using `getOrElse()`, since it's always safe... But I'm not sure if it's right in this case.  If we create the metadata file incorrectly, it might actually be more useful for this line to fail on a plain `get()`.  Same on line 36.

publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  val values: Map[String, ReportMetaData] = lines.drop(1).map { line =>
+    val data = ReportMetaData.fromString(line)
+    data.reportId -> data
+  }.toMap
+
+}
+
+case class ReportMetaData(
+  reportId: String,
```
nickgrippin: If this is being stored in a map with the report ID as the key, I think this field might be redundant.

schbetsy: Strictly yes, that field isn't needed for report generation. But including `reportId` in the `ReportMetaData` was the cleanest way to get access to it on line 15 of this file, so I could make it the key of the Map. Since it's useful in that case, maybe it's okay to leave it?

publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
schbetsy: I wondered about that too. That works for me! Updated in c64cc9a544786bb21cbf3a3de3e8d4be78ed42fa.

publication/src/main/scala/hmda/publication/reports/util/ReportsMetadataLookup.scala
nickgrippin: Hmmmmmm... How about moving this line: 
`val values = parse(line, '\\', ',', '"').getOrElse(List())` 
up to line 14, then passing in the resulting array to the `forString` method and using `values.head` as the key?

schbetsy: That would totally work, but that seems like that is a worse separation of concerns. 

I guess I don't see why it's important to make the `ReportMetaData` structure as small as possible. Want to explain your reasoning?

schbetsy: Ignore my last comment :). I added a potential fix in 703af60000d7bea2a2eeec2d59eccc5223916859. Let me know what you think.

PR #1143, opened 2017-09-26
multistage build using scala 2.12.2, sbt 0.13.15

charandas:
Just wanted to run this by your team that this is possible with docker as of late.

https://medium.com/travis-on-docker/multi-stage-docker-builds-for-creating-tiny-go-images-e0e1867efe5a

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 # HMDA Platform Dockerfile
 # Version: 1.0.0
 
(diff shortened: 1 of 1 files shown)


jmarin:
@charandas thanks for your contribution. We are still on `docker` 1.13, so this will have to wait a little bit before we can incorporate it. We are upgrading to 17.x soon, I will test this and merge if there are no problems, it looks good! Thank you

PR #1144, opened 2017-09-27
D53

### LINKED ISSUE #1056: Table 5-3: Disclosure by Respondent ID
**Table 5-3**: _"Disposition of applications for refinancings on 1-to- 4 family and manufactured home dwellings"_

Overall table filters
---------------------
Property Type: 1, 2
Purpose of Loan: 3

Borrower Characteristic Categorization
---------------------------------------------
* [Race](https://github.com/cfpb/hmda-platform/wiki/Race-Categorization)
* [Ethnicity](https://github.com/cfpb/hmda-platform/wiki/Ethnicity-Categorization)
* [Minority Status](https://github.com/cfpb/hmda-platform/wiki/Minority-Status-Categorization)

Report Structure
-------------------
Data model [here](https://github.com/cfpb/hmda-platform/wiki/A&D-Reports-data-model)

`respondent_id`
`institution_name`
`table`: "5-3"
`type`: "Disclosure"
`desc`: "Disposition of Applications to Refinance Loans on 1-to-4 Family and Manufactured Home Dwellings, by Income, Race, and Ethnicity of Applicant"
`year`
`report-date`
`msa`
`applicantincomes`

Example of  `json` output [here](https://github.com/cfpb/hmda-viz-prototype/blob/gh-pages/disclosure/2013/wells_fargo_bank_na/anniston-oxford/5-1/5-1.json)

Dispositions
---------------------------------------------
This table includes the following dispositions:

- Applications Received: Action Taken = 1, 2, 3, 4, and 5
- Loans Originated: Action Taken = 1
- Applications Approved but not Accepted: Action Taken = 2
- Applications Denied by Financial Institution: Action Taken = 3
- Applications Withdrawn by Applicant: Action Taken = 4
- File Closed for Incompleteness: Action Taken = 5

Types of A&D Reports
---------------------------------------------
[Here](https://github.com/cfpb/hmda-platform/wiki/Types-of-A&D-reports)

schbetsy:
Closes #1056 

I found a few small ways to reduce duplication in report generation code, but there's still a ton of duplicated code. Something to look for as we continue is for ways to share more code between A&D reports that are very similar to each other.

### DIFF
--- a/publication/src/main/scala/hmda/publication/reports/disclosure/D53.scala
+++ b/publication/src/main/scala/hmda/publication/reports/disclosure/D53.scala
@@ -1,0 +1,77 @@
+package hmda.publication.reports.disclosure
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+import hmda.model.publication.reports.{ ApplicantIncome, Disposition, MSAReport }
+import hmda.publication.reports._
+import hmda.publication.reports.util.ReportUtil._
+import hmda.publication.reports.util.ReportsMetaDataLookup
+import hmda.query.model.filing.LoanApplicationRegisterQuery
+
+import scala.concurrent.Future
+
+case class D53(
+  respondentId: String,
+  institutionName: String,
+  year: Int,
+  msa: MSAReport,
+  applicantIncomes: List[ApplicantIncome],
+  total: List[Disposition],
+  reportDate: String = formattedCurrentDate,
+  table: String = D53.metaData.reportTable,
+  description: String = D53.metaData.description
+) extends DisclosureReport
+
+object D53 {
+  val metaData = ReportsMetaDataLookup.values("D53")
+  val dispositions = metaData.dispositions
+
+  // Table filters:
+  // Property Type: 1, 2
+  // Purpose of Loan: 3
+  def generate[ec: EC, mat: MAT, as: AS](
+    larSource: Source[LoanApplicationRegisterQuery, NotUsed],
+    fipsCode: Int,
+    respondentId: String,
+    institutionNameF: Future[String]
+  ): Future[D53] = {
+
+    val lars = larSource
+      .filter(lar => lar.respondentId == respondentId)
+      .filter(lar => lar.msa != "NA")
+      .filter(lar => lar.msa.toInt == fipsCode)
+      .filter { lar =>
+        (lar.loanType == 2 || lar.loanType == 3 || lar.loanType == 4) &&
+          (lar.propertyType == 1 || lar.propertyType == 2) &&
+          (lar.purpose == 1)
+      }
+    val larsWithIncome = lars.filter(lar => lar.income != "NA")
+
+    val msa = msaReport(fipsCode.toString)
+
+    val incomeIntervals = calculateMedianIncomeIntervals(fipsCode)
+    val applicantIncomesF = applicantIncomesWithBorrowerCharacteristics(larsWithIncome, incomeIntervals, dispositions)
+
+    val yearF = calculateYear(larSource)
+    val totalF = calculateDispositions(lars, dispositions)
+
+    for {
+      institutionName <- institutionNameF
+      year <- yearF
+      applicantIncomes <- applicantIncomesF
+      total <- totalF
+    } yield {
+
+      D53(
+        respondentId,
+        institutionName,
+        year,
+        msa,
+        applicantIncomes,
+        total
+      )
+    }
+
+  }
+
+}
--- a/publication/src/main/scala/hmda/publication/reports/disclosure/D53.scala
+++ b/publication/src/main/scala/hmda/publication/reports/disclosure/D53.scala
@@ -41,9 +41,8 @@ object D53 {
       .filter(lar => lar.msa != "NA")
       .filter(lar => lar.msa.toInt == fipsCode)
       .filter { lar =>
-        (lar.loanType == 2 || lar.loanType == 3 || lar.loanType == 4) &&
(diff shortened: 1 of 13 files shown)


publication/src/main/scala/hmda/publication/reports/util/ReportUtil.scala
```diff
@@ -62,6 +69,52 @@ object ReportUtil extends SourceUtils {
     )
   }
 
+  def applicantIncomesWithBorrowerCharacteristics[ec: EC, mat: MAT, as: AS](
```
schbetsy: It was hard to find a meaningful name for this method, since the data structure it produces is nested so deeply. I'm open to suggestions for improved names.

schbetsy:
FYI: my comments above are just to let you (the reviewers) in on my thought processes, and to provide seeds for continued conversations. This is definitely READY FOR REVIEW--that's why it's in an open PR, and why the issue is in the "ready for review" column of the project boards.

publication/src/main/scala/hmda/publication/reports/disclosure/DisclosureReports.scala
```diff
@@ -33,6 +33,8 @@ class DisclosureReports(val sys: ActorSystem, val mat: ActorMaterializer) extend
     d51F.map { d51 =>
       println(d51.toJson.prettyPrint)
     }
+
+    //val d53F = D53.generate(larSource, fipsCode, respId, institutionNameF)
```
nickgrippin: I know we're not actually generating these reports yet, but just out of curiosity, why is this commented out?

publication/src/main/scala/hmda/publication/reports/protocol/disclosure/D53Protocol.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import hmda.model.publication.reports.{ ApplicantIncome, Disposition, MSAReport }
+import hmda.model.publication.reports.ReportTypeEnum.Disclosure
+import hmda.publication.reports.disclosure.D53
+import hmda.publication.reports.protocol.{ ApplicantIncomeProtocol, MSAReportProtocol, ReportTypeEnumProtocol }
+
+import spray.json._
+import spray.json.DefaultJsonProtocol
+
+object D53Protocol
```
nickgrippin: This is almost an exact copy of `D51Protocol`.  Is there any way to reduce this duplication?

publication/src/main/scala/hmda/publication/reports/util/ReportUtil.scala
nickgrippin: We seem to be following a pattern of `calculateThis` with the methods we're calling.  Maybe `calculateApplicantIncomes`?  It returns a list of Applicant Income objects, and it's assigned to a value called `applicantIncomesF`.

publication/src/main/scala/hmda/publication/reports/disclosure/DisclosureReports.scala
schbetsy: tl;dr It seemed pointless to include it. ¯\\\_(ツ)_/¯

I don't want to add any more `println`s to the project, but if I uncomment that line as-is, it would change the return type of the `generateReports` method... and it seemed pointless to wrestle with that since this code isn't hooked up anyway.

publication/src/main/scala/hmda/publication/reports/protocol/disclosure/D53Protocol.scala
schbetsy: Excellent question. This duplication is bothering me too. (side note: there's also blatant duplication in `DisclosureReportGenerators`, since each generator is almost an exact copy of the others).

I'd love to come up with a better system for making these, and I want to give this the thought it deserves. Should we keep this PR open while I think about that?

publication/src/main/scala/hmda/publication/reports/util/ReportUtil.scala
schbetsy: Hmm. I hadn't thought about the inconsistency of the naming conventions in the `ReportUtil` file. The question of when to name a method with a verb (like `calculateDispositions`) vs with a noun (like `larsByIncomeInterval`) is interesting--I'm still forming my opinion on it. 

In our system, these methods don't have side effects, so the only part that matters is the return value. Thus, our code can kind of treat them like objects, so I think I prefer naming them as nouns. If we wanted to make them all consistent, I would vote to never use a verb, and to switch them all to be named with nouns. 

This kind of consistency isn't 100% essential, but I don't think I want to add any functions with verb names.

schbetsy: I could use `applicantIncomesByInterval` if we want a shorter name?

publication/src/main/scala/hmda/publication/reports/disclosure/D53.scala
```diff
@@ -0,0 +1,77 @@
...
+    respondentId: String,
+    institutionNameF: Future[String]
+  ): Future[D53] = {
+
+    val lars = larSource
+      .filter(lar => lar.respondentId == respondentId)
+      .filter(lar => lar.msa != "NA")
+      .filter(lar => lar.msa.toInt == fipsCode)
+      .filter { lar =>
+        (lar.loanType == 2 || lar.loanType == 3 || lar.loanType == 4) &&
```
nickgrippin: Are these the right filters for this report?

schbetsy: 😬 
Fixed in cfa1a80.

PR #1145, opened 2017-09-28
Loader refactor

### LINKED ISSUE #1112: Panel Loader needs HttpHeaders for https:// requests
Whenever making a request to an https:// endpoint, the admin API won't accept the request unless it has a header "Content-Type":"application/json".

### LINKED ISSUE #1132: Modify panel loader to use TCP server

jmarin:
Closes #1132
Closes #1112


api/src/main/scala/hmda/api/tcp/admin/InstitutionAdminTcpApi.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  val tcpHandler: Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString]
+      .via(framing)
+      .drop(1)
+      .via(byte2StringFlow)
+      .map(x => InstitutionParser(x))
+      .mapAsync(parallelism = buffer)(i => createInstitution(i))
+      .mapAsync(parallelism = buffer)(i => createFiling(i))
+      .map(e => ByteString(e.toString))
```
nickgrippin: Just for clarification, is this the Filing object being sent back to the source?

jmarin: Yes. The Flow requires something to be sent back. We could change what that is, for now I've opted to send back the last element type in the stream

nickgrippin:
Just a couple things from testing
- What should the settings be for a full docker-compose setup?  I tried setting `HMDA_PANEL_LOADER_HOST=192.168.99.100` (leaving the port as-is), but wasn't able to connect.  Are there any changes necessary to the docker-compose file?
- There's currently no way to see if the loader actually connected to the api.  The program just exits with a "Success" message.  Maybe we could do something with the [`Sink.last` value?](https://github.com/cfpb/hmda-platform/pull/1145/files#diff-e252979c1d13f878d8d80b00b1429fc1R43)

PR #1150, opened 2017-10-02
Email on sign

### LINKED ISSUE #1126: Email the user when we've received their signature
The user's email is available in the JWT.

Sample content:
Email subject: __HMDA Filing Successful__
Email body:
Congratulations, you've completed filing your HMDA data for filing period `<filing period>`.
We received your filing on `<time stamp>`.
Your receipt is `<hash>`.

nickgrippin:
Closes #1126


api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
```diff
@@ -85,8 +93,58 @@ trait SubmissionSignPaths
 
     onComplete(fSubmission) {
       case Success(sub) =>
+        if (sub.status.code == 10) {
```
jmarin: Can this be replaced with `sub.status == Signed`? It would read clearer and be more type safe.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
```diff
@@ -85,8 +93,58 @@ trait SubmissionSignPaths
...
+
+    val session = Session.getDefaultInstance(properties)
+    val message = new MimeMessage(session)
+
+    val date = getFormattedDate
+
+    val text = s"$username,\n\nCongratulations, you've completed filing your HMDA data for filing period ${submission.id.period}.\n" +
+      s"We received your filing on: $date\n" +
+      s"Your receipt is: ${submission.receipt}"
+    message.setFrom(new InternetAddress("no-reply@test.com"))
```
jmarin: Maybe make this email address configurable?

jmarin:
Just a couple of minor comments, otherwise LGTM. Good job on this one @nickgrippin

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
```diff
@@ -85,8 +93,58 @@ trait SubmissionSignPaths
...
+    val emails = findEmailsById(submission.id.institutionId)
+    emails.map(emailSeq => {
+      emailSeq.foreach(t => {
+        val username = t._1 + " " + t._2
+        sendMail(t._3, username, submission)
+      })
+    })
+  }
+
+  private def sendMail(address: String, username: String, submission: Submission) = {
```
schbetsy: The team implied that the platform will eventually have other reasons to send email. Could this functionality move into a module so we can reuse it?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
```diff
@@ -85,8 +93,58 @@ trait SubmissionSignPaths
...
+    val text = s"$username,\n\nCongratulations, you've completed filing your HMDA data for filing period ${submission.id.period}.\n" +
+      s"We received your filing on: $date\n" +
+      s"Your receipt is: ${submission.receipt}"
+    message.setFrom(new InternetAddress("no-reply@test.com"))
+    message.setRecipients(Message.RecipientType.TO, address)
+    message.setSubject("HMDA Filing Successful")
+    message.setText(text)
+
+    log.info(s"Sending message to $address with the message \n$text")
+    Transport.send(message)
```
schbetsy: Is it possible to have a spec for this? It would be great to test that the message gets sent correctly and exactly once.

README.md
```diff
@@ -157,6 +157,11 @@ docker-compose -f docker-dev.yml up
 
 When finished, use `docker-compose down` to gracefully stop the running containers.
 
+In order to not get Postgres connection error upon signing a submission, you may want to start up the keycloak image as well.
+```shell
+docker-compose up keycloak
+```
+Not having the keycloak image running will produce an error in the console, but will not cause the platform to crash.
```
schbetsy: Is there a way to receive the signature email when running the platform from sbt? Like, if I were running a maildev image, could I get the email?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
```diff
@@ -85,8 +93,58 @@ trait SubmissionSignPaths
...
+    val properties = System.getProperties
+    properties.put("mail.smtp.host", host)
+    properties.put("mail.smtp.port", port)
+
+    val session = Session.getDefaultInstance(properties)
+    val message = new MimeMessage(session)
+
+    val date = getFormattedDate
+
+    val text = s"$username,\n\nCongratulations, you've completed filing your HMDA data for filing period ${submission.id.period}.\n" +
```
schbetsy: Should we include the institution name in the email text? Seems like that would be helpful for users who represent multiple institutions. @keileyo

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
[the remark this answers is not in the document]
nickgrippin: The front end requests the `/sign` endpoint multiple times, which calls the `completeWithSubmissionReceipt()` method.  The status would only not be `Signed` if the `fSign` Future didn't complete successfully, in which case we return an error.

That being said, I didn't account for a user returning to this page _after_ a signature.  I think that might trigger another email, so I'm switching up my method of detecting if an email is needed.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
nickgrippin: I'm going to go with the same approach we used for the A&D reports.  I'll leave it here until now, and if we later need it in another section of the code, we can work on it then.

README.md
nickgrippin: Yes, everything should work properly if all the necessary images are running.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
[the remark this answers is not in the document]
schbetsy: Do they really request the `POST` method of the `/sign` endpoint multiple times? I know they request the `GET` method to view the receipt, but if they're doing the `POST` more than once, that sounds like a bug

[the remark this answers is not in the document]
schbetsy: > The status would only not be Signed if the fSign Future didn't complete successfully, in which case we return an error.

In that case, why is the if statement necessary?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
nickgrippin: Given some of the research I've done, I don't think it's currently possible in our unit tests.  Whenever we start automating integration tests, that's definitely something we can look into.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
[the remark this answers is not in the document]
schbetsy: (Sorry for my erroneous comments. I'm reading more closely before I comment again 😬 )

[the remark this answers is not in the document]
schbetsy: Looks like the call to `emailSignature` should perhaps move to line 70 of this file, so it can only get triggered by the `POST` method of this endpoint.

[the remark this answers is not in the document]
nickgrippin: Both the `GET` and `POST` endpoints call the same method, `completedWithSubmissionReceipt()`.  I call the `emailSignature()` method from here because of the convenience of already having the `Submission` object, which includes fields necessary for the email.

However, from within the `completedWithSubmissionReceipt()`, there is no way of telling whether the user is requesting the `GET` or `POST` endpoint.  I remedied this by checking if the submission was signed, which would only happen after a successful call to the `POST` endpoint.  

My previous approach had a potential bug (detailed above), which I'm fixing soon in a new commit.

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
schbetsy: Cool, in that case, please add a GitHub issue for that work.

README.md
schbetsy: Could you add instructions for that in the README?

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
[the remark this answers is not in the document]
schbetsy: Thanks for the explanation. The fix looks 👍

api/src/main/scala/hmda/api/http/institutions/submissions/SubmissionSignPaths.scala
keileyo: This would be a nice to have, I think. What is LOE?

nickgrippin: It's not difficult.  Could you provide an updated email template with the institution name included?

README.md
nickgrippin: Update: Just realized that wouldn't be possible.  We retrieve the email addresses from Keycloak, and the user has to be verified in order for them to receive the submission email.  In a local `sbt` setup, the query for users would return an empty list.  It would theoretically be possible to manually insert users into the Keycloak database for purposes of testing, but that seems like too much trouble in comparison to the full docker-compose setup.

schbetsy: Thanks for looking into it! That makes sense.

PR #1157, opened 2017-10-10
Add specific dispatcher for cluster operations

### LINKED ISSUE #1154: Provide bulkhead dispatcher for cluster operations

jmarin:
Closes #1154

### DIFF
--- a/cluster/src/main/resources/application.conf
+++ b/cluster/src/main/resources/application.conf
@@ -33,6 +33,10 @@ akka {
 
     metrics.enabled = off
 
+    use-dispatcher = cluster-dispatcher
+
+    auto-down-unreachable-after = off
+
   }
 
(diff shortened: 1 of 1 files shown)


nickgrippin:
This is from the Akka docs.... Is this the reason for not using the default?

> If you have Cluster related problems when using the default-dispatcher that is typically an indication that you are running blocking or CPU intensive actors/tasks on the default-dispatcher. Use dedicated dispatchers for such actors/tasks instead of running them on the default-dispatcher, because that may starve system internal tasks.

jmarin:
Yes, potentially. I'm pretty sure we are running most actors on their own dispatchers, but I'm not 100% sure. This should help with isolating the cluster related operations (there are other things to look at, this one is easy)

PR #1159, opened 2017-10-10
Combine Disclosure 5-X Reports - Input Requested

schbetsy:
Since all of the Disclosure 5-1 through 5-5 reports have the exact same structure (and are generated in exactly the same way), could we use the same object to represent all of them? This could mean we use approximately 1/5 the amount of code, since we could share the code that generates the reports (`D5X`, instead of `D51`, `D52`, `D53`, etc), the json protocol (`D5XProtocol`), the spec data generator (`d5XGen`), and more.

If this works, we could extend this strategy to also consolidate Aggregate 5-X reports and other series of reports.

I understand that usually there is a reason to use different types for different objects: we want our data processing to be typesafe, and we don't want to risk combining or comparing un-like data. However, in this case, I think that using different types is unnecessary. Since we never read in or analyze these reports, we only generate these reports and send them away, I think shared types are typesafe enough. 

Agree? Disagree? I hope you'll give your opinions. I don't want to make a big change like this if y'all aren't on board.


publication/src/main/scala/hmda/publication/reports/disclosure/D51.scala
```diff
@@ -3,75 +3,25 @@ package hmda.publication.reports.disclosure
...
-        institutionName,
-        year,
-        msa,
-        applicantIncomes,
-        total
-      )
-    }
+  ): Future[D5X] = {
 
+    D5X.generate("D51", filters, larSource, fipsCode, respondentId, institutionNameF)
```
nickgrippin: I think this structure will work, especially due to the fact that we specify the table number in the object. 
👍

schbetsy:
Any more comments, or is this ready to merge? :octocat:

PR #1162, opened 2017-10-12
Load testing

### LINKED ISSUE #1155: Design load testing scenarios

jmarin:
Closes #1155


nickgrippin:
This adds a new dependency to our project, is that going to impact anything?

jmarin:
It's a test dependency, which doesn't get pulled into the jar when running the `assembly` command. I've double checked by commenting out the `test` option in `Dependencies.scala` and in that case it does get pulled in. The other testing libraries (`scalacheck`, `scalatest`) are treated the same way.

PR #1163, opened 2017-10-12
Cluster docs updated

### LINKED ISSUE #1100: Update docs with Cassandra cluster connection

nickgrippin:
Closes #1100

### DIFF
--- a/Documents/cluster.md
+++ b/Documents/cluster.md
@@ -19,12 +19,12 @@ The persistence cluster role contains actors responsible for tracking and persis
 
 ### Query
 
-This cluster role contains actors responsible for interacting with the backend database (currently Cassandra)
+This cluster role contains actors responsible for projecting data to a read-only database (currently Cassandra)
 
 ### Publication
 
-The publication role is responsible for creating and publishing the A&D reports from submitted LAR data
+The publication role is responsible for generating publication artifacts based on HMDA data (i.e. A&D reports, Modified LAR, etc.)
 
 ## Running cluster roles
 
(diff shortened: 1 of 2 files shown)


Documents/cluster.md
```diff
@@ -3,4 +3,28 @@
...
+
+The API cluster role is responsible for the four different APIs: Admin, Public, Filing, and the TCP Panel Loader
+
+### Persistence
+
+The persistence cluster role contains actors responsible for tracking and persisting information about submissions, institutions, and LAR validation
+
+### Query
+
+This cluster role contains actors responsible for interacting with the backend database (currently Cassandra)
```
jmarin: I think the following reads better and is more accurate: "This cluster role contains actors responsible for projecting data to a read-only database (currently Cassandra)"

Documents/cluster.md
```diff
@@ -3,4 +3,28 @@
...
+
+### Publication
+
+The publication role is responsible for creating and publishing the A&D reports from submitted LAR data
```
jmarin: I would change to something like this: "The publication role is responsible for generating publication artifacts based on HMDA data (i.e. A&D reports, Modified LAR, etc.)"

Documents/cluster.md
```diff
@@ -3,4 +3,28 @@
...
+
+## Running cluster roles
+
+The configuration setting `"akka.cluster.roles"` in `cluster/src/main/resources/application.conf` defines which cluster roles will start when the HMDA Platform starts.  By default, all four roles are set to start at the same time, but this can be changed to any single role or combination of roles.  Currently, this can only be done from within the config file; there is no environment variable override.
```
jmarin: The last statement is actually not true. The `HMDA_CLUSTER_ROLES` environment variable allows overriding the specific role(s) for a particular node.

nickgrippin: I thought I was going crazy, I knew there was some environment variable!  I forgot that it was defined in `HmdaPlatform.scala` rather than the `application.conf` file.

PR #1207, opened 2017-10-23
2016 lar loader

### LINKED ISSUE #1123: Provide mechanism to load previous years LAR data

jmarin:
Closes #1123

### DIFF
--- a/loader/src/main/scala/hmda/loader/lar/HmdaBatchLarLoader.scala
+++ b/loader/src/main/scala/hmda/loader/lar/HmdaBatchLarLoader.scala
@@ -3,8 +3,8 @@ package hmda.loader.lar
 import java.io.File
 import java.time.Instant
 
-import akka.actor.{ ActorPath, ActorRef, ActorSystem }
 import akka.pattern.ask
+import akka.actor.{ ActorPath, ActorRef, ActorSystem }
 import akka.cluster.client.{ ClusterClient, ClusterClientSettings }
 import akka.stream.{ ActorMaterializer, IOResult }
 import akka.stream.scaladsl.{ FileIO, Sink, Source }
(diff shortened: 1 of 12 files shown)


loader/src/main/scala/hmda/loader/lar/HmdaBatchLarLoader.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  def main(args: Array[String]): Unit = {
+
+    if (args.length < 2) {
+      exitSys(log, "Please provide a directory containing files with LAR data and period to process", 1)
+    }
+
+    val path = new File(args(0))
+    if (!path.isDirectory) {
+      exitSys(log, "Argument must be the full path to a directory containing files with LAR data", 2)
```
schbetsy: When I smoke test, I get this error message every time, even with full file paths (starting with `/Users/`), that are definitely LAR files.

schbetsy: When I switched this line to allow me to proceed, I get another error:

```
Association with remote system [akka.tcp://hmda@127.0.0.1:0] has failed, 
  address is now gated for [5000] ms. Reason: [Association failed with [akka.tcp://hmda@127.0.0.1:0]] 
  Caused by: [Can't assign requested address]
```

Maybe this is because I chose the wrong value for the `HMDA_CLUSTER_PORT` environment variable. If that's the case, where should I look to find the correct value?

jmarin: On the first error, you mention "full file paths", if you are passing individual files that might be the issue. It's expecting a directory (without trailing slash) with the files in it, followed by the period (i.e. 2016). In addition to this, the file names need to be in a specific format (the files we are using to test this are not public; ask about this offline). 

The association problem is most likely the port as you mention. You can do two things:
1) Set the application port explicitly with the `APP_PORT` environment variable when running the cluster, and use that same value in the `HMDA_CLUSTER_PORT` environment variable when running the `LAR` loader.
2) Look at the beginning of the logs when running the application, and taking note of the port that `Akka` assigns to the clustered application and assign it to the `HMDA_CLUSTER_PORT`. The 5th line or so after the HMDA Platform banner should have the information (i.e. _Remoting started; listening on addresses :[akka.tcp://hmda@127.0.0.1:**62096**]_). This is documented [here](https://github.com/jmarin/hmda-platform/blob/2016-lar-loader/Documents/2016_Data_load.md) as part of this pull request.

PR #1317, opened 2017-11-06
Switch actor lifecycle logs to debug

### LINKED ISSUE #1316: Improve actor logging
The application is too verbose right now even with `INFO` log level. Now that we are sure that Actors are started and stopped appropriately, log messages associated with the lifecycle of actors can be done at `DEBUG` level.

nickgrippin:
Closes #1316

### DIFF
--- a/cluster/src/main/scala/hmda/cluster/HmdaPlatform.scala
+++ b/cluster/src/main/scala/hmda/cluster/HmdaPlatform.scala
@@ -107,11 +107,11 @@ object HmdaPlatform extends App {
 
     (supervisorProxy ? FindActorByName(SingleLarValidation.name))
       .mapTo[ActorRef]
-      .map(a => log.debug(s"Started single lar validator at ${a.path}"))
+      .map(a => log.info(s"Started single lar validator at ${a.path}"))
 
     (supervisorProxy ? FindActorByName(InstitutionPersistence.name))
       .mapTo[ActorRef]
-      .map(a => log.debug(s"Started institutions at ${a.path}"))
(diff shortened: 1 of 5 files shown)


cluster/src/main/scala/hmda/cluster/HmdaPlatform.scala
```diff
@@ -107,11 +107,11 @@ object HmdaPlatform extends App {
...
     (supervisorProxy ? FindActorByName(SingleLarValidation.name))
       .mapTo[ActorRef]
-      .map(a => log.info(s"Started single lar validator at ${a.path}"))
+      .map(a => log.debug(s"Started single lar validator at ${a.path}"))
 
     (supervisorProxy ? FindActorByName(InstitutionPersistence.name))
       .mapTo[ActorRef]
-      .map(a => log.info(s"Started institutions at ${a.path}"))
+      .map(a => log.debug(s"Started institutions at ${a.path}"))
   }
```
jmarin: I would actually like to maintain these two as `INFO`, since they happen only once per application run and are proof that the `ActorSystem` and the main actors have started up correctly

validation/src/main/scala/hmda/validation/ValidationStats.scala
```diff
@@ -115,7 +115,7 @@ class ValidationStats extends HmdaPersistentActor {
   }
 
   override def preStart(): Unit = {
-    log.info(s"Actor started at ${self.path}")
+    log.debug(s"Actor started at ${self.path}")
```
jmarin: Same comment as above for this cluster singleton actor

PR #1321, opened 2017-11-09
Project LAR data to query side when submission is signed

### LINKED ISSUE #1282: Project LAR data to read side only when submission has been signed

jmarin:
Closes #1282

**Important Note**: this improves on the previous model by only projecting into the `lars2017` table when a submission is signed, as opposed to every time a submission is validated. It still doesn't address deleting previously signed submissions, so it does not fully address resubmissions. This [ticket](https://github.com/cfpb/hmda-platform/issues/1320) has been created for that.

This PR uses the [DistributedPubSub](https://doc.akka.io/docs/akka/current/scala/cluster-usage.html?_ga=2.72327060.605812294.1510233192-1301779606.1510233191#distributed-publish-subscribe) extension from `Akka` to deliver events across the cluster, and implements a publish / subscribe model to trigger actions based on subscribing to different topics (in this case, only the `SubmissionSignedPubSub` event is listened to in the `SubmissionSignedEventQuerySubscriber` class, but the model can be extended to other events).

To test, open a terminal and run the cluster with the `api` and `persistence` roles. On another terminal window, run the `query` and `publication` roles. Go through a normal submission process and upon signing, you should see logs written in the terminal with the `query` role displaying details about the `LAR` inserted into `Cassandra`. To check that data has been loaded correctly the following query can be used:

```shell
cqlsh> select count(*) from hmda_query.lars2017;  
```

### DIFF
--- a/cluster/src/main/resources/application.conf
+++ b/cluster/src/main/resources/application.conf
@@ -40,7 +40,11 @@ akka {
 
   }
 
-  extensions = ["de.heikoseeberger.constructr.ConstructrExtension", "akka.cluster.client.ClusterClientReceptionist"]
+  extensions = [
+    "de.heikoseeberger.constructr.ConstructrExtension",
+    "akka.cluster.client.ClusterClientReceptionist",
+    "akka.cluster.pubsub.DistributedPubSub"
+  ]
 
 }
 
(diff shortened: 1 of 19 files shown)


schbetsy:
The code here looks good to me. I have to head out soon, so I don't think I have time to smoke test, but I saw the demo yesterday, and it looked good. 

Also, a question: Why does this PR include `inst_data_2017_end.csv`?

jmarin:
That file was used for testing. Will remove it, not needed

PR #1332, opened 2017-11-14
Q030 fix

### LINKED ISSUE #1324: Edit Q030 triggering incorrectly
Beta users are reporting that Q030, which is checking that "MSA/MD, state, county, census tract should equal a valid combination and not NA" is triggering when it should not. The edit is triggering often when none of those fields are reported NA, and these combinations are passing validity edits V290, V295 and V300 (and I've confirmed that the MSA/MD, state, county and census tract combinations are valid using the FFIEC geocoder). 

I am able to replicate the error:
image

nickgrippin:
This fix offloads the logic of determining the CRA status of an Institution to the filer, allowing them to confirm that the information is correct or fix the LAR with a new upload.

Closes #1324

### DIFF
--- a/validation/src/main/scala/hmda/validation/rules/lar/quality/Q030.scala
+++ b/validation/src/main/scala/hmda/validation/rules/lar/quality/Q030.scala
@@ -1,48 +1,20 @@
 package hmda.validation.rules.lar.quality
 
-import hmda.model.fi.lar.{ Geography, LoanApplicationRegister }
-import hmda.model.institution.Institution
-import hmda.validation.context.ValidationContext
+import hmda.model.fi.lar.LoanApplicationRegister
 import hmda.validation.dsl.PredicateCommon._
-import hmda.validation.dsl.PredicateGeo._
 import hmda.validation.dsl.PredicateSyntax._
-import hmda.validation.dsl.{ Failure, Result, Success }
-import hmda.validation.rules.{ EditCheck, IfInstitutionPresentIn }
+import hmda.validation.dsl.Result
+import hmda.validation.rules.EditCheck
 
-class Q030 private (institution: Institution) extends EditCheck[LoanApplicationRegister] {
+object Q030 extends EditCheck[LoanApplicationRegister] {
   override def name: String = "Q030"
 
   override def apply(lar: LoanApplicationRegister): Result = {
     when(lar.actionTakenType is containedIn(1 to 6)) {
-      val geo: Geography = lar.geography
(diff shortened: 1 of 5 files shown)


validation/src/main/scala/hmda/validation/rules/lar/quality/Q030.scala
```diff
@@ -1,48 +1,20 @@
...
-              when(geo.tract not "NA") {
-                geo is validCompleteCombination
-              }
-          } and when(geo.msa is "NA") {
-            when(geo.tract not "NA") {
-              geo is validStateCountyTractCombination
-            }
-          }
-      }
+      (lar.geography.msa not "NA") and
```
schbetsy: This only works if the user exactly follows the FIG, using `"NA"` and not `"na"` or `"not applicable"` or `""` or anything. After Juan's #1336 is in, do we need to protect against any of those options?

validation/src/test/scala/hmda/validation/rules/lar/quality/Q030Spec.scala
```diff
@@ -1,210 +1,33 @@
...
     }
   }
 
-  implicit class LarChecker(lar: LoanApplicationRegister) {
-    def mustFail(implicit fi: Institution) = check(lar, fi) mustBe a[Failure]
-    def mustPass(implicit fi: Institution) = check(lar, fi) mustBe a[Success]
-    def check(lar: LoanApplicationRegister, fi: Institution) = Q030.inContext(ValidationContext(Some(fi), None)).apply(lar)
+  property("fails when action taken is in range and geography is NA") {
+    forAll(larGen, Gen.choose(1, 6)) { (lar, i) =>
+      val invalidGeo = Geography("NA", "NA", "NA", "NA")
```
schbetsy: We might want a spec or two that shows what happens when only one, two, or three of the Geography fields is "NA".

validation/src/main/scala/hmda/validation/rules/lar/quality/Q030.scala
nickgrippin: [This line](https://github.com/cfpb/hmda-platform/pull/1336/files#diff-d44a9bbfd5df9409b42d4f4649baa8d2R175) in his PR means that all the values will be strictly "NA".

schbetsy:
Code here looks good to me. Have you been working with @keileyo to double-check that the business logic is exactly what she asked for?

If so, I think we're good to merge.

keileyo:
looks good to me

PR #1336, opened 2017-11-15
Stuck validation

### LINKED ISSUE #1325: Beta users reporting hanging on validation
Multiple users on HMDA Help are reporting that the application open beta is hanging during the "validation" process. Some users have reported they waited "hours" before giving up. Some have tried again but still experience hanging.

jmarin:
Closes #1325


nickgrippin:
I've tested with the problematic files located in the Google Drive, they upload and report back with the expected parsing errors.
👍

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
```diff
@@ -145,16 +145,49 @@ object LarCsvParser {
...
     }
   }
+
+  def checkNA(value: String, fieldName: String): ValidationNel[String, List[Int]] = {
+    if (value == "") {
+      s"$fieldName cannot have empty value".failure.toValidationNel
+    } else if (value == "NA") {
+      List(0).success
+    } else {
+      val dot = value.indexOf(".")
```
nickgrippin: This if-else logic could be simplified to `value.replaceFirst(raw"\.", "")`, while still making sure that a string like "1.2.3" will fail.

parser/shared/src/main/scala/hmda/parser/fi/lar/LarCsvParser.scala
```diff
@@ -145,16 +145,49 @@ object LarCsvParser {
...
       )
 
-      val validationList = numericFields.map { case (key, value) => toIntOrFail(value, key) }
+      val NAFields = ListMap(
+        "Date Application Received" -> fields(4),
+        "MSA" -> fields(13),
+        "State" -> fields(14),
+        "County" -> fields(15),
+        "Census Tract" -> fields(16),
+        "Rate Spread" -> fields(36)
```
schbetsy: "Applicant Income" can also be NA.

PR #1343, opened 2017-11-21
Tract update

### LINKED ISSUE #1323: Update census file data for edits V290, V295 and V300
Users are reporting errors w/ edits V290, V295 and V300. I can confirm that the platform is triggering these edits on geocoding data produced by the ffiec geocoder.  Working assumption is because we are using a census file from july 2015 as that was the latest at the time of upload see: #656 

Need to generate an updated file with the latest census data. 2016 data can be found here: https://www.census.gov/data/datasets/2016/demo/popest/total-metro-and-micro-statistical-areas.html

nickgrippin:
This updates our tract data source to the FFIEC Census flat file.

Closes #1323

### DIFF
--- a/census/src/main/scala/hmda/census/model/TractLookup.scala
+++ b/census/src/main/scala/hmda/census/model/TractLookup.scala
@@ -1,17 +1,20 @@
 package hmda.census.model
 
+import com.github.tototoshi.csv.CSVParser._
+
 // The file contains a list of all census tracts and what county and state they are in
 // site: https://www.census.gov/geo/maps-data/data/relationship.html
 // file: http://www2.census.gov/geo/docs/maps-data/data/rel/trf_txt/us2010trf.txt
 
 object TractLookup extends CbsaResourceUtils {
   val values: Set[Tract] = {
-    val lines = csvLines("/us2010trf.txt")
+    val lines = resourceLines("/ffiec_census_fields.txt", "ISO-8859-1")
 
-    lines.map { values =>
-      val stateFips2010 = values(9)
-      val countyFips2010 = values(10)
-      val tractFips2010 = values(11)
+    lines.drop(1).map { line =>
+      val values = parse(line, '\\', '|', '"').getOrElse(List())
+      val stateFips2010 = values(2)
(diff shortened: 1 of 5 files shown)


census/src/main/scala/hmda/census/model/TractLookup.scala
```diff
@@ -1,17 +1,20 @@
...
 
+import com.github.tototoshi.csv.CSVParser._
+
 // The file contains a list of all census tracts and what county and state they are in
 // site: https://www.census.gov/geo/maps-data/data/relationship.html
 // file: http://www2.census.gov/geo/docs/maps-data/data/rel/trf_txt/us2010trf.txt
 
 object TractLookup extends CbsaResourceUtils {
   val values: Set[Tract] = {
-    val lines = csvLines("/us2010trf.txt")
```
schbetsy: Do we still need `us2010trf.txt` in the `/resources` folder, or can we delete it?

census/src/main/scala/hmda/census/model/TractLookup.scala
```diff
@@ -1,17 +1,20 @@
 package hmda.census.model
 
+import com.github.tototoshi.csv.CSVParser._
+
 // The file contains a list of all census tracts and what county and state they are in
 // site: https://www.census.gov/geo/maps-data/data/relationship.html
 // file: http://www2.census.gov/geo/docs/maps-data/data/rel/trf_txt/us2010trf.txt
```
schbetsy: Are these comments still correct?

schbetsy:
I also noticed a weird issue: The classes in the `census` project have the correct package name (`hmda.census.model`), but the specs for those classes don't match (they just have `model` as their package name).

Files I noticed that have that issue: `StateAbrvLookupSpec`, `CbsaLookupSpec`, `PrPopLookupSpec`, `StatesPopLookupSpec`, `MsaIncomeLookupSpec`, and `TractLookupSpec`. And `TractToCbsaSpec` doesn't have a package designation at all.

Would you be willing to update those package names to `hmda.census.model` as part of this PR? It should be a quick fix.

PR #1344, opened 2017-11-21
Modified lar

### LINKED ISSUE #1284: Create Modified LAR data stream

jmarin:
Closes #1284 
Note: to test, an `S3` bucket needs to be set up and the application set up with the appropriate credentials. The publication stream for modified `LAR` will fail silently if these two requirements are not met.


schbetsy:
What is the reasoning for having the publication stream fail silently if the setup is incorrect? Seems like it would be much more useful and user-friendly if we showed a descriptive error message.

Also, is there documentation for this feature anywhere?

jmarin:
That behavior is the default for `Akka Streams`, so this is common to all the streams in the project. The `S3` upload is implemented as an `Akka Stream` with `Alpakka`. I created this [ticket](https://github.com/cfpb/hmda-platform/issues/1349) last week to review this and change where appropriate.

query/src/main/scala/hmda/query/repository/filing/LarConverter.scala
```diff
@@ -45,10 +44,9 @@ object LarConverter {
       lar.applicant.coSex,
       lar.applicant.income,
       lar.denial.reason1,
-      lar.denial.reason2,
       lar.denial.reason3,
-      "",
-      ""
+      lar.denial.reason3,
```
schbetsy: Typo here: `reason3` is listed twice, and `reason2` is missing.

query/src/main/scala/hmda/query/model/filing/ModifiedLoanApplicationRegister.scala
```diff
@@ -1,44 +1,43 @@
 package hmda.query.model.filing
 
 case class ModifiedLoanApplicationRegister(
```
schbetsy: In addition to this `ModifiedLoanApplicationRegister` class, there's also one defined in `api-model/src/main/scala/hmda/api/model/ModifiedLoanApplicationRegister.scala`. The one in the `api-model` project can be deleted, right?

Issue #1416, opened 2017-12-26
Error on rate spread endpoint


awolfe76:
https://github.com/cfpb/hmda-platform-tools/issues/77 was created but the error is actually on the endpoint.

Some log info:

```
[2017-12-26 10:13:34,132][INFO ][hmda-akka.actor.default-dispatcher-19] Setting up scheduled job 'AporCalculator', with 'com.typesafe.akka.extension.quartz.QuartzCronSchedule@7d5582ae'
[2017-12-26 10:13:34,133][ERROR][hmda-akka.actor.default-dispatcher-287] Unable to store Job : 'DEFAULT.AporCalculator_Job', because one already exists with this identification.
akka.actor.ActorInitializationException: akka://hmda/user/supervisor/singleton/hmda-apor-persistence: exception during creation
    at akka.actor.ActorInitializationException$.apply(Actor.scala:193)
    at akka.actor.ActorCell.create(ActorCell.scala:608)
    at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:462)
    at akka.actor.ActorCell.systemInvoke(ActorCell.scala:484)
    at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:282)
    at akka.dispatch.Mailbox.run(Mailbox.scala:223)
    at akka.dispatch.Mailbox.exec(Mailbox.scala:234)
    at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
    at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
    at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
    at akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
Caused by: org.quartz.ObjectAlreadyExistsException: Unable to store Job : 'DEFAULT.AporCalculator_Job', because one already exists with this identification.
    at org.quartz.simpl.RAMJobStore.storeJob(RAMJobStore.java:279)
    at org.quartz.simpl.RAMJobStore.storeJobAndTrigger(RAMJobStore.java:251)
    at org.quartz.core.QuartzScheduler.scheduleJob(QuartzScheduler.java:886)
    at org.quartz.impl.StdScheduler.scheduleJob(StdScheduler.java:249)
[...]

PR #1417, opened 2017-12-26
Start HmdaAPORPersistence actor even if scheduler has errors

schbetsy:
Addresses #1416


schbetsy:
Closing in favor of #1418

PR #1418, opened 2017-12-27
Start APOR load scheduler upon system startup

schbetsy:
Previously, it was started every time the APORPersistence actor started
up, which caused an unrecoverable error whenever the actor re-started
after timeout.

Addresses #1416


nickgrippin:
With the scheduler now being created in the cluster project, would it make sense to move the associated configuration over to that application.conf file?  It will work either way, but moving it to the same project might make it easier to find.

schbetsy:
Sure, I'll try that out. 👍

PR #1437, opened 2018-01-09
Apor modify

### LINKED ISSUE #1435: Modify persisted data (APOR tables)
Allow updates to APOR tables by date that are already in the system

jmarin:
Closes #1435


api/src/test/scala/hmda/api/http/admin/APORAdminHttpApiSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+        responseAs[AporCreated] mustBe AporCreated(apor1, FixedRate)
+      }
+    }
+    "Modify APOR" in {
+      val modifyAporJsonRequest = ByteString(ModifyAporRequest(apor2, FixedRate).toJson.toString)
+      val putRequest = createRequest(modifyAporJsonRequest, HttpMethods.PUT)
+      putRequest ~> aporRoutes(supervisor) ~> check {
+        status mustBe StatusCodes.Accepted
+        responseAs[AporModified] mustBe AporModified(apor2, FixedRate)
+      }
```
schbetsy: Currently, when a user requests the Modify APOR endpoint with a date + rate type combination that does not exist in the system, the request times out and this error appears in the logs: 

```
Ask timed out on [Actor[akka://hmda/user/supervisor/singleton/hmda-apor-persistence#1228394427]] after [10000 ms]. 
  Sender[null] sent message of type "hmda.persistence.messages.commands.apor.APORCommands$ModifyApor"
```

Please update it to return a `404` instead, and add a spec here showing that the behavior is correct.

Documents/admin-api.md
```diff
@@ -173,3 +173,136 @@ This API is for internal use only. The following endpoints are available
...
+        			4.03,
+        			4.03,
+        			4.03,
+        			4.03
+        		]
+        	}
+        }
+        ```
+
+    If a value for that date already exists, this new entry will not be created (see below for endpoint to update values).
```
schbetsy: From my tests, I think this line of documentation isn't accurate: "If a value for that date already exists, this new entry will not be created". In my testing, I've seen it create a new APOR record each time there's a `POST`, even if the new record conflicts with existing ones. 

If it **should** be rejecting any conflicting requests, we'll need to update line 116 of `HmdaAPORPersistence.scala` to check whether the date + amortization combo already exists in memory.

If we do that, we should remove this bit, since it won't be accurate anymore: "If an APOR record already exists in the system with the given rate type and date, the new record will effectively overwrite the existing one. Both will exist in the system, but the most recently added one will be used in rate spread calculations"

jmarin: I'm removing that line in the documentation and not changing the behavior. The APOR values are in general not going to be inserted through this mechanism (they will be ingested through S3). This method is just a fallback in case that fails.

PR #1448, opened 2018-01-16
Institution delete

### LINKED ISSUE #1441: Code change to allow for deletion of duplicate institution IDs
There is a duplicate institution in platform that needs to be removed. The duplicate institution has a space at the end of its Respondent ID value and, as a result, it also cannot be modified.

jmarin:
Closes #1441

### DIFF
--- a/Documents/admin-api.md
+++ b/Documents/admin-api.md
@@ -118,7 +118,57 @@ This API is for internal use only. The following endpoints are available
     }
     ```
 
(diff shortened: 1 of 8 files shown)


nickgrippin:
Do we want to delete any possible associated Filing, or are we all right with leaving that as an orphan?

jmarin:
Good question. I think it's fine to leave them, in case someone asks "what filings did that removed institution have".

Issue #1410, opened 2017-12-19
Large LAR files generating error message


colleenms:
Users uploading large LAR files (>100k) are reporting receiving the below error message and needing to refresh the page multiple times during the submission process. They are able to complete a submission.

![image](https://user-images.githubusercontent.com/32906749/34018052-4e6436ba-e0f6-11e7-8515-efeb9c8e7e48.png)

keileyo:
Can confirm that this happens w/ a 22k LAR test file. Test file that replicates this error is in drive.

keileyo:
Some behavior that I've exhibited testing on HMDA3 today with large files (100k). 

Uploaded a 100k file with parsing errors in the transmittal sheet, after an hour of "uploading" i logged out and logged back in. At that point, i was prompted with the error message in the original comment above. I refreshed the page and it showed me the parsing errors. 

Logs show that the parsing with errors actually finished 10 minutes after the original upload, so it seems that leaving the platform and logging back in prompted the _refresh the page_ error & new status of _your file has formatting errors_ to be present in the UI. 

Uploaded another 100k file without parsing errors, and saw it move from uploading to validating the edits within 5 minutes. From there:
- The file stayed in _your data is being analyzed_ for over two hours
- After two hours, I logged back in and got only a spinning wheel
- Refreshed the page and received the "Sorry, and error has occured. Please refresh the page" message. Refreshed the page, and syntactical and validity edits were served in the UI.

keileyo:
Testing prod during off hours:
- both the 100k file and 250k files produced parsing errors stating that the an unexpected number of data fields was found in a single LAR. In both cases, the files actually had the correct number of data fields in the LAR.  In the 100k file, this happened in row 98,945 and in the 250k file, this happened in row 198,897. 

- a second upload of the same 100k file made it past parsing. After 20 minutes of showing _uploading_ in the UI, I signed out and signed back in immediately and saw that the status moved forward to _your data is being analyzed_. After another ~1.5 hours, i logged out and logged back in and saw that my status moved to _validated with errors_.  
- Upon attempting to review the edits, I received just a spinning wheel in the UI, refreshed the page, received the error message in the first comment, refreshed again, and got stuck with a spinning wheel. The edits never appeared.

![image](https://user-images.githubusercontent.com/13036571/34880104-61186f0e-f77d-11e7-9462-257733f5dff7.png)

keileyo:
[...]

PR #1450, opened 2018-01-17
Improve API performance for large files (stream edits)

schbetsy:
Addresses #1410 

Main changes, with notes about the drawbacks of this approach:
- `HmdaFileValidator` now holds its state in 3 separate variables. 
  - `SVState` and `QMState` hold the list of unique edit names that this submission triggered (one for each different named edit, *not* one per edit on each line). *This currently uses `Set` logic to persist only unique names, which is slow when re-hydrating state after the validator actor stops.*
  - `HmdaVerificationState` holds basic information about the submission: whether quality/macro edits have been verified, the validated Transmittal Sheet, and the number of validated lars (for the submission summary).
- `/edits` uses the `SVState` and `QMState` to find the list of unique edits found in this submission. It no longer needs to analyze the entire list of edits on every line.
- `/edits/<edit>` uses the event stream to find a) the 20 individual edits to return in this page of edits, and b) the 20 lars that those edits happened on. *This is a slow process, and I'm not sure how to speed it up without changing the event model of the actor.*

Known: large files (>100k lars) still have some issues
- the first load of `/edits` times out if the actor was asleep. Subsequent calls are fast.
- calls to `/edits/<edit>` are slow, and frequently time out.

I'm open to feedback on the current approach. I'm also open to merging this now and continuing to look for improvements in future PRs.


api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
```diff
@@ -94,9 +81,14 @@ class SubmissionEditPathsSpec extends InstitutionHttpApiSpec with LarGenerators
...
+      val fState = for {
+        s <- fHmdaValidatorActor(2)
+        xs <- (s ? GetState).mapTo[HmdaVerificationState]
+        svState <- (s ? GetSVState).mapTo[SVState]
+      } yield (xs, svState.containsSVEdits)
+      val (state: HmdaVerificationState, containsSVEdits: Boolean) = Await.result(fState, 5.seconds)
+
+      containsSVEdits mustBe false
       state.macroVerified mustBe false
       state.qualityVerified mustBe false
```
jmarin: You can get rid of the blocking `Await.result` by extending `InstitutionHttpApiAsyncSpec` instead of `InstitutionHttpApiSpec` and replacing the previous 4 lines with:

```scala
fState.map {
        case (state: HmdaVerificationState, containsSVEdits: Boolean) =>
          containsSVEdits mustBe false
          state.macroVerified mustBe false
          state.qualityVerified mustBe false
      }
```

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
```diff
@@ -113,9 +105,9 @@ class SubmissionEditPathsSpec extends InstitutionHttpApiSpec with LarGenerators
       responseAs[EditsVerifiedResponse] mustBe EditsVerifiedResponse(true, currentStatus)
 
       // test that it updates validation state
-      val state: HmdaFileValidationState = Await.result(fValidationState, 5.seconds)
-      state.macroVerified mustBe true
+      val state = Await.result(fVerificationState, 5.seconds)
       state.qualityVerified mustBe false
+      state.macroVerified mustBe true
```
jmarin: You can get rid of the blocking `Await.result` by replacing with 

```scala
fVerificationState.map { state =>
  state.qualityVerified mustBe false
  state.macroVerified mustBe true
}
```

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
```diff
@@ -130,8 +122,9 @@ class SubmissionEditPathsSpec extends InstitutionHttpApiSpec with LarGenerators
       responseAs[EditsVerifiedResponse] mustBe EditsVerifiedResponse(true, currentStatus)
 
       // test that it updates validation state
-      val state: HmdaFileValidationState = Await.result(fValidationState, 5.seconds)
+      val state = Await.result(fVerificationState, 5.seconds)
       state.qualityVerified mustBe true
+      state.macroVerified mustBe true
```
jmarin: You can get rid of the blocking call by replacing with the following:

```scala
fVerificationState.map { state =>
  state.qualityVerified mustBe true
  state.macroVerified mustBe true
}
```

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
```diff
@@ -142,20 +135,10 @@ class SubmissionEditPathsSpec extends InstitutionHttpApiSpec with LarGenerators
...
-      status mustBe StatusCodes.MethodNotAllowed
-    }
-  }
 
-  "Edit Type endpoint: return 405 when posting verification to validity endpoint" in {
-    postWithCfpbHeaders("/institutions/0/filings/2017/submissions/0/edits/validity") ~> Route.seal(institutionsRoutes(supervisor, querySupervisor, validationStats)) ~> check {
-      status mustBe StatusCodes.MethodNotAllowed
+      val state = Await.result(fVerificationState, 5.seconds)
+      state.qualityVerified mustBe false
+      state.macroVerified mustBe true
```
jmarin: You can get rid of the blocking `Await.result` by changing to the following:

```scala
 fVerificationState.map { state =>
  state.qualityVerified mustBe false
  state.macroVerified mustBe true
}
```

api/src/test/scala/hmda/api/http/institutions/submissions/SubmissionEditPathsSpec.scala
schbetsy: Good call. Done in 67ef1ebc8b75c7128696d4c1f08138479d75095e.

jmarin:
I'm ok with merging as is. Have been testing it, and while it doesn't address the whole problem, it's an improvement. 
@schbetsy can you create tickets for the known issues? (if you haven't done so already)

schbetsy:
Good call. I created #1465 and #1466 for the two outstanding issues I know of.

PR #1451, opened 2018-01-17
Update S025 language

### LINKED ISSUE #1449: Update edit language for S025
A large number of users are generating the S025 edit and contacting HMDA Help for guidance. If we add the following to the edit language, I think users will be more clear on what needs to be changed:

"Please note the respondent-id should not include any leading zeros or a dash"

I believe the current language in platform is as follows: 
"Control number must = a valid respondent identifier/agency code combination for date processed.,The respondent-id and agency code provided in the file does not match our records. Please verify the reported respondent-id and agency code and update your file accordingly."

@keileyo any other thoughts on this?

nickgrippin:
Closes #1449


jmarin:
@colleenms Can you check the changes here? If good, I will merge. Thanks

colleenms:
Yes, changes look good.

PR #1459, opened 2018-01-25
Generate and publish disclosure reports

### LINKED ISSUE #1395: Hook up Disclosure reports to backend
Disclosure reports are not being generated automatically at this point. Incorporate generation of disclosure reports as an event triggered upon submission being signed, and store `json` content in the corresponding `S3` bucket.

nickgrippin:
Local testing instructions:
* Verify that AWS environment variables are set correctly
* Run the backend, and get a submission through to the "sign" step.  I recommend using very small files initially.  A file with 30 different MSAs will take roughly an hour to publish the 1000+ reports.
* After signing, the report generation will begin.  This can be verified in the logs at the INFO level.
* Check the S3 bucket to verify that reports are being published at `/reports/disclosure/{instId}/{msa}/{reportId}.txt`
* To test the publication again without uploading another file, POST to the admin API at `/disclosure/{instId}/{year}/{submissionId}`

Things left to do:
* Try to implement more parallelism to speed up the process without overloading Cassandra
* Documentation
* Tests

Closes #1395


api/src/main/scala/hmda/api/http/admin/PublicationAdminHttpApi.scala
```diff
@@ -0,0 +1,51 @@
...
+trait PublicationAdminHttpApi extends HmdaCustomDirectives with ApiErrorProtocol {
+
+  implicit val system: ActorSystem
+  implicit val materializer: ActorMaterializer
+  implicit val timeout: Timeout
+
+  val log: LoggingAdapter
+
+  def disclosureGenerationPath(supervisor: ActorRef) =
+    path("disclosure" / Segment / IntNumber / IntNumber) { (instId, year, subId) =>
```
schbetsy: If the submission hasn't been signed, the data won't be in the LAR repository yet, so the reports will all be empty. Let's make sure that doesn't happen by returning a `400` if the submission isn't signed.

nickgrippin: Done in most recent commit

publication/src/main/scala/hmda/publication/reports/disclosure/DisclosureReportPublisher.scala
```diff
@@ -0,0 +1,164 @@
...
+    case _ => //do nothing
+  }
+
+  private def generateReports(submissionId: SubmissionId): Future[Unit] = {
+    val futures = for {
+      i <- getInstitution(submissionId.institutionId).mapTo[Institution]
+      irs <- getMSAFromIRS(submissionId)
+    } yield (i, irs)
+
+    futures.map(f => {
```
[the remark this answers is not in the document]
nickgrippin: Good proposal, but I'm going to leave that for a future refactor if we deem it necessary.

PR #1460, opened 2018-01-27
Fix unquoted commas in 2018_File_Spec_Table2.csv

tmeneau:
Quotes any descriptions for the 2018 file format that contain commas per the CSV specification (see references below). In addition to making the file valid CSV, it allows Github to enable it's helpful CSV features when viewed online (like nice table formatting and search).

Just for future reference (mostly my own), I used the following regex and replace values in Sublime Text 3 to find and quote the offending columns:

```regex
regex: ^(([^,\n]*, ){4})((?:[^,\n]*, [^,\n]*)+)(, [^,\n]*,)$
replace: $1"$3"$4
```

#### References

* This [Stack Overflow answer](https://stackoverflow.com/questions/769621/dealing-with-commas-in-a-csv-file)
* And the [IETF RFC for the CSV format](https://tools.ietf.org/html/rfc4180#page-2) _(section 2(6))_


jmarin:
@tmeneau Thanks for your contribution!

tmeneau:
Thanks for the fast merge!

PR #1469, opened 2018-01-31
Decimal parsing

### LINKED ISSUE #1452: Files getting stuck in validating
Files get stuck in validating when income is formatted incorrectly. Enhance parser to catch this. 

After further review, we determined the issue is being caused by the income field, which had values containing decimal places, e.g. 163.956. The income field needs to be rounded to the nearest thousand, without commas, or reported as NA.

~~We replicated an issue that results in a file getting stuck in validation. What we know:~~
~~- Certain required transmittal sheet fields are left blank~~
~~- Even when the blank transmittal sheet fields are populated, the issue still gets stuck in validation~~

jmarin:
Closes #1452

### DIFF
--- a/parser/shared/src/test/scala/hmda/parser/fi/lar/LarCsvParserSpec.scala
+++ b/parser/shared/src/test/scala/hmda/parser/fi/lar/LarCsvParserSpec.scala
@@ -46,13 +46,13 @@ class LarCsvParserSpec extends PropSpec with PropertyChecks with MustMatchers wi
   property("Fail parsing when NA is not properly spelled") {
     forAll(larGen) { lar =>
       val na = lar.copy(rateSpread = "na")
-      LarCsvParser(na.toCSV).left.get mustBe LarParsingError(0, List("Rate Spread is not numeric or NA"))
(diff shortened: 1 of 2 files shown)


parser/shared/src/test/scala/hmda/parser/fi/lar/LarCsvParserSpec.scala
```diff
@@ -46,13 +46,25 @@ class LarCsvParserSpec extends PropSpec with PropertyChecks with MustMatchers wi
   property("Fail parsing when NA is not properly spelled") {
     forAll(larGen) { lar =>
       val na = lar.copy(rateSpread = "na")
-      LarCsvParser(na.toCSV).left.get mustBe LarParsingError(0, List("Rate Spread is not numeric or NA"))
+      LarCsvParser(na.toCSV).left.get mustBe LarParsingError(0, List("Rate Spread is not double or NA"))
```
schbetsy: I think the word "double" is specific to programming. The message will probably be more useful to our users if it says "decimal" instead.

PR #1483, opened 2018-02-05
Use supervisor to find ActorRef

### LINKED ISSUE #1477: Files with known edits can rarely pass validation
This happens sporadically, I've encountered it in both my local setup and in production (yikes).
I don't have a firm repro, but it feels like it happens very soon after startup and only on the first upload of that startup (2/5: this is happening in production much more commonly than this theory would predict. It might have something to do with specific actors timing out or it might just be a plain race condition).

When this bug occurs, files with known edits will pass validation. I've seen it happen with files with only syntactical edits and files with only quality/macro edits. I can't confirm whether submission is prohibited, but I will the next time I encounter the bug.

My gut says there is a race condition somewhere such that no validation checks are run on that upload, thus resulting in a file without edits. This is rampant speculation, though.

nickgrippin:
Closes #1477

### DIFF
--- a/persistence/src/main/scala/hmda/persistence/processing/SubmissionManager.scala
+++ b/persistence/src/main/scala/hmda/persistence/processing/SubmissionManager.scala
@@ -149,7 +149,9 @@ object SubmissionManager {
         case SubmissionFSM.name => sender() ! submissionFSM
         case HmdaRawFile.name => sender() ! submissionUpload
         case HmdaFileParser.name => sender() ! submissionParser
-        case HmdaFileValidator.name => submissionValidator.map(sender() ! _)
+        case HmdaFileValidator.name =>
+          val replyTo = sender()
+          submissionValidator.map(replyTo ! _)
         case SubmissionLarStats.name => sender() ! submissionLarStats
       }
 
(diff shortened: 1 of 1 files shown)


persistence/src/main/scala/hmda/persistence/processing/SubmissionManager.scala
```diff
@@ -151,7 +149,7 @@ object SubmissionManager {
         case SubmissionFSM.name => sender() ! submissionFSM
         case HmdaRawFile.name => sender() ! submissionUpload
         case HmdaFileParser.name => sender() ! submissionParser
-        case HmdaFileValidator.name => sender() ! submissionValidator
+        case HmdaFileValidator.name => submissionValidator.map(sender() ! _)
```
jmarin: This closes over the sender in a future, and can have pretty bad consequences and cause a shared mutable state bug or race conditions. Easy change: 

```scala
case HmdaFileValidator.name =>
  val replyTo = sender()
  submissionValidator.map(client ! _)
```

For more information on this, see this [article](http://helenaedelson.com/?p=879)

PR #1486, opened 2018-02-06
V2 ts parser

### LINKED ISSUE #1485: Transmittal Sheet Parser
Parse Transmittal Sheet records according to the [Filing Instruction Guide](https://www.consumerfinance.gov/data-research/hmda/static/for-filers/2018/2018-HMDA-FIG.pdf)

jmarin:
Closes #1485

### DIFF
--- a/src/main/scala/hmda/parser/ParserErrorModel.scala
+++ b/src/main/scala/hmda/parser/ParserErrorModel.scala
@@ -1,11 +1,6 @@
 package hmda.parser
 
-import com.typesafe.config.ConfigFactory
-
 object ParserErrorModel {
-
-  val config = ConfigFactory.load()
-  val expectedLegth = config.getInt("hmda.filing.ts.length")
 
   def notAnInteger(fieldName: String) = s"$fieldName is not an integer"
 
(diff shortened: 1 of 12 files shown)


src/main/scala/hmda/parser/ParserErrorModel.scala
```diff
@@ -0,0 +1,22 @@
...
+  val config = ConfigFactory.load()
+  val expectedLegth = config.getInt("hmda.filing.ts.length")
+
+  def notAnInteger(fieldName: String) = s"$fieldName is not an integer"
+
+  trait ParserValidationError {
+    def errorMessage: String
+  }
+
+  case class IncorrectNumberOfFields(length: Int)
```
nickgrippin: I'm just curious, why is this error modeled as a case class and the others as case object?  

Is this object specific to the TS length?  It seems like it can be reused for the LAR, with some changes to the config

jmarin: This needs to be a case class because it takes a constructor parameter. Case objects (or traits) can't take parameters, compilation will fail.

PR #1513, opened 2018-02-22
Trim check digit csv lines before checking length and validity

### LINKED ISSUE #1509: Check Digit: CSV upload breaks on Windows-style line endings
Users have reported receiving a blank output file when uploading a csv file to the check digit batch feature

@wpears

wpears:
Closes #1509


jmarin:
Confirmed that this works with `dos` line carriage returns in the file. 
Thanks @wpears!

PR #1532, opened 2018-03-09
LAR Field Mapping

### LINKED ISSUE #1496: Provide friendly field names (LAR)
The LAR 2018 model needs friendly field names for mapping to internal domain model
Depends on #1443

nickgrippin:
For fields that hold an `Option`, I wasn't sure if we wanted to return the option itself or handle the `getOrElse` logic in the mapping.  I opted for returning the Option and letting the function caller decide how to handle a missing field, but let me know if there's a reason to code it another way.

Also, some field names don't exactly match up to the FIG, since they can be rather word-heavy.

Closes #1496


jmarin:
hey @keileyo when you get a chance can you review this? The file to look at is `LarFieldMapping.scala`, and it contains the user friendly names for the `LAR` fields. Coding wise it looks good, just want to make sure it looks good on your end. Thanks

keileyo:
@jmarin, sorry i missed this. just reviewed, looks good to me.

PR #1565, opened 2018-03-30
Remove Modified LAR from markdown API documentation

schbetsy:


jmarin:
This removes the documentation from the project, but does not affect the `gh-pages` branch. The document to edit is [this one](https://github.com/cfpb/hmda-platform/blob/gh-pages/institutions/index.md). The change required in that document is to remove the section regarding modified lar.  The PR affecting edits to that document should go against `gh-pages` directly. It's probably easiest to open another PR for that branch

PR #1575, opened 2018-04-05
Update 2018_File_Spec_Table2.csv

jmarin:
@keileyo If I'm reading the [FIG]() correctly, both ethnicity and race 1 for applicant/co-applicant can be empty under some circumstances. 
The paragraph in question is as follows (for Ethnicity, similar for Race):

```
If the Applicant or Borrower did not 
select any ethnicity(ies), but only 
provided ethnicity(ies) in the Ethnicity 
of Applicant or Borrower: Free Form 
Text Field for Other Hispanic or Latino, 
either leave this this data field blank or 
enter Code 14
```

This updates those field descriptions


keileyo:
Thanks for catching that, you are right. Looks good - merging.

PR #1591, opened 2018-04-19
Filers MSA/MD endpoint

### LINKED ISSUE #1570: New disclosure report endpoints
_This is just the start of the conversation about these endpoints. If you have other thoughts, please say so._

The flow to get to a disclosure report looks like:

Select a year > Select an institution > Select an MSA/MD > Select a report

The new [HMDA filers](https://github.com/cfpb/hmda-platform/blob/master/Documents/public-api.md#hmda-filers) endpoint covers the first 2 steps, "select a year" and "select an institution".

The last 2 steps require new endpoints.

---

### 1. Select an MSA/MD

This endpoint should provide a list of MSA/MDs for a given filer in a given year.

#### Possible URL

`/filers/<period>/<institutionId>/msaMds`

#### Proposed response

```json
{
  "year": "2017",
  "institution": {
    "name": "Bank 0",
    "id": "0",
    "respondentId": "Bank0_RID"
  },
  "msaMds": [
    {
      "id": "1234",
      "name": "Friendly MSA Name 1"
    },
    {
      "id": "2345",
      "name": "Friendly MSA Name 2"
    }
  ]
}
```

---

### ~~2. Select a report~~

~~This endpoint should return a list of reports for a given filer in a given year in a given MSA. _This assumes that it's possible that not all reports are generated for those combinations._~~

#### ~~Possible URL~~

~~`/filers/<period>/<institutionId>/msaMds/<msaMdId>/reports`~~

#### ~~Proposed response~~

```json
{
  "year": "2017",
  "institution": {
    "name": "Bank 0",
    "id": "0",
    "respondentId": "Bank0_RID"
  },
  "msaMd": {
      "id": "1234",
      "name": "Friendly MSA Name 1"
  },
  "reports": [
    {
      "id": "1",
      "name": "Applications by Tract"
    },
    {
      "id": "2",
      "name": "Loans Purchased by Tract"
    }
  ]
}
```
---

__Just as a side note__: The flow to get to an aggregate report is very different than the flow to get to a disclosure report. The aggregate report flow looks like:

Select a year > Select a state > Select an MSA/MD > Select a report

So, when that time comes, one or more endpoints will be required for that as well. I just wanted to point it out because the new endpoint(s) for aggregate would likely have a different starting path than `filers/`.

nickgrippin:
Closes #1570

### DIFF
--- a/api/src/main/scala/hmda/api/http/public/HmdaFilerPaths.scala
+++ b/api/src/main/scala/hmda/api/http/public/HmdaFilerPaths.scala
@@ -61,48 +61,51 @@ trait HmdaFilerPaths extends HmdaCustomDirectives with HmdaFilerProtocol with Ap
             sendResponse(filersF, uri)
           }
         }
-      } ~
-      path("filers" / Segment / Segment / "msaMds") { (period, instId) =>
-        encodeResponse {
-          timedGet { uri =>
-            val fSubmissions = (supervisor ? FindSubmissions(SubmissionPersistence.name, instId, period)).mapTo[ActorRef]
-            val hmdaFilerPersistenceF = (supervisor ? FindHmdaFilerPersistence(HmdaFilerPersistence.name)).mapTo[ActorRef]
+      }
+  }
 
-            val fMsas = for {
-              subActor <- fSubmissions
-              submissions <- (subActor ? GetState).mapTo[Seq[Submission]]
+  def hmdaFilerMsasPath(supervisor: ActorRef) = {
+    path("filers" / Segment / Segment / "msaMds") { (period, instId) =>
+      encodeResponse {
+        timedGet { uri =>
+          val fSubmissions = (supervisor ? FindSubmissions(SubmissionPersistence.name, instId, period)).mapTo[ActorRef]
+          val hmdaFilerPersistenceF = (supervisor ? FindHmdaFilerPersistence(HmdaFilerPersistence.name)).mapTo[ActorRef]
 
-              manager <- (supervisor ? FindProcessingActor(SubmissionManager.name, getLatestSignedSub(submissions))).mapTo[ActorRef]
-              larStats <- (manager ? GetActorRef(SubmissionLarStats.name)).mapTo[ActorRef]
-              stats <- (larStats ? FindIrsStats(getLatestSignedSub(submissions))).mapTo[Seq[Msa]]
+          val fMsas = for {
+            subActor <- fSubmissions
+            submissions <- (subActor ? GetState).mapTo[Seq[Submission]]
 
-              a <- hmdaFilerPersistenceF
-              filer <- (a ? FindHmdaFiler(instId)).mapTo[Option[HmdaFiler]]
-            } yield (stats, filer)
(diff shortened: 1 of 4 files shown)


api/src/main/scala/hmda/api/http/public/HmdaFilerPaths.scala
```diff
@@ -54,6 +61,47 @@ trait HmdaFilerPaths extends HmdaCustomDirectives with HmdaFilerProtocol with Ap
             sendResponse(filersF, uri)
           }
         }
+      } ~
+      path("filers" / Segment / Segment / "msaMds") { (period, instId) =>
```
schbetsy: Having this as a separate, named path would make this easier to read and maintain. e.g. `def hmdaFilerMsasPath = `

api/src/main/scala/hmda/api/http/public/HmdaFilerPaths.scala
```diff
@@ -54,6 +61,47 @@ trait HmdaFilerPaths extends HmdaCustomDirectives with HmdaFilerProtocol with Ap
...
+
+            onComplete(fMsas) {
+              case Success((msas, instOpt)) =>
+                val msasJson = msas.filterNot(_.id == "NA").map(msa => s"""{"id":"${msa.id}","name":"${msa.name}"}""").mkString("[", ",", "]")
+                val inst = instOpt.getOrElse(HmdaFiler("", "", "", ""))
+                if (inst.institutionId == "") {
+                  complete(ToResponseMarshallable(StatusCodes.NotFound -> s"Unable to find institution $instId"))
+                } else {
+                  val response =
+                    s"""
```
schbetsy: What's the reasoning for manipulating raw JSON here instead of having a Scala object with a JSON protocol (similar to the `HmdaFiler` object above)?

nickgrippin: The `HmdaFiler` API has a separate actor/persistence model that stores and retrieves specific objects.  The Filer MSA endpoint aggregates information from three different sources, and so I opted for creating the JSON in place.

api/src/main/scala/hmda/api/http/public/HmdaFilerPaths.scala
```diff
@@ -57,6 +64,50 @@ trait HmdaFilerPaths extends HmdaCustomDirectives with HmdaFilerProtocol with Ap
...
+                complete(ToResponseMarshallable(StatusCodes.NotFound -> s"Unable to find institution $instId"))
+              } else {
+                val response =
+                  s"""
+                       |{
+                       |  "year": "$period",
+                       |  "institution": {"name":"${inst.name}", "id":"${inst.institutionId}", "respondentId": "${inst.respondentId}"},
+                       |  "msaMds": $msasJson
+                       |}
+                   """.stripMargin.parseJson
```
jmarin: I'm going to merge this as is, but I don't think this is a good style for producing JSON output. Moving forward we should keep the same conventions that we have used in the project, by serializing/deserializing to case classes.

PR #1614, opened 2018-05-09
2018 Institutions Data Model Specs

colleenms:
Specifications for the 2018 Institutions Data Model


keileyo:
Looking good - just a couple of things:

- Tax ID and 2017 Institution ID should be alphanumeric bc they both have or can have a dash
- I get where we are going with "can't be null" for LEI because it is the primary institution identifier, but won't we have a lot of null LEIs, especially before we are able to collect them & for the institutions that are recognized in the inst data base, but are not expected filers. let's check w/ @jmarin on that

jmarin:
@colleenms Can you put this file in [this](https://github.com/cfpb/hmda-platform/tree/master/docs/v2/spec) folder please?. Just want to keep things organized. Thanks

keileyo:
@colleenms - based on @jmarin's response, just go ahead and update the LEI null to a "Yes" and then we'll be good to merge.

PR #1621, opened 2018-05-18
Institutions model

### LINKED ISSUE #1615: Implement Institutions Data Model
Version 2, including [LEI](https://www.gleif.org/en/about-lei/introducing-the-legal-entity-identifier-lei)

### LINKED ISSUE #1623: Can't test Akka Typed Persistence + Akka HTTP
The first requires mixing in the `akka.testkit.typed.scaladsl.ActorTestkit`, while the second uses `akka.http.scaladsl.testkit.ScalatestRouteTest`. Both bring an implicit `ActorSystem` into scope, with the same variable name (system), which causes a type conflict. The problem is that the first one is a typed `ActorSystem`, while the second is untyped. The first one is also final (`ActorTestkit.scala` in package `akka.testkit.typed.scaladsl`, line 83) so can't be overriden.

### LINKED ISSUE #1624: Add Admin Route tests
Possible once #1623 is solved

### LINKED ISSUE #1640: Large Files Fail to Parse for v2
When uploading larger files (50-100 lars) to the `public/hmda/parse` endpoint the `v2-hmda-platform` returns: "Substream Source cannot be materialized more than once".

There don't seem to be any related logs even on `DEBUG`.

Further:

1. Smaller files (1-40 lars) upload and parse correctly.
2. Larger files do validate properly intermittently locally.
3. This issues occurs whether there are 3 or 1 instances of v2 running.

jmarin:
Closes #1623.
Closes #1624
Closes #1615 
Closes #1640


hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -0,0 +1,147 @@
...
+          val fCreated: Future[InstitutionCreated] = institutionPersistence ? (
+              ref => CreateInstitution(institution, ref))
+
+          onComplete(fCreated) {
+            case Success(InstitutionCreated(i)) =>
+              complete(ToResponseMarshallable(StatusCodes.Created -> i))
+            case Failure(error) =>
+              val errorResponse =
+                ErrorResponse(500, error.getLocalizedMessage, uri.path)
+              println(errorResponse)
```
nickgrippin: `println` looks like it was left in accidentally

hmda/src/main/scala/hmda/model/institution/Institution.scala
```diff
@@ -0,0 +1,37 @@
...
+
+case class Institution(
+    activityYear: Int,
+    LEI: Option[String],
+    agency: Option[Agency],
+    institutionType: Option[InstitutionType],
+    institutionId_2017: Option[String],
+    taxId: Option[String],
+    rssd: Option[String],
+    emailDomain: Option[String],
```
nickgrippin: Are we allowing only one email domain per institution?

jmarin: Actually, I don't know. 
@keileyo Should we allow several email domains per institution?

colleenms: yes, we should allow multiple domains per institution @keileyo

nickgrippin:
Not worrying about Travis, this looks good to me.

PR #1683, opened 2018-07-25
S3 Headers

nickgrippin:
Adds the `attachment` file type to the dynamic data publication process.

Shoutout to Kenneth for figuring out the correct headers!


publication/src/main/scala/hmda/publication/regulator/lar/RegulatorLarPublisher.scala
```diff
@@ -91,7 +91,7 @@ class RegulatorLarPublisher extends HmdaActor with LoanApplicationRegisterCassan
         publicBucket,
         s"$environment/dynamic-data/$fileName",
         ContentType(MediaTypes.`text/csv`, HttpCharsets.`UTF-8`),
-        S3Headers(ServerSideEncryption.AES256)
```
kgudel: Since this is being pushed to into our public repo we shouldn't need this header. However, it would be good if we could figure out how to include both headers. From this [documentation](https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/S3Headers.scala#L85) it looks like we can copy what `ServerSideEncryption.AES256` does by including `"x-amz-server-side-encryption" -> "AES256"` in our map.

PR #1829, opened 2018-08-28
Update 2018_Institution_Data_Model_Spec.csv

keileyo:
Updated RSSD to an integer & updated the "Can be null" field to indicate that the integer fields cannot be null (unknown values will be -1).


jmarin:
Can we add a field that includes the empty value for those fields that can be null?
Also, minor correction. The HMDA filer field type is Boolean, not List

keileyo:
I've updated the HMDA Filer flag to Boolean & not able to be null (this makes sense to me).

PR #1848, opened 2018-08-31
LAR Parsing Fix

### LINKED ISSUE #1843: Edit logic for V628-2, V635-2 and V635-4
Upon testing, some validity edits are flagging when they shouldn't:

1. V628-2 should only be flagged if Ethnicity: 2-5 are something other than 1, 11, 12, 13, 14, 2 or blank. Upon testing a file that has those 4 fields blank, the edit is incorrectly flagged. 

2. V635-2 is similar, the Race: 2-5 fields should be either a list of numbers or blank. Upon testing a file that has those 4 fields blank, the edit is incorrectly flagged.  

3.  V635-4 is another similar one. If Race: 1 = 6 or 7, then then Race: 2-5 should be blank. Upon testing a file where Race: 1 = 6 and Race: 2-5 are blank, the edit is being incorrectly flagged.  This should be very similar to edit V475 from 2017. 

@nickgrippin please look into these and let me know if you have any questions.

nickgrippin:
Closes #1843

### DIFF
--- a/hmda/src/main/scala/hmda/parser/filing/lar/LarParser.scala
+++ b/hmda/src/main/scala/hmda/parser/filing/lar/LarParser.scala
@@ -84,7 +84,7 @@ trait LarParser {
       parserValidationError: ParserValidationError)
     : LarParserValidationResult[A] = {
     if (value == "") {
-      larCodeEnum.valueOf(-1).validNel
+      larCodeEnum.valueOf(0).validNel
     } else {
       validateLarCode(larCodeEnum, value, parserValidationError)
     }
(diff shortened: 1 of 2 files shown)


nickgrippin:
Note:  The only Enum that does not follow the pattern of 0 => empty is the [PurchaserType](https://github.com/cfpb/hmda-platform/blob/master/hmda/src/main/scala/hmda/model/filing/lar/enums/PurchaserEnum.scala#L10).  This PR means that an empty string in this field will map to a `PurchaserTypeNotApplicable`, which I don't think is the correct behavior.  This could be fixed by having that enum map to a different value (which would seem to be `10`, following the pattern of the other enums).

jmarin:
Please fix Travis failures

PR #1910, opened 2018-09-05
Prevent TS Parser from stripping trailing empty values

nickgrippin:


nickgrippin:
This PR fixes 2 sporadic test failures, both caused by the Generator's `sample()` method returning `None`.  

An empty TS CSV will incorrectly parse, as the `split()` method strips trailing empty segments by default.  

The other fix switches the ordering of the bad LAR data with the 10 generated good LARs, so that the line number of the error message is not dependent on the result of `sample()`.

PR #2042, opened 2018-09-26
Email table update on InstitutionModified

### LINKED ISSUE #1998: Update on institution email doesn't get projected properly
Changes are done in the journal, but don't get projected properly to email domains table. Probably because of mismatch in primary key / id of that table.

nickgrippin:
Closes #1998

### DIFF
--- a/institutions-api/src/main/scala/hmda/institution/projection/InstitutionDBProjection.scala
+++ b/institutions-api/src/main/scala/hmda/institution/projection/InstitutionDBProjection.scala
@@ -13,6 +13,7 @@ import hmda.messages.institution.InstitutionEvents.{
 import hmda.projection.ResumableProjection
 import hmda.query.DbConfiguration._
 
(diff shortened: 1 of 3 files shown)


institutions-api/src/main/scala/hmda/institution/projection/InstitutionDBProjection.scala
```diff
@@ -30,6 +31,8 @@ object InstitutionDBProjection
   implicit val institutionEmailsRepository = new InstitutionEmailsRepository(
     dbConfig)
 
+  implicit val ec: ExecutionContext = ExecutionContext.global
```
jmarin: This could potentially produce thread starvation under load, but until we have measurements it's hard to tell. I wouldn't change it for now, but let's try to keep in mind that if we see performance issues or errors with this approach it might be because of that issue.

PR #2047, opened 2018-09-28
Change Akka remote configuration to use artery

### LINKED ISSUE #1666: Explore using Artery for message transport

jmarin:
Closes #1666


nickgrippin:
The documentation states that the TCP based transport is not ready for use in production yet.  Could this potentially cause issues if we move to production with this library before it is complete?

jmarin:
Yeah, it's a potential issue. I will try to do some testing with the Aeron transport, and see if we can use that

PR #2097, opened 2018-10-19
Filing submission modified

### LINKED ISSUE #2094: Add SubmissionModifed to Filing Persistence
Add `ModifySubmission` logic to [`FilingPersistence`](https://github.com/cfpb/hmda-platform/blob/master/hmda/src/main/scala/hmda/persistence/filing/FilingPersistence.scala#L66)

### LINKED ISSUE #2102: GetLatestSubmission should return the newest Submission
The `GetLatestSubmission` method in the `FilingPersistence` actor should return the `Submission` with the highest sequence number.

jmarin:
Closes #2094
Closes #2102


hmda/src/main/scala/hmda/persistence/filing/FilingState.scala
```diff
@@ -24,6 +25,14 @@ case class FilingState(filing: Filing = Filing(),
         } else {
           FilingState(this.filing, submission :: submissions)
         }
+      case SubmissionUpdated(updated) =>
+        if (submissions.map(_.id).contains(updated.id)) {
+          val updatedList = updated :: submissions.filterNot(s =>
```
nickgrippin: This seems like it could mess with the behavior of `GetLatestSubmission` in the `FilingPersistence` actor.  It would put the most recently updated Submission at the head of the list, which could cause problems if someone creates a new submission, and an old one finishes uploading/validating later.

PR #2113, opened 2018-10-28
Remove specific docker images by name

omnipresent07:
This pull request removes the docker images after they've been pushed to docker hub. 

The final merge request should have few commands removed. These commands were added for debugging.

### DIFF
--- a/institutions-api/Jenkinsfile
+++ b/institutions-api/Jenkinsfile
@@ -26,7 +26,7 @@ volumes: [
               sh "docker tag ${env.DOCKER_HUB_USER}/institutions-api ${env.DOCKER_HUB_USER}/institutions-api:${gitBranch}"
               sh "docker login -u ${env.DOCKER_HUB_USER} -p ${env.DOCKER_HUB_PASSWORD} "
               sh "docker push ${env.DOCKER_HUB_USER}/institutions-api:${gitBranch}"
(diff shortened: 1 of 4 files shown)


institutions-api/Jenkinsfile
```diff
@@ -26,7 +26,9 @@ volumes: [
               sh "docker tag ${env.DOCKER_HUB_USER}/institutions-api ${env.DOCKER_HUB_USER}/institutions-api:${gitBranch}"
               sh "docker login -u ${env.DOCKER_HUB_USER} -p ${env.DOCKER_HUB_PASSWORD} "
               sh "docker push ${env.DOCKER_HUB_USER}/institutions-api:${gitBranch}"
-              sh "docker rmi -f ${env.DOCKER_HUB_USER}/institutions-api:${gitBranch}"
+              sh "docker images"
+              sh "docker images | grep 'institutions-api' | awk '{print \$3}' | uniq"
+              sh "docker rmi -f \$(docker images | grep 'institutions-api' | awk '{print \$3}' | uniq)"
```
kgudel: How would this respond if there was a running `institutions-api` from kubernetes on the node where this jenkins pipeline is run?

omnipresent07: We're deleting the `image` and not the `container`. If an `institutions-api` container was running, it wouldn't get affected by deleting the `image`.

PR #2118, opened 2018-10-30
WIP: Fix theme-provider Dockerfile Path

kgudel:


jenkins/seed.groovy
```diff
@@ -17,6 +17,8 @@ projects.each { project ->
                 repoOwner('cfpb')
                 repository(project.repo)
                 scanCredentialsId('github')
+                buildForkPRHead(buildForkPRHead = true)
+                buildForkPRMerge(buildForkPRMerge = false)
```
omnipresent07: I believe this should be `buildForkPRHead(true)` as opposed to `buildForkPRHead(buildForkPRHead = true)`

PR #2119, opened 2018-10-30
Stream check-digit response

### LINKED ISSUE #1667: Check Digit: Large files generate invalid results
For the check digit batch upload, large files (greater than ~6k records or 150 kb) fail to generate expected results. A user reported that the tool generated a file with truncated and blank ULI records.

nickgrippin:
Switches the response of the file upload, non-csv endpoints to stream a response back instead of creating a large JSON object.  This will help avoid the timeouts and allow larger files to be checked and validated.

Closes #1667

### DIFF
--- a/check-digit/src/main/scala/hmda/uli/api/http/ULIHttpApi.scala
+++ b/check-digit/src/main/scala/hmda/uli/api/http/ULIHttpApi.scala
@@ -1,5 +1,6 @@
 package hmda.uli.api.http
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.common.{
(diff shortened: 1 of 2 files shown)


check-digit/src/main/scala/hmda/uli/api/http/ULIHttpApi.scala
```diff
@@ -109,21 +115,25 @@ trait ULIHttpApi extends HmdaTimeDirectives {
...
-                          StatusCodes.BadRequest -> errorResponse))
-                    }
+                    val uStart = ByteString.fromString("{\"ulis\":[")
+                    val uMiddle = ByteString.fromString(",")
+                    val uEnd = ByteString.fromString("]}")
+                    val withUliWrapper: Flow[ByteString, ByteString, NotUsed] =
+                      Flow[ByteString]
+                        .intersperse(uStart, uMiddle, uEnd)
+                        .map(s => {
+                          println(s.utf8String)
```
jmarin: We don't need this `println`, right?

PR #2139, opened 2018-11-02
Submission Serializer

### LINKED ISSUE #2136: 504 timeout on GET for latest submission
When trying to make a call to [`/v2/filing/institutions/B90YWS6AFX2LGWOXJ1LD/filings/2018/submissions/latest`](https://github.com/cfpb/hmda-platform/blob/8222f7e8e5f3089e87eb210fc977bec0be8b0d98/docs/v2/api/filing-api.md#latest-submission) I'm seeing the timeout issue that we've seen on other endpoints, eg #2128 and #2130.

nickgrippin:
Closes #2136


common/src/main/scala/hmda/serialization/submission/SubmissionSerializer.scala
```diff
@@ -0,0 +1,35 @@
+package hmda.serialization.submission
+
+import java.io.NotSerializableException
+
+import akka.serialization.SerializerWithStringManifest
+import hmda.model.filing.submission.Submission
+import hmda.persistence.serialization.submission.SubmissionMessage
+import hmda.serialization.submission.SubmissionProtobufConverter._
+
+class SubmissionSerializer extends SerializerWithStringManifest {
```
jmarin: This class is not necessary. The `SubmissionCommandsSerializer` already has support for serializing `Submission` objects. All that is required is the following line in the `serialization.conf` file: 

`hmda.model.filing.submission = submission-commands`

PR #2140, opened 2018-11-02
prune images

omnipresent07:
We're running version `1.30` in Jenkins. Beginning with `1.25+`, docker provides the `image prune` and `system prune` commands. 

https://docs.docker.com/engine/reference/commandline/image_prune/

`image prune` should remove all dangling images. It will still not remove tagged but unused (not referenced by a container) images. To assess this I've added `docker images` and `docker ps -a` commands so that we can assess how many unused  images are being accumulated after each build.


kgudel:
Looks good. I also like how this cleans up those lines.

PR #2153, opened 2018-11-07
include error response codec

### LINKED ISSUE #2151: JSON response not serialized for 400 on submissions
When trying to upload a file for a submission that doesn't yet exist, a correct `400` response is sent, however, the JSON data is not serialized. 

Similar to #2132 

```
{
    "httpStatus": 400,
    "message": "Submission 54930084UKLVMY22DS16-2019-5 not available for upload",
    "path": {
        "Slash": {
            "tail": {
                "Segment": {
                    "head": "institutions",
                    "tail": {
                        "Slash": {
                            "tail": {
                                "Segment": {
                                    "head": "54930084UKLVMY22DS16",
                                    "tail": {
                                        "Slash": {
                                            "tail": {
                                                "Segment": {
                                                    "head": "filings",
                                                    "tail": {
                                                        "Slash": {
                                                            "tail": {
                                                                "Segment": {
                                                                    "head": "2019",
                                                                    "tail": {
                                                                        "Slash": {
                                                                            "tail": {
                                                                                "Segment": {
                                                                                    "head": "submissions",
                                                                                    "tail": {
                                                                                        "Slash": {
                                                                                            "tail": {
                                                                                                "Segment": {
                                                                                                    "head": "5",
                                                                                                    "tail": {
                                                                                                        "Empty": {}
                                                                                                    }
                                                                                                }
                                                                                            }
                                                                                        }
                                                                                    }
                                                                                }
                                                                            }
                                                                        }
                                                                    }
                                                                }
                                                            }
                                                        }
                                                    }
                                                }
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

omnipresent07:
Fix #2151 

Same issue was happening in `ULIHttpApi`. This takes care of that too.


hmda/src/main/scala/hmda/api/http/filing/submissions/UploadHttpApi.scala
```diff
@@ -96,30 +98,31 @@ trait UploadHttpApi extends HmdaTimeDirectives {
...
                   } else {
-                    submissionNotAvailable(submissionId, uri)
+                    val errorResponse = ErrorResponse(
+                      404,
+                      s"Submissidson ${submissionId.toString} not available for upload",
+                      uri.path)
+                    complete(
+                      ToResponseMarshallable(
+                        StatusCodes.NotFound -> errorResponse)
+                    )
```
jmarin: I think this is worse than what we had before. The point of the `submissionNotAvailable` function is to extract this common boilerplate code to an object that can be brought in scope (`FilingResponseUtils`). It's possible that it needs refactoring or an additional method to account for this `404`, but I don't think introducing this back makes the code more readable or maintainable. To use this object in the check digit project as well, I would refactor `FilingResponseUtils` to the `common` project so it can be used from both places.

omnipresent07:
The recent refactor extracts `FilingResponseUtils` to common, makes `submissionNotAvailable` part of it, and changes signature for `failedResponse` to accept `StatusCodes.<statuscode>` as a parameter in an effort to make it more generic and respond to `400` or `500`

jmarin:
Travis is not happy

PR #2157, opened 2018-11-07
Include all emails by LEI

### LINKED ISSUE #2154: Institutions-API findByEmail method doesn't return all emails
The `findByEmail` method in `InstitutionComponent` will only return the email passed into the method.  If an institution has multiple entries in the email repository, it won't return the additional emails.

nickgrippin:
Closes #2154

### DIFF
--- a/institutions-api/src/main/scala/hmda/institution/query/InstitutionComponent.scala
+++ b/institutions-api/src/main/scala/hmda/institution/query/InstitutionComponent.scala
@@ -154,11 +154,11 @@ trait InstitutionComponent {
       institutions <- db.run(institutionQuery(leis).result)
       emails <- db.run(emailTotalQuery(leis).result)
     } yield {
-      institutions.map(institution => {
+      institutions.map { institution =>
         val filteredEmails =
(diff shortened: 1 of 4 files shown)


jmarin:
Before I review, can you check what's up with the CI server?
Thanks

institutions-api/src/main/scala/hmda/institution/query/InstitutionComponent.scala
```diff
@@ -139,21 +139,26 @@ trait InstitutionComponent {
...
+      emailEntities <- db.run(emailSingleQuery.result)
       leis = emailEntities.map(_.lei)
       institutions <- db.run(institutionQuery(leis).result)
+      emails <- db.run(emailTotalQuery(leis).result)
     } yield {
-      institutions.map(
-        institution =>
-          InstitutionConverter.convert(institution,
-                                       emailEntities.map(_.emailDomain)))
+      institutions.map(institution => {
```
jmarin: I think the extra `{` is unnecessary

nickgrippin:
Addresses #2177

PR #2180, opened 2018-11-09
return 400 for duplicate lei

### LINKED ISSUE #2178: Update message for duplicate LEIs
Update the message returned when a record with LEI that is duplicative of an existing LEI is added to the institutions data. @jmarin confirmed that the underlying institutions database will not be updated in this case, but we would like to return a message that indicates and error &  that the LEI is a duplicate and can be served to the HMDA Help app UI. 

I think the message itself can just be a simple "Error - Duplicate LEI"

omnipresent07:
This should fix #2178 from the backend but most likely needs a change on the front end as well @awolfe76

### DIFF
--- a/hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
+++ b/hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
@@ -14,6 +14,7 @@ import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
 import hmda.api.http.directives.HmdaTimeDirectives
 import hmda.api.http.codec.institution.InstitutionCodec._
 import hmda.api.http.model.ErrorResponse
+import hmda.api.http.codec.ErrorResponseCodec._
 import hmda.api.http.model.admin.InstitutionDeletedResponse
 import hmda.persistence.institution.InstitutionPersistence
 import io.circe.generic.auto._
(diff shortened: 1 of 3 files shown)


hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -50,19 +51,35 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
...
-                  ErrorResponse(500, error.getLocalizedMessage, uri.path)
+            onComplete(fInstitution) {
+              case Success(Some(_)) =>
+                val errorResponse = ErrorResponse(
+                  400,
+                  s"Institution ${institution.LEI} already exists",
+                  uri.path)
                 complete(ToResponseMarshallable(
-                  StatusCodes.InternalServerError -> errorResponse))
+                  StatusCodes.BadRequest -> errorResponse))
```
jmarin: You can use the utilities in `FilingResponseUtils` to make this look a bit cleaner

hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -50,19 +51,35 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
...
+                  : Future[InstitutionCreated] = institutionPersistence ? (
+                    ref => CreateInstitution(institution, ref))
+                onComplete(fCreated) {
+                  case Success(InstitutionCreated(i)) =>
+                    complete(ToResponseMarshallable(StatusCodes.Created -> i))
+                  case Failure(error) =>
+                    val errorResponse =
+                      ErrorResponse(500, error.getLocalizedMessage, uri.path)
+                    complete(ToResponseMarshallable(
+                      StatusCodes.InternalServerError -> errorResponse))
```
jmarin: Same comment as above

Issue #1966, opened 2018-09-13
Quiet down logs (Cassandra, Kafka, Zookeeper client)


jmarin:
The application prints out too many log entries when interacting with `Cassandra`, `Kafka` and the `Zookeeper` client (INFO level). Configure `logback.xml` so that these don't show up by default

PR #2188, opened 2018-11-13
Quiet logs

omnipresent07:
Fixes https://github.com/cfpb/hmda-platform/issues/1966


hmda/src/main/resources/logback.xml
```diff
@@ -10,8 +10,8 @@
         <appender-ref ref="STDOUT" />
     </appender>
 
-    <root level="INFO">
+    <root level="${PLATFORM_LOG_LEVEL:-INFO}">
         <appender-ref ref="ASYNC" />
     </root>
-    <logger name="org.apache.zookeeper" level="WARN"/>
+    <logger name="org.apache.zookeeper" level="${ZOOKEEPER_LOG_LEVEL:-WARN}"/>
```
jmarin: Aren't we missing settings here to remove `Cassandra` and `Kafka` driver logs by default?

Issue #2213, opened 2018-11-15
Create SPI for V2


kgudel:
To include the SPI in a deployment of keycloak using helm we need to create a docker image that contains the the spi jar and the modify our `values.yaml` file with the following (the below sample code is what I used to get the keycloak spi test working and so the directories used in the init container and that container image will change):

Add another `extraInitContainers`:
```
 - name: spi-provider
      image: kgudel/spi-provider:latest
      imagePullPolicy: Always
      command:
        - sh
      args:
        - -c
        - |
          echo "Copying SPI..."
          cp -rv /providers/authenticator/hmda/target/keycloak-authenticator-hmda-4.2.1.Final.jar /spi
      volumeMounts:
        - name: spi
          mountPath: /spi
```
Add to `cli.custom`:
```
/subsystem=keycloak-server/spi=form-action:add()
/subsystem=keycloak-server/spi=form-action/provider=registration-institution-action:add(properties={institutionSearchUri => "${env.INSTITUTION_SEARCH_URI}", institutionSearchValidateSsl => "${env.INSTITUTION_SEARCH_VALIDATE_SSL}"}, enabled=true)

```
Add another `extraVolumes`:
```
- name: spi
  mountPath: /opt/jboss/keycloak/standalone/deployments
```

kgudel:
Keycloak also needs a mapper configured in the hmda2-api client in order to get the `lei` attribute.

You can set this up by doing the following:

1. `Open Admin Consol` -> `clients` -> `hmda2-api` -> `mappers` -> `create`
1. `Mapper Type`: `User Attribute`
1. `Name`: `lei`
1. `User Attribute`: `lei`
1. `Token Claim Name`: lei
1. `Claim JSON Type`: `string`

PR #2226, opened 2018-11-16
Add Keycloak SPI project

### LINKED ISSUE #2225: Build Keycloak Service Provider Interface
Migrate project from `hmda-platform-auth` into this repository, and provide mechanism to build and package the resulting `jar` file

jmarin:
Closes #2225


jmarin:
@kgudel I have added the `Keycloak` SPI code into a new directory within this project. It's still configured to build with `Maven` and is not integrated into the rest of the build pipeline (I think it's fine this way). I have also added a minimal `Jenkinsfile` that creates the `jar` file with the SPI code. The work here should be the necessary dependency for #2213.
As far as the code goes, it will need to be modified. Once we have tested that this builds and can be integrated with out `Keycloak` deployment pipeline we can introduce those changes (they should be fairly small)

auth/Jenkinsfile
```diff
@@ -0,0 +1,33 @@
...
+]) {
+   node('buildSbtPod') {
+     def repo = checkout scm
+     def gitCommit = repo.GIT_COMMIT
+     def gitBranch = repo.GIT_BRANCH
+     def shortGitCommit = "${gitCommit[0..10]}"
+
+     stage('Build and package Java Code') {
+       container('maven') {
+         sh "mvn -B clean package"
```
kgudel: This currently tries to build the base repo and fails. I changed it to `mvn -B -f ./auth/ clean package` and the `jar` built correctly. Also, in order to test it in keycloak we still need to package this jar into a docker container.

PR #2244, opened 2018-11-20
Fix packaging naming collision

jmarin:


omnipresent07:
https://github.cfpb.gov/HMDA-Operations/hmda-devops/issues/850 for reference

PR #2254, opened 2018-11-20
the public API still uses LEI instead of lei

awolfe76:
@kgudel - this should make the institutions checkbox clickable during registration ... the public API uses `LEI` vs `lei`.


jmarin:
I don't think this should be fixed in the frontend. The filing API returns `lei`, while the public API is returning `LEI`. We should make the two consistent. No changes in the frontend would be necessary then

PR #2260, opened 2018-11-21
Quality verify post

### LINKED ISSUE #2256: POST to /quality status incorrect
I've attempted this a few times with different files (`clean_file_100_rows_Bank0_syntax_validity.txt`, `clean_file_10_rows_Bank0_syntax_validity.txt`) and am seeing the `POST` to the `/quality` endpoint return the same status code, `11`. When quality are verified the status should get bumped to either `12: NO_MACRO_EDITS` (which is really `14:VALIDATED`) or `13:MACRO_EDITS` (but this doesn't apply until we implement macro edits).

---

#### Status is `11`
```json
{"id":{"lei":"B90YWS6AFX2LGWOXJ1LD","period":"2018","sequenceNumber":8},"status":{"code":11,"message":"Your data has quality edits that need to be reviewed.","description":"Your file has been uploaded, but the filing process may not proceed until edits are verified or the file is corrected and re-uploaded."},"start":1542743363542,"end":0,"fileName":"clean_file_100_rows_Bank0_syntax_validity.txt","receipt":""}
```

---

#### Status is still `11` but `verified` is `true`
```json
{"verified":true,"status":{"code":11,"message":"Your data has quality edits that need to be reviewed.","description":"Your file has been uploaded, but the filing process may not proceed until edits are verified or the file is corrected and re-uploaded."}}
```

jmarin:
Closes #2256

### DIFF
--- a/hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
+++ b/hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
@@ -197,13 +197,21 @@ object HmdaValidationError
       case VerifyQuality(submissionId, verified, replyTo) =>
         if (List(Quality.code, QualityErrors.code, Macro.code, MacroErrors.code)
               .contains(state.statusCode)) {
-          Effect.persist(QualityVerified(submissionId, verified)).thenRun { _ =>
-            if (state.macroVerified) {
-              val updatedStatus = Verified
-              updateSubmissionStatus(sharding, submissionId, updatedStatus, log)
-              replyTo ! QualityVerified(submissionId, verified)
+          Effect
+            .persist(
+              QualityVerified(submissionId,
+                              verified,
+                              SubmissionStatus.valueOf(state.statusCode)))
+            .thenRun { _ =>
+              if (state.macroVerified) {
+                val updatedStatus = Verified
+                updateSubmissionStatus(sharding,
+                                       submissionId,
+                                       updatedStatus,
+                                       log)
+                replyTo ! QualityVerified(submissionId, verified, updatedStatus)
+              }
             }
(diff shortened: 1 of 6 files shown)


hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -197,13 +197,21 @@ object HmdaValidationError
...
+                              verified,
+                              SubmissionStatus.valueOf(state.statusCode)))
+            .thenRun { _ =>
+              if (state.macroVerified) {
+                val updatedStatus = Verified
+                updateSubmissionStatus(sharding,
+                                       submissionId,
+                                       updatedStatus,
+                                       log)
+                replyTo ! QualityVerified(submissionId, verified, updatedStatus)
```
nickgrippin: Just want to make sure this behavior is correct.  This won't wait for the submission status to be updated, it fires off both requests (to update the submission status and the reply that the status is updated) simultaneously.  I'm completely fine with this solution, as it returns the expected result for the POST request.  But it is slightly different behavior than V1, when we waited for the status to be updated before sending a response.

jmarin: Trade-offs....I thought about it and decided that the code was going to get complicated, and we're doing this all over the place anyway (updating the status at the same time we reply). I'm OK with creating a ticket to suggest a change here, but this is probably fine for now.

nickgrippin: Sounds good!  It's definitely a trade-off and I agree with the design decision here to keep the code a bit simpler.

Issue #2246, opened 2018-11-20
Add test for edit details JSON encoder


jmarin:
Add the relevant generators and property based tests for proving correctness of `EditDetailsSummaryCodec`

PR #2265, opened 2018-11-22
EditDetailsSummaryCodec   Spec

omnipresent07:
#2246


omnipresent07:
Update docs https://github.com/cfpb/hmda-platform/blob/85329e2403849f7db9bc24c898c7685183da86f5/docs/v2/api/filing-api.md#edit-details

PR #2266, opened 2018-11-23
Refactor Institutions API configuration

nickgrippin:
Adds SSL configuration to the JDBC URL, and moves the username/password settings out of the official PostrgreSQL `db` config object.


kgudel:
Addresses #2272 

This corrects the db connection when paired with #2271 

LGTM

PR #2275, opened 2018-11-27
Remove tagging

jmarin:


nickgrippin:
We have tagging in `InstitutionPersistence` as well.  Are we leaving that in for now?

jmarin:
The problem is that, without it, we can't project events from the resumable projection, see [here](https://github.com/cfpb/hmda-platform/blob/master/common/src/main/scala/hmda/projection/ResumableProjection.scala#L52). This one has to stay for now, we would have to change the implementation. Using tagging like this is not a bad idea, I just think there are some settings in `Cassandra` that are preventing us from fully leveraging this capability

Issue #2277, opened 2018-11-28
Create spec for EditsHttpAPI


omnipresent07:
Endpoints:

*  institutions/:lei/filings/:period/submissions/:submissionId/edits
*  institutions/:lei/filings/:period/submissions/:submissionId/edits/csv
*  institutions/:lei/filings/:period/submissions/:submissionId/edits/edit

PR #2278, opened 2018-11-28
match Q613 and Q613-4

### LINKED ISSUE #2276: Calls to edits with name such as QXXX-XX fail
The regex [here](https://github.com/cfpb/hmda-platform/blob/master/hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala#L99) needs to be updated to account for edits such as Q615-1

omnipresent07:
Fix #2276 

From https://github.com/cfpb/hmda-platform/blob/master/common/src/main/resources/2018EditsDescriptions.txt I don't see `QXXX-XX` as mentioned in the bug. Please correct me if I'm wrong. 

Separate ticket opened https://github.com/cfpb/hmda-platform/issues/2277 to create spec for this class.

### DIFF
--- a/hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala
+++ b/hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala
@@ -96,7 +96,7 @@ trait EditsHttpApi extends HmdaTimeDirectives {
 
   //institutions/<institutionId>/filings/<period>/submissions/<submissionId>/edits/edit
   def editDetailsPath(oAuth2Authorization: OAuth2Authorization): Route = {
-    val editNameRegex: Regex = new Regex("""[SVQ]\d\d\d""")
+    val editNameRegex: Regex = new Regex("""[SVQ]\d\d\d(?:-\d)*""")
     path(
       "institutions" / Segment / "filings" / Segment / "submissions" / IntNumber / "edits" / editNameRegex) {
       (lei, period, seqNr, editName) =>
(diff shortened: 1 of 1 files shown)


hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala
```diff
@@ -96,7 +96,7 @@ trait EditsHttpApi extends HmdaTimeDirectives {
 
   //institutions/<institutionId>/filings/<period>/submissions/<submissionId>/edits/edit
   def editDetailsPath(oAuth2Authorization: OAuth2Authorization): Route = {
-    val editNameRegex: Regex = new Regex("""[SVQ]\d\d\d""")
+    val editNameRegex: Regex = new Regex("""[SVQ]\d\d\d(?:-\d)*""")
```
nickgrippin: This could probably be simplified to `(-\d)?`, we only want to match this pattern 0 or 1 times.

omnipresent07: `(?:)` forms a non-capturing group. Which, from my testing, seems to be necessary for Akka PatternMatcher DSL. without the `(?:)` we just get `-2` for `editName` when, for example, `Q613-2` get passed. I have however changed `*` to `?` to capture 0 or 1

nickgrippin: Nice, that sounds perfect! 👍

PR #2280, opened 2018-11-28
Memory config

### LINKED ISSUE #2279: Add JVM memory settings

jmarin:
Closes #2279

### DIFF
--- a/hmda/build.sbt
+++ b/hmda/build.sbt
@@ -17,3 +17,10 @@ dockerCommands :=
       Seq(Cmd("ENTRYPOINT", args.mkString(" ")))
     case v => Seq(v)
   }
+
+javaOptions in Universal ++= Seq(
+  "-J-Xms1G",
+  "-J-Xmx4G",
+  "-J-XX:+UnlockExperimentalVMOptions",
+  "-J-XX:+UseCGroupMemoryLimitForHeap"
+)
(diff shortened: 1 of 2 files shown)


hmda/build.sbt
```diff
@@ -17,3 +17,10 @@ dockerCommands :=
       Seq(Cmd("ENTRYPOINT", args.mkString(" ")))
     case v => Seq(v)
   }
+
+javaOptions in Universal ++= Seq(
+  "-J-Xms1G",
+  "-J-Xmx4G",
+  "-J-XX:+UnlockExperimentalVMOptions",
+  "-J-XX:+UseCGroupMemoryLimitForHeap"
```
nickgrippin: Will this work if `Xmx` is set as well?  I'm still looking through the documentation, but it looks like this is used in place of directly specifying max heap size.  I could be misunderstanding though.

jmarin: See documentation [here](https://doc.akka.io/docs/akka/2.5/additional/deploy.html). The problem as I understand it is that in versions of the JVM prior to 8u131 these settings didn't do what you would expect them to do, and a higher heap size setting would override whatever is set in `docker`. You could end up with containers that consume more memory than what is configured in `docker`. In any case, I'm only making explicit something that is already in place, since the default for the max heap size in these containers is 4GB

PR #2287, opened 2018-11-29
Pagination refactor

### LINKED ISSUE #2284: Edit Details endpoint doesn't paginate
Request to, eg,
`v2/filing/institutions/B90YWS6AFX2LGWOXJ1LD/filings/2018/submissions/54/edits/Q614?page=2`

results in a `504` with an `upstream request timeout`.

nickgrippin:
* Moves common pagination methods into the `WithPagination` trait
* Fix - uses existing logic in `PaginatedResponse` to calculate `take()` logic
* Simplifies EditDetails gathering by using existing `EditDetailsSummary` object

Closes #2284

### DIFF
--- a/hmda/src/main/resources/hmda.conf
+++ b/hmda/src/main/resources/hmda.conf
@@ -40,6 +40,8 @@ hmda {
     shardNumber = 100
   }
 
+  pageSize = 20
+
   loader {
     parallelism = 4
     institution {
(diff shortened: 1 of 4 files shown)


hmda/src/main/resources/hmda.conf
```diff
@@ -40,6 +40,8 @@ hmda {
     shardNumber = 100
   }
 
+  pageSize = 20
```
jmarin: I would add an environment variable here so that we can change this value at runtime without having to recompile

PR #2302, opened 2018-12-03
update realm file with default role and redirect URIs

awolfe76:
This updates the realm to include the correct `redirectUris` and `defaultRoles`.
Edit: It now also prevents access to the account management screens.

These diffs are hard to view because the export isn't consistent ... see inline comments.

### DIFF
--- a/kubernetes/keycloak/realm.json
+++ b/kubernetes/keycloak/realm.json
@@ -345,8 +345,6 @@
       "secret": "**********",
       "defaultRoles": ["hmda-user"],
       "redirectUris": [
-        "https://192.168.99.100/silent_renew.html",
-        "https://192.168.99.100/oidc-callback",
         "http://192.168.99.100/*",
         "http://localhost:3000/*",
         "/*",
--- a/kubernetes/keycloak/realm.json
+++ b/kubernetes/keycloak/realm.json
@@ -579,7 +579,7 @@
       "enabled": true,
       "clientAuthenticatorType": "client-secret",
       "secret": "**********",
-      "defaultRoles": ["view-profile", "manage-account"],
+      "defaultRoles": [],
       "redirectUris": ["/auth/realms/hmda2/account/*"],
       "webOrigins": [],
       "notBefore": 0,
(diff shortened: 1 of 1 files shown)


kubernetes/keycloak/realm.json
```diff
@@ -34,29 +34,572 @@
...
+  "clients": [
+    {
+      "id": "c749d73d-de71-42fa-aff3-44456890e908",
+      "clientId": "hmda2-api",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "**********",
+      "defaultRoles": ["hmda-user"],
+      "redirectUris": [
```
awolfe76: Most of these are to allow for local development, the most important one for our cluster environments is the relative path, `/*`.

kubernetes/keycloak/realm.json
```diff
@@ -34,29 +34,570 @@
...
+  ],
+  "clients": [
+    {
+      "id": "c749d73d-de71-42fa-aff3-44456890e908",
+      "clientId": "hmda2-api",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "**********",
+      "defaultRoles": ["hmda-user"],
```
awolfe76: This sets the `hmda-user` role as the default for the `hmda2-api` client.

kubernetes/keycloak/realm.json
```diff
@@ -34,29 +34,570 @@
...
+    {
+      "id": "63989b58-df41-4462-b339-055e842c1bb0",
+      "clientId": "account",
+      "name": "${client_account}",
+      "baseUrl": "/auth/realms/hmda2/account",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "**********",
+      "defaultRoles": [],
```
awolfe76: Removing the default roles in the `account` client prevents users from accessing the account mgmt screens.

awolfe76:
@omnipresent07 @kgudel - This should be ready for review.

omnipresent07:
This LGTM 👍

PR #2304, opened 2018-12-03
Allow pod-namespace to be set for cluster.bootstrap when using kubernetes-api

charandas:
While trying to run hmda v2 in our datacenter, I discovered that Akka ClusterBootstrap requires this property to be set for discovery to correctly take place, if one is not using `default` namespace for the deployment.

The `service-namespace` in the same file is not used when using `kubernetes-api`.

The [docs](https://developer.lightbend.com/docs/akka-management/current/bootstrap/kubernetes-api.html) themselves mention of the `pod-namespace` key, but also a `namespace` ENV var. That var did not work for me.


jmarin:
Thanks for your contribution @charandas!

PR #2316, opened 2018-12-06
 rR

PatrickHSI:


hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -213,7 +213,7 @@ object HmdaValidationError
         if (state.statusCode == Verified.code) {
           val timestamp = Instant.now().toEpochMilli
           val signed = SubmissionSigned(submissionId, timestamp, Signed)
-          if (state.qualityVerified && state.macroVerified) {
+          if ((state.qualityVerified && state.macroVerified) || state.statusCode == Verified.code) {
```
nickgrippin: We check `state.statusCode == Verified.code` in an `if` statement a few lines above, so this `if` statement will now always return true.

PatrickHSI: Well that's embarrassing , I'll rectify this asap  @nickgrippin .

PR #2321, opened 2018-12-06
WIP: V609

jmarin:


hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -319,6 +330,29 @@ object HmdaValidationError
...
+        x.collect {
+          case Left(errors) => errors
+          case Right(_)     => Nil
+        }
+      }
+      .zip(Source.fromIterator(() => Iterator.from(2)))
+      .map { x =>
+        x._1
+          .map { errors =>
+            PersistHmdaRowValidatedError(submissionId, x._2, errors, None)
```
nickgrippin: I understand that Edit Details aren't part of this PR (and that should probably be in another ticket), but it seems like it will be difficult to pull the relevant field information given the current structure of the Flow.

Issue #1753, opened 2018-08-13
2018 Edits: Q604


chynnakeys:
Edit #: Q604
Edit Type: Quality
Category: LAR
Data Fields: County; State 
Description:

_Instructional Text for Users_: Please verify the information below and update your file, if needed:

_Edit Logic:_ 1) The reported State and County are not a valid combination. If neither State nor County were reported NA, then the County must be located within the State.

Issue #1752, opened 2018-08-13
2018 Edits: Q603


chynnakeys:
Edit #: Q603
Edit Type: Quality
Category: LAR
Data Fields: County; Census Tract 
Description:

_Instructional Text for Users_: An invalid Census Tract was reported. Please review the information below and update your file accordingly:

_Edit Logic:_ 1) The County has a population of greater than 30,000 according to the most recent decennial census and was not reported NA; however Census Tract was reported NA.

Issue #1702, opened 2018-08-09
2018 Edits: V627


chynnakeys:
Edit #: V627
Edit Type: Validity
Category: LAR
Data Fields: Census Tract; County
Edit Description:

_Instructional Text for Users:_ An invalid Census Tract or County was provided. Please review the information below and update your file accordingly:

_Edit Logic:_ 1) If County and Census Tract are not reported NA, they must be a valid combination of information. The first five digits of the Census Tract must match the reported five digit County FIPS code.

nickgrippin:
On hold: Requires microservice communication

Issue #1701, opened 2018-08-09
2018 Edits: V626


chynnakeys:
Edit #: V626
Edit Type: Validity
Category: LAR
Data Fields: County
Edit Description:

_Instructional Text for Users:_ An invalid County was provided. Please review the information below and update your file accordingly:

_Edit Logic:_ 1) The required format for County is a five digit FIPS code or NA, and it cannot be left blank.

nickgrippin:
On hold: Requires microservice communication

PR #2327, opened 2018-12-07
Census Microservice

### LINKED ISSUE #2308: Census microservice
Something to keep in mind is that this microservice can be used for both edits and publication (there may be other use cases). But, the immediate focus is to complete the remaining edits.

- [x] Pull / create the datasource
- [x] Load datasource
- [x] Code required to complete the edits

We'll be able to use the check digit microservice (coming soon) as a template for communication between 2 services.

omnipresent07:
Closes #2308 , #1701 , #1700 , #1702 , #1753 , #1752 

Implement:

Syntactical 
```
V625-1|The required format for Census Tract is an eleven-digit number or NA, and it cannot be left blank.|Census Tract
V625-2|If Census Tract is not reported NA, then the number provided must be a valid census tract number defined by the U.S. Census Bureau.|Census Tract
V626|The required format for County is a five-digit FIPS code or NA, and it cannot be left blank.|County
V627|If County and Census Tract are not reported NA, they must be a valid combination of information. The first five digits of the Census Tract must match the reported five-digit County FIPS code.|Census Tract; County

```

Quality

```
Q603|The County has a population of greater than 30,000 according to the most recent decennial census and was not reported NA; however Census Tract was reported NA.|County; Census Tract
Q604|The reported State and County are not a valid combination. If neither State nor County were reported NA, then the County must be located within the State.|County; State

```


census-api/src/main/scala/hmda/census/records/CensusRecords.scala
```diff
@@ -0,0 +1,55 @@
+package hmda.census.records
+
+import com.typesafe.config.ConfigFactory
+import hmda.model.census.Census
+import hmda.model.ResourceUtils._
+
+trait CensusRecords {
```
jmarin: It's a matter of preference, but this should probably be an `object`, not a `trait`. That way it becomes a singleton, and you probably don't want to extend / override the functions defined here anyway

hmda/src/main/scala/hmda/validation/rules/lar/validity/V625_1.scala
```diff
@@ -0,0 +1,19 @@
...
+import hmda.validation.rules.EditCheck
+
+object V625_1 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V625-1"
+
+  override def parent: String = "V625"
+
+  override def apply(lar: LoanApplicationRegister): ValidationResult = {
+    (lar.geography.tract is numeric and (lar.geography.tract.size is equalTo(
+      11))) or (lar.geography.tract.toLowerCase is equalTo("na"))
```
jmarin: Is it `na` or `NA`?

omnipresent07:
Feedback addressed.

PR #2356, opened 2018-12-17
Institutions API Kafka migration

### LINKED ISSUE #2273: Institutions API stops projecting after startup
After the Institutions API has read through the Cassandra journal at startup, any new Institution events will be ignored until the API is restarted again.  This has been observed in HMDA4 when a user creates an Institution record in the HMDA Help app (`POST` to the admin API), then tries to create an account with that domain name (`GET` to the Institutions API).  The expected result would be for the LEI to be returned from the Institutions API, but instead it returns a 404 since the event was not projected to the database.

Some unusual log messages that may or may not be related:
```
[hmda-institutions-akka.actor.default-dispatcher-4] [akka://hmda-institutions/system/cassandra-journal/$b] 
Invalid replayed event [sequenceNr=5, writerUUID=15dfbc25-633e-4a63-a0fa-270eca48d856] from a new writer. 
An older writer already sent an event [sequenceNr=5, writerUUID=a4bf5c09-9f62-4313-bd3b-9d9075827104] whose 
sequence number was equal or greater for the same persistenceId [InstitutionDBProjector]. Perhaps, the new 
writer journaled the event out of sequence, or duplicate persistentId for different entities?
```

nickgrippin:
Closes #2273

### DIFF
--- a/common/src/main/scala/hmda/publication/KafkaUtils.scala
+++ b/common/src/main/scala/hmda/publication/KafkaUtils.scala
@@ -10,12 +10,10 @@ import akka.kafka.{ConsumerSettings, ProducerSettings, Subscriptions}
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.{Keep, Sink, Source}
 import com.typesafe.config.ConfigFactory
+import hmda.model.institution.Institution
 import org.apache.kafka.clients.consumer.ConsumerConfig
 import org.apache.kafka.clients.producer.{KafkaProducer, ProducerRecord}
-import org.apache.kafka.common.serialization.{
-  StringDeserializer,
-  StringSerializer
-}
+import org.apache.kafka.common.serialization.{StringDeserializer, StringSerializer}
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -32,6 +30,20 @@ object KafkaUtils {
         .withBootstrapServers(kafkaHosts)
 
     producerSettings.createKafkaProducer()
+  }
+
+  def produceInstitutionRecord(topic: String, key: String, value: Institution)(
+    implicit system: ActorSystem,
+    materializer: ActorMaterializer): Future[Done] = {
+
+    val producerSettings =
+      ProducerSettings(system, new StringSerializer)
+        .withBootstrapServers(kafkaHosts)
+
+    Source
+      .single(new ProducerRecord(topic, key, value))
+      .toMat(Producer.plainSink(producerSettings))(Keep.right)
+      .run()
   }
 
   def produceRecord(topic: String, key: String, value: String)(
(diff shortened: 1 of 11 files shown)


common/src/main/scala/hmda/publication/KafkaUtils.scala
```diff
@@ -74,4 +97,33 @@ object KafkaUtils {
...
+      ec: ExecutionContext) = {
+
+    val config = system.settings.config.getConfig("akka.kafka.consumer")
+
+    val consumerSettings: ConsumerSettings[String, InstitutionKafkaEvent] =
+      ConsumerSettings(config,
+                       new StringDeserializer,
+                       new InstitutionKafkaEventsDeserializer)
+        .withBootstrapServers(kafkaHosts)
+        .withGroupId(UUID.randomUUID().toString)
```
jmarin: You need to have a stable `groupId` here. Otherwise when the process restarts the new consumer will not respect the offset

common/src/main/scala/hmda/publication/KafkaUtils.scala
```diff
@@ -74,4 +97,33 @@ object KafkaUtils {
...
+        .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
+
+    Consumer
+      .committableSource(consumerSettings, Subscriptions.topics(topic))
```
jmarin: Probably worth adding a second, "admin" topic so that this event can be triggered on demand without sending additional, repeated events to the journal

institutions-api/src/main/scala/hmda/institution/HmdaInstitutionApi.scala
```diff
@@ -24,17 +46,58 @@ object HmdaInstitutionApi extends App {
...
-  implicit val system = ActorSystem("hmda-institutions")
-
   system.actorOf(HmdaInstitutionQueryApi.props(), "hmda-institutions-api")
+
+  val consumerSettings: ConsumerSettings[String, InstitutionKafkaEvent] =
+    ConsumerSettings(kafkaConfig,
+                     new StringDeserializer,
+                     new InstitutionKafkaEventsDeserializer)
+      .withBootstrapServers(kafkaHosts)
+      .withGroupId(UUID.randomUUID().toString)
```
jmarin: Same comment as above regarding `groupId`

jmarin:
There are a couple of things that need resolving. Also, Travis is not happy.
This also seems to lack the `kubernetes` configuration work so that this can be deployed

common/src/main/scala/hmda/publication/KafkaUtils.scala
nickgrippin: I was messing around with this, but since each message holds a serialized Institution, I don't think you can send a custom message using the command line (unlike strings).  Is there a way to pass in some `Institution(Bank0,test,test)` using the console producer?

PR #2375, opened 2018-12-18
Summary serializer

### LINKED ISSUE #2369: Summary endpoint 504s
With the merge of https://github.com/cfpb/hmda-platform/pull/2355, the summary endpoint should work, but instead results in a 504 Gateway Timeout.

omnipresent07:
Closes #2369


common/src/main/scala/hmda/serialization/filing/FilingCommandsSerializer.scala
```diff
@@ -49,6 +50,8 @@ class FilingCommandsSerializer(system: ExtendedActorSystem)
       updateSubmissionToProtobuf(cmd, resolver).toByteArray
     case cmd: GetLatestSubmission =>
       getLatestSubmissionToProtobuf(cmd, resolver).toByteArray
+    case cmd: GetSubmissionSummary =>
+      getSubmissionSummaryToProtobuf(cmd, resolver).toByteArray
```
nickgrippin: I think you need the other conversion as well in this file

omnipresent07: Great catch! Updated.

PR #2376, opened 2018-12-18
HMDA Filer Flag

### LINKED ISSUE #2344: Set hmda filer to true upon signing
Update the `hmdaFiler` to true when a filer signs their submission. When this is done, the `InstitutionModified` event that ensues should be published to `Kafka` so that the read only `institution-api` microservice picks up the change.

nickgrippin:
Closes #2344


omnipresent07:
Probably worth increasing duration in HmdaValidationErrorSpec . I've restarted travis about 3 times but it won't bulge.

PR #2403, opened 2018-12-23
Change grpc to http

omnipresent07:
This PR addresses the memory leak issue with GRPCs


kubernetes/hmda-platform/values.yaml
```diff
@@ -81,6 +81,14 @@ grpc:
     port: 60081
     host: census-api-grpc.default.svc.cluster.local
 
+http:
+  check_digit:
+    port: 9091
+    host: check-digit-http.default.svc.cluster.local
```
kgudel: this should be: `check-digit.default.svc.cluster.local`

census-api/src/main/scala/hmda/census/validation/CensusValidation.scala
```diff
@@ -9,6 +9,10 @@ object CensusValidation {
     indexedTract.contains(tract)
   }
 
+  def isTractValidTest(tract: String): Boolean = {
```
PatrickHSI: @omnipresent07 does this need to be implemented? Looks like it's just a placeholder to default to true

check-digit/src/main/scala/hmda/uli/HmdaUli.scala
```diff
@@ -26,5 +26,5 @@ object HmdaUli extends App {
   implicit val system = ActorSystem("hmda-uli")
 
   system.actorOf(HmdaUliApi.props(), "hmda-uli-api")
-  system.actorOf(HmdaUliGrpc.props(), name = "hmda-uli-grpc")
+//  system.actorOf(HmdaUliGrpc.props(), name = "hmda-uli-grpc")
```
PatrickHSI: If you're removing GRPC you can go ahead remove this commented out code. Doesn't look like you're leaving any other references to GRPC in place for roll back.

omnipresent07: New commits fix the review comments.

PR #2461, opened 2019-01-03
make year value variable

zencircle:


omnipresent07:
hmda-regulator doesn't use this value

PR #2501, opened 2019-01-22
Analytics test

omnipresent07:
Reduce parallelism to 1.

### DIFF
--- a/hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
+++ b/hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
@@ -132,7 +132,7 @@ object HmdaAnalyticsApp
         }
         .runWith(Sink.ignore)
 
-    def deleteLeiRows: Future[Done] =
+    def deleteLarRows: Future[Done] =
       readRawData(submissionId)
         .map(l => l.data)
         .drop(1)
@@ -146,7 +146,7 @@ object HmdaAnalyticsApp
         }
         .runWith(Sink.ignore)
 
-    def insertLeiRows: Future[Done] =
+    def insertLarRows: Future[Done] =
       readRawData(submissionId)
         .map(l => l.data)
         .drop(1)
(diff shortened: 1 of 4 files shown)


hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -110,9 +111,24 @@ object HmdaAnalyticsApp
...
 
             delete <- transmittalSheetRepository.deleteByLei(ts.lei)
-            insert <- transmittalSheetRepository.insert(ts)
 
-          } yield insert
+          } yield delete
+        }
+        .runWith(Sink.ignore)
+    def step1: Future[Done] =
+      readRawData(submissionId)
```
kgudel: Do we never store the parsed ts data in cassandra? We certainly want to minimize what we write to cassandra but it feels strange that we are duplicating the parsing step when we write to postgres.

omnipresent07: parsed ts data is stored in cassandra. these steps aren't parsing the data but rather inserting data into PG. The raw data from cassandra needs to be parsed into TransmittalSheet prior to inserting into PG

kgudel: This may just be a misunderstanding of how the data is stored on my part but we do call `TsCsvParser` on line 122. If we are also storing the parsed TS couldn't we pull that from cassandra, instead of the data in csv form, and skip that step?

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -161,13 +161,13 @@ object HmdaAnalyticsApp
...
     def result =
       for {
-        _ <- step0
+        _ <- deleteTsRow
         _ = log.info(s"Deleting data from TS for  $submissionId")
-        _ <- step1
+        _ <- insertTsRow
         _ = log.info(s"Adding data into TS for  $submissionId")
-        _ <- step2
+        _ <- deleteLeiRows
```
kgudel: Should this be `deleteLarRows` instead of `Lei`? Same for line 170.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
kgudel: #2505

kgudel:
LGTM

PR #2507, opened 2019-01-24
Jenkins Improvements for HMDA-Platform

kgudel:
1. Add Short Commit as Env Variable to force redeployment
1. Clean up code
1. Remove helm package step

In order to check the currently deployed version you can use something like:
```
kubectl describe pods hmda-platform | grep image
```

### DIFF
--- a/hmda/Jenkinsfile
+++ b/hmda/Jenkinsfile
@@ -7,21 +7,17 @@ volumes: [
   hostPathVolume(mountPath: '/var/run/docker.sock', hostPath: '/var/run/docker.sock'),
 ]) {
    node('buildSbtPod') {
-     def repo = checkout scm
-     def gitCommit = repo.GIT_COMMIT
-     def gitBranch = repo.GIT_BRANCH
-     def gitDescribe = sh(returnStdout: true, script: 'git describe --always --tags')
-     def gitTag = sh(returnStdout: true, script: "git tag --sort version:refname | tail -1").trim()
+     sh(env | sort)
+
      def commitId = sh(returnStdout: true, script: 'git rev-parse HEAD')
      def shortCommit = commitId[0..7]
-     def result = sh(returnStatus: true, script: "git log -1 | grep '\\[deploy pr\\]'")
-     def isDeployPR = result == 0 ? true : false
+     def result = sh(returnStdout: true, script: "git log -1").contains("[deploy pr]")
+     
      if (gitBranch == "master") {
        env.DOCKER_TAG = "latest"
      }
      else if (isDeployPR) {
         env.DOCKER_TAG = "${env.JOB_NAME?.split('/')[1]}-${env.BUILD_NUMBER}"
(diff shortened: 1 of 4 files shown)


hmda/Jenkinsfile
```diff
@@ -6,34 +6,31 @@ podTemplate(label: 'buildSbtPod', containers: [
...
+     
+    if (gitBranch == "master") {
+      env.DOCKER_TAG = "latest"
+    } else if (taggedBranch) {
+      env.DOCKER_TAG = env.TAG_NAME
+    } else {
+      env.DOCKER_TAG = env.BRANCH_NAME
+    }
+
+    println "DOCKER_TAG: ${env.DOCKER_TAG}, gitTag: ${gitTag}, commitId: ${commitId}, isDeployPR: ${isDeployPR}"
```
awolfe76: Should we update this to print other information?

hmda/Jenkinsfile
```diff
@@ -6,34 +6,31 @@ podTemplate(label: 'buildSbtPod', containers: [
...
+  node('buildSbtPod') {
+    sh "env | sort"
 
-     println "DOCKER_TAG: ${env.DOCKER_TAG}, gitDescribe: ${gitDescribe}, gitTag: ${gitTag}, commitId: ${commitId}, isDeployPR: ${isDeployPR}"
+    def repo = checkout scm
+    def gitBranch = repo.GIT_BRANCH
+    def gitCommit = repo.GIT_COMMIT
+    def shortCommit = repo.GIT_COMMIT[0..7]
+    def isDeployPR = sh(returnStdout: true, script: "git log -1").contains("[deploy pr]")
+    def taggedBranch = env.TAG_NAME != null
```
awolfe76: Just to match other patterns (`GIT_` and `DOCKER_`), would it be better to use

`gitTag` vs `taggedBranch`

and 

`env.GIT_TAG` vs `env.TAG_NAME`?

kgudel: `env.TAG_NAME` is jenkins generated so we don't have much choice about that one and since `taggedBranch` is a bool instead of actually holding the tag I don't think it should be exactly like the others. How about `gitTagged`?

hmda/Jenkinsfile
kgudel: Anything you think would be useful? Maybe `gitbranch`?

hmda/Jenkinsfile
awolfe76: 👍

hmda/Jenkinsfile
awolfe76: Yeah, I think so. Sorry, I really just meant because I don't think `gitTag` exists anymore.

kgudel: Oh, so it doesn't. I'll update that.

awolfe76:
LGTM, thanks for the updates.

PR #2512, opened 2019-01-25
Finalized strategy for institution  filters across  micro services

PatrickHSI:
Hmda4 Testing :

 - [x] HMDA-Regulator 

- [x] HMDA-Analytics

- [x] Modified-LAR


hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -58,14 +58,14 @@ object HmdaAnalyticsApp
 
   val kafkaConfig = system.settings.config.getConfig("akka.kafka.consumer")
   val config = ConfigFactory.load()
-
+  val bankFilter = config.getConfig("filter")
+  val bankFilterList = bankFilter.getString("bank-filter-list").toUpperCase.split(",")
```
omnipresent07: Can we move these two lines in `processData` method? That way the `bankFilter` and `bankFilterList` will be read each time without requiring a restart of the pod when we make a change to the environment variable. For this to happen you'll need `config=ConfigFactory.load` inside `processData` as well

kgudel:
I believe that if we are filtering the institutions at the hmda-analytics level we shouldn't need to do so again in the regulator app, since regulator pulls the data that hmda-analytics writes. I think in future we would want to pull the modified lar data from that postgres table as well. Along with more conforming to the CQRS model that would let us filter the institutions at only 1 point.

PatrickHSI:
@kgudel  That is true , however since these tables have been populated before any of this filter logic we'll need to repopulate the table with the filter logic or else the'll start showing up in the regulator data due to past entries.

kgudel:
In that case we should note in the deployment documentation to clean up postgres when we deploy these changes.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
PatrickHSI: we attempted this in hmda-regulator and the config still was not reloaded. Resulting in us having to scale down and restart the application regardless.

PR #2527, opened 2019-02-01
Metrics

omnipresent07:
This PR brings in lightbend telemetry commercial plugin. The metrics will be exposed on port 9009. Currently two separate builds happen - one for docker hub and another for DTR. Ideally, the second build should skip tests in order to make the builds faster. There will be another PR to do that part.


omnipresent07:
> In general I think that this all looks good. My only question would be whether the cinnamon related configuration elements in the `application.conf` would introduce problems in the public build?

It will not. An improvement that needs to happen to this is since we're now building the docker image twice in hmda-platform/Jenkins (one for public, one for internal), it will take significantly more time. Improvement should be to skip running tests the second time we build the docker image.

kgudel:
#2564

PR #2548, opened 2019-02-13
remove spaces from edits csv header

### LINKED ISSUE #1566: Inconsistent CSV header formatting in API responses
The private endpoints use header rows with spaces after the commas, while the public endpoints use header rows without spaces.  I only noticed due to a 3rd party CSV parser not being able to handle the difference, so it might not be a huge deal to others, but could be nice to have some consistency.

For example:

[https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md#edits-csv](https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md#edits-csv) 

This returns a response with a header row of `editType, editId, loanId`

[https://github.com/cfpb/hmda-platform/blob/master/Documents/public-api.md#check-digit](https://github.com/cfpb/hmda-platform/blob/master/Documents/public-api.md#check-digit)

The batch version of this returns a response with a header row of `loanId,checkDigit,uli`

awolfe76:
Closes #1566 

Although I think the best practice is to ignore them if they exist, this leaves [the check digit](https://github.com/cfpb/hmda-platform/blob/b3d6d4766d4c8dd906460ccf0cfed3d3399942f8/check-digit/src/main/scala/hmda/uli/api/http/ULIHttpApi.scala#L138) as is because I think it's more common to not have spaces in the headers.

**🙏 Please double check this! 🙏**

### DIFF
--- a/hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala
+++ b/hmda/src/main/scala/hmda/api/http/filing/submissions/EditsHttpApi.scala
@@ -184,7 +184,7 @@ trait EditsHttpApi extends HmdaTimeDirectives {
   }
 
   private val csvHeaderSource =
-    Source.fromIterator(() => Iterator("editType, editId, ULI\n"))
+    Source.fromIterator(() => Iterator("editType,editId,ULI\n"))
 
   private def validationErrorEventStream(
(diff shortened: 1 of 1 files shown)


kgudel:
LGTM

PR #2551, opened 2019-02-14
Write submissionId to sql database

### LINKED ISSUE #2546: Add submission-id to transmittal sheet postgres table
This will be a change in `hmda-analytics` and the field will take the form of `<lei>-2018-<id>`.

kgudel:
Closes #2546

We will need to regen our data our sql data to get these changes.

We will also have to add the new column using the following command:

```
ALTER TABLE transmittalsheet2018
ADD COLUMN submission_id varchar;
```


hmda-regulator/src/main/scala/hmda/regulator/query/RegulatorComponent.scala
```diff
@@ -152,6 +152,7 @@ trait RegulatorComponent {
     def agency = column[Int]("agency")
     def totalLines = column[Int]("total_lines")
     def taxId = column[String]("tax_id")
+    def submissionId = column[String]("tax_id")
```
omnipresent07: This doesn't seem right. If we're adding the submissionId, we should pass the submission id.

kgudel: Oops, copy paste error, on it.

omnipresent07:
LGTM 👍

PR #2552, opened 2019-02-14
Disclosure prep

omnipresent07:
1 - New endpoints in census to fetch all census records
2 - Common downloader from census to fetch the records
3 - Had dependency on modified-lar to fetch all census records at time of startsup
4 - New modified-lar table (including census data)
5 - Change modified-lar to persist to S3 and PG
6 - Retry 10 times w/ 30 sec delay incase of failure


modified-lar/src/main/scala/hmda/publication/lar/publication/ModifiedLarPublisher.scala
```diff
@@ -78,20 +98,74 @@ object ModifiedLarPublisher {
...
+              // We listen on the completion of both materialized values but we only keep the S3 as the result
+              // since that is a meaningful value
+              case (futPostgresRes, futS3Res) =>
+                for {
+                  _ <- futPostgresRes
+                  s3Res <- futS3Res
+                } yield s3Res
+            }
+
+          val finalResult: Future[Unit] = for {
```
kgudel: It seems from the logic that removing old psql records and populating the new ones might happen simultaneously and could lead to a race condition. It seems like we should remove LEI and repopulate with new records synchronously.

modified-lar/src/main/scala/hmda/publication/lar/ModifiedLarApp.scala
```diff
@@ -30,63 +37,115 @@ object ModifiedLarApp extends App {
...
+          .toMat(Sink.ignore)(Keep.both)
+          .run()
+
+      streamCompleted.onComplete {
+        case Failure(exception) =>
+          log.error("Kafka consumer infinite stream failed", exception)
+          kafkaControl
+            .shutdown()
+            .flatMap(_ => system.terminate())
+            .foreach(_ => sys.exit(1)) // The Cassandra connection stays up and prevents the app from shutting down even when the Kafka stream is not active, so we explicitly shut the application down
```
kgudel: 👍

modified-lar/src/main/scala/hmda/publication/lar/publication/ModifiedLarPublisher.scala
omnipresent07: `removeLei ` and `postgresOut ` are declared as `def` as opposed to `val` so they will happen synchronously.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.2.0
+sbt.version=1.2.8
```
PatrickHSI: Was this version update intentional?

modified-lar/src/main/scala/hmda/publication/lar/ModifiedLarApp.scala
PatrickHSI: 👍

PatrickHSI:
Has this been tested in hmda4 with a million row file?  @omnipresent07

project/build.properties
omnipresent07: yes. had to upgrade because of metrics PR might as well update to latest.

omnipresent07:
Yup, it has been.

PatrickHSI:
@omnipresent07  I think you need to tweak your code around the Option[String] loging we have to use for the new nullable column in transmittal sheet 


```
[error] /home/travis/build/cfpb/hmda-platform/hmda/src/main/scala/hmda/query/ts/TsComponent.scala:53:36: type mismatch;
[error]  found   : ((String, Int, String, Int, Int, String, String, String, String, String, String, String, Int, Int, String, Option[String])) => hmda.query.ts.TransmittalSheetEntity
[error]  required: ((String, Int, String, Int, Int, String, String, String, String, String, String, String, Int, Int, String, String)) => ?
[error]       ) <> (TransmittalSheetEntity.tupled, TransmittalSheetEntity.unapply)
[error]                                    ^
[error] /home/travis/build/cfpb/hmda-platform/hmda/src/main/scala/hmda/query/ts/TsComponent.scala:53:67: type mismatch;
[error]  found   : Option[(String, Int, String, Int, Int, String, String, String, String, String, String, String, Int, Int, String, Option[String])]
[error]  required: Option[(String, Int, String, Int, Int, String, String, String, String, String, String, String, Int, Int, String, String)]
[error]       ) <> (TransmittalSheetEntity.tupled, TransmittalSheetEntity.unapply)
[error] ```

PR #2553, opened 2019-02-14
WIP Removed  and organized Unused Imports

PatrickHSI:


census-api/src/main/scala/hmda/census/api/grpc/CensusServiceImpl.scala
```diff
@@ -1,8 +1,8 @@
 package hmda.census.api.grpc
 
 import akka.stream.Materializer
-import hmda.grpc.services._
 import hmda.census.validation.CensusValidation._
+import hmda.grpc.services._
```
omnipresent07: Why are we adding a grpc import here?

PR #2559, opened 2019-02-19
Fix Trimming

omnipresent07:
Trim the lar data prior to validating it. There are cases when lar data ends in `|` followed by windows end of line character. This makes the number of records in a lar row as 111 as opposed to 110 and the validation fails. This trimming matches the trimming done elsewhere in HmdaValidationError.


awolfe76:
Closes [GHE]/HMDA-Operations/hmda-devops/issues/953.

PR #2562, opened 2019-02-19
Filers Route

kgudel:
Depends on #2551


hmda/src/main/scala/hmda/query/ts/TsComponent.scala
```diff
@@ -0,0 +1,96 @@
...
+    def phone = column[String]("phone")
+    def email = column[String]("email")
+    def street = column[String]("street")
+    def city = column[String]("city")
+    def state = column[String]("state")
+    def zipCode = column[String]("zip_code")
+    def agency = column[Int]("agency")
+    def totalLines = column[Int]("total_lines")
+    def taxId = column[String]("tax_id")
+    def submissionId = column[String]("tax_id")
```
omnipresent07: copy/paste issue on submissionId. I believe this was fixed in the other PR.

hmda/src/main/scala/hmda/query/ts/TsComponent.scala
```diff
@@ -0,0 +1,96 @@
+package hmda.query.ts
```
omnipresent07: This entire class is quite similar to `TransmittalSheetComponent` in hmda-analytics as well as `TransmittalSheetTable` in hmda-regulator. Maybe not as part of this PR but we should extract all these into common rather than duplicating the code base. Maybe a separate ticket.

kgudel:
https://github.com/cfpb/hmda-platform/issues/2579

hmda/src/main/scala/hmda/query/ts/TsComponent.scala
kgudel: https://github.com/cfpb/hmda-platform/issues/2579

omnipresent07:
LGTM 👍

Issue #2557, opened 2019-02-19
HMDA-Regulator No Longer Generating LAR Data


PatrickHSI:
There has been a persistent issue since February 13th 2019 in which HMDA-Regulator fails to generate  LAR regulator data during the nightly builds. 

@zencircle  has identified the process that generates the HMDA-Regulator data is killed soon after it starts due to resource issues. 

Possible Scala bug as to why we're not seeing any errors :

https://github.com/scala/bug/issues/9554

PatrickHSI:
It appears the production deployment consumes the following error. Hinting  that the query loading all of the table data into memory is causing the memory issue. 

Looking into an alternative method to stream the data from postgres.

```
hmda-regulator [ERROR] [02/19/2019 02:41:44.151] [larTask-scheduler-1] [akka.actor.ActorSystemImpl(larTask)] exception on LARS’ timer thread
hmda-regulator java.lang.OutOfMemoryError: GC overhead limit exceeded
hmda-regulator 	at akka.dispatch.AbstractNodeQueue.<init>(AbstractNodeQueue.java:32)
hmda-regulator 	at akka.actor.LightArrayRevolverScheduler$TaskQueue.<init>(LightArrayRevolverScheduler.scala:304)
hmda-regulator 	at akka.actor.LightArrayRevolverScheduler$$anon$4.nextTick(LightArrayRevolverScheduler.scala:269)
hmda-regulator 	at akka.actor.LightArrayRevolverScheduler$$anon$4.run(LightArrayRevolverScheduler.scala:235)
hmda-regulator 	at java.lang.Thread.run(Thread.java:748)

PR #2563, opened 2019-02-20
OOM LAR Data Thread

PatrickHSI:
Resolves issue [2557](https://github.com/cfpb/hmda-platform/issues/2557) by streaming data from postgres LAR table in lieu of loading the millions of records into memory.


hmda-regulator/src/main/resources/application.conf
```diff
@@ -53,6 +53,5 @@ aws {
 }
 
 filter {
```
omnipresent07: Can you put the filter back?

PatrickHSI: done.

hmda-regulator/src/main/scala/hmda/regulator/scheduler/LarScheduler.scala
```diff
@@ -80,24 +82,29 @@ class LarScheduler extends HmdaActor with RegulatorComponent {
...
+            log.info(
+              "Pushing to S3: " + s"$bucket/$environment/lar/$fileName" + ".")
+          }
+          case Failure(t) =>
+            println("An error has occurred getting LAR Data in Future: " + t.getMessage)
+
         }
-        case Failure(t) =>
-          println("An error has occurred getting LAR Data: " + t.getMessage)
+      } catch {
```
kgudel: This is minor but generally the more scala-y approach to error handling is to use `try` and then `case Success` and `case Failure` instead of `catch`. Here is a [short article](https://www.codacy.com/blog/why-use-try-instead-of-try-catch-in-scala/) on it. I think it might be a good change just to keep our codebase consistent.

hmda-regulator/src/main/scala/hmda/regulator/query/RegulatorComponent.scala
```diff
@@ -490,14 +490,21 @@ trait RegulatorComponent {
...
-      db.run(table.filterNot(_.lei.toUpperCase inSet bankIgnoreList).result)
+        bankIgnoreList: Array[String]): DatabasePublisher[LarEntityImpl] = {
+      db.stream(
+        table
+          .filterNot(_.lei.toUpperCase inSet bankIgnoreList)
+          .result
+          .withStatementParameters(
+            rsType = ResultSetType.ForwardOnly,
+            rsConcurrency = ResultSetConcurrency.ReadOnly,
+            fetchSize = 1000
```
kgudel: `fetchSize` might work well as part of the configuration incase we decide to tune it down the road.

hmda-regulator/src/main/resources/application.conf
```diff
@@ -3,16 +3,16 @@ akka {
...
+  //  LarScheduler = """akka.quartz.schedules.LARScheduler.expression = "0 0/2 * * * ?""""
+  //  TsScheduler = """akka.quartz.schedules.TsScheduler.expression = "0 0/2 * * * ?""""
   //Prod Schedules
   //fire at 11:00PM everyday
-  PanelScheduler = """akka.quartz.schedules.PanelScheduler.expression = "0 20 06 * * ?""""
- // fire at 11:10PM everyday
-  LarScheduler = """akka.quartz.schedules.LARScheduler.expression ="0 55 05 * * ?""""
- // fire at 11:20PM everyday
-  TsScheduler = """akka.quartz.schedules.TsScheduler.expression = "0 10 06 * * ?""""
+  PanelScheduler = "akka.quartz.schedules.PanelScheduler.expression = "${PANEL_SCHEDULE}""
```
kgudel: It might be worthwhile keeping the original schedules as a default.

kubernetes/hmda-regulator/templates/deployment.yaml
```diff
@@ -106,6 +106,21 @@ spec:
               secretKeyRef:
                 name: inst-postgres-credentials
                 key: url
+          - name: TS_SCHEDULE
```
kgudel: I am so happy to have this as part of the configuration! 👍

hmda-regulator/src/main/scala/hmda/regulator/scheduler/LarScheduler.scala
PatrickHSI: This try and catch block is a relic from testing/implementing the streaming logic. It's actually redundant now that everything is in place and we're tracking the Futures completion via case Success/Failure

hmda-regulator/src/main/resources/application.conf
PatrickHSI: Ill add a note outlining what these values could reflect. Really need to add this to Regulator documentation

hmda-regulator/src/main/scala/hmda/regulator/scheduler/LarScheduler.scala
```diff
@@ -82,29 +82,24 @@ class LarScheduler extends HmdaActor with RegulatorComponent {
...
+      results onComplete {
+        case Success(result) => {
+          log.info(
+            "Pushing to S3: " + s"$bucket/$environment/lar/$fileName" + ".")
         }
-      } catch {
-        case e: Exception =>
-          println("An error has occurred getting LAR Data: " + e.getMessage)
+        case Failure(t) =>
+          println(
```
kgudel: Should this be a `log` instead of `print`?

PR #2572, opened 2019-02-21
Jenkins fix

### LINKED ISSUE #2571: Fix Jenkinsfile for building tags
hmda/Jenkinsfile needs to be fixed for building tags. The username for DTR doesn't match the repository name which causes errors like `No such image: ****/****-platform:latest`

omnipresent07:
Closes #2571 

1. Create separate docker images for public and internal
2. Delete created docker file from public build
3. Respect docker user for DTR
4. Deploy from DTR
5. Delete images in the end

### DIFF
--- a/hmda/Jenkinsfile
+++ b/hmda/Jenkinsfile
@@ -107,7 +107,7 @@ volumes: [
 
     stage('Deploy') {
         withCredentials([string(credentialsId: 'internal-docker-registry', variable: 'DOCKER_REGISTRY_URL')]){
-          if (true) {
+          if (env.BRANCH_NAME == 'master' || isDeployPR) {
             container('helm') {
               sh """
                 helm upgrade --install --force \
--- a/hmda/Jenkinsfile
+++ b/hmda/Jenkinsfile
@@ -84,24 +84,21 @@ volumes: [
       }
 
       container('docker') {
-        withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'dockerhub',
-          usernameVariable: 'DOCKER_HUB_USER', passwordVariable: 'DOCKER_HUB_PASSWORD']]) {
           withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'hmda-platform-jenkins-service',
             usernameVariable: 'DTR_USER', passwordVariable: 'DTR_PASSWORD']]) {
             withCredentials([string(credentialsId: 'internal-docker-registry', variable: 'DOCKER_REGISTRY_URL')]){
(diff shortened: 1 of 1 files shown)


hmda/Jenkinsfile
```diff
@@ -38,13 +38,15 @@ volumes: [
         withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'dockerhub',
                   usernameVariable: 'DOCKER_HUB_USER', passwordVariable: 'DOCKER_HUB_PASSWORD']]) {
             if (gitTagged || gitBranch == "master") {
```
kgudel: Let's include `isDelpoyPr` in this.

hmda/Jenkinsfile
```diff
@@ -82,30 +84,24 @@ volumes: [
...
             usernameVariable: 'DTR_USER', passwordVariable: 'DTR_PASSWORD']]) {
             withCredentials([string(credentialsId: 'internal-docker-registry', variable: 'DOCKER_REGISTRY_URL')]){
-              sh "docker build --rm -t=${env.DTR_USER}/hmda-platform hmda/target/docker/stage"
               if (gitTagged || gitBranch == "master" || isDeployPR) {
+                sh "docker build --rm -t=${env.DOCKER_HUB_USER}/hmda-platform:internal hmda/target/docker/stage"
 
                 //Push to DTR
                 sh """
-                  docker tag ${env.DTR_USER}/hmda-platform:${env.DOCKER_TAG} ${DOCKER_REGISTRY_URL}/${env.DTR_USER}/hmda-platform:${env.DOCKER_TAG}
+                  docker tag ${env.DOCKER_HUB_USER}/hmda-platform:internal ${DOCKER_REGISTRY_URL}/${env.DOCKER_HUB_USER}/hmda-platform:${env.DOCKER_TAG}
```
kgudel: I think at this point it might just be easier to put `/hmda` here instead of `env.DOCKER_HUB_USER`. Since we already just have `/hmda` in the `helm` upgrade line. It will also make it more clear in our code where we are pushing out images to.

hmda/Jenkinsfile
omnipresent07: This block doesn't deploy so it would be redundant. We never deploy from docker hub block anymore.

hmda/Jenkinsfile
omnipresent07: done

hmda/Jenkinsfile
kgudel: Got you, looks good then.

PR #2584, opened 2019-02-25
Refactor Configmaps

### LINKED ISSUE #2542: config-map issues for bank filter list
since bank-filter-configmap already exists as part of hmda-regulator, deploying other microservices causes errors `Error: UPGRADE FAILED: no ConfigMap with the name "bank-filter-configmap" found`.

To resolve this the config-map needs to be taken out from configmap.yaml. There needs to be a better solution to handle this.

kgudel:
This PR:
1. Breaks configs up based on their use rather than the projects that they are associated with as most configuration is relevant to many services.
1. Removes configs from the project definitions as they no longer fit well within them.

Closes #2542


kubernetes/modified-lar/values.yaml
```diff
@@ -48,6 +48,8 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+<<<<<<< HEAD
```
omnipresent07: It seems like some merge conflicts are added as commits?

kubernetes/config-maps/http-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: http-configmap
+data:
+  check.digit.host: check-digit.default.svc.cluster.local
```
omnipresent07: What is using `check.digit.host` and `census.host` at the moment? The platform no longer has dependencies into census and check-digit. It'll be more clear if we can remove these from this config-map so there are no false assumptions.

omnipresent07:
LGTM 👍

PR #2585, opened 2019-02-25
dont show TS row for S305

### LINKED ISSUE #2583: S305 response
When users get an S305 error, the platform is showing the Transmittal Sheet in the response field rather than the record that is actually duplicated. This leads users to (falsely) believe that they have a duplicate TS record. The correct behavior would be identifying the duplicate LAR record and displaying that in the "All Data Fields in the LAR."  In the short term, it would be much better just to show a message saying "Your LAR contains one or more duplicate records." than it would be to show the TS. 

I think this was one that originally showed an "Error - mismatch" in that field, and the fix shows it now returning the TS. 

image

omnipresent07:
Closes #2583

### DIFF
--- a/common/src/main/scala/hmda/model/filing/ts/TsFieldMapping.scala
+++ b/common/src/main/scala/hmda/model/filing/ts/TsFieldMapping.scala
@@ -17,6 +17,6 @@ object TsFieldMapping {
     "Total Number of Entries Contained in Submission" -> ts.totalLines.toString,
     "Federal Taxpayer Identification Number" -> ts.taxId,
     "Legal Entity Identifier (LEI)" -> ts.LEI,
-    "All data fields in the LAR" -> ts.toCSV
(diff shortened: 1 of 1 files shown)


common/src/main/scala/hmda/model/filing/ts/TsFieldMapping.scala
```diff
@@ -17,6 +17,6 @@ object TsFieldMapping {
     "Total Number of Entries Contained in Submission" -> ts.totalLines.toString,
     "Federal Taxpayer Identification Number" -> ts.taxId,
     "Legal Entity Identifier (LEI)" -> ts.LEI,
-    "All data fields in the LAR" -> ts.toCSV
+    "All data fields in the LAR" -> "Error - mismatch"
```
kgudel: I believe that we want this one to say: `Your LAR contains one or more duplicate records.`

PR #2590, opened 2019-02-26
Ignore case while deleting LEIs

### LINKED ISSUE #2592: Ignore case in hmda-analytics delete
When hmda-analytics deletes lei's it isn't ignoring case.

omnipresent07:
Closes #2592 

ignore case while deleteing LEI's from PG table

### DIFF
--- a/hmda-analytics/src/main/scala/hmda/analytics/query/LarComponent.scala
+++ b/hmda-analytics/src/main/scala/hmda/analytics/query/LarComponent.scala
@@ -20,7 +20,7 @@ trait LarComponent {
         sqlu"""INSERT INTO #${tableName}
         VALUES (
           ${le.id},
-          ${le.lei},
+          ${le.lei.toUpperCase},
           ${le.uli},
           ${le.appDate},
           ${le.loanType},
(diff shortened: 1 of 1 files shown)


hmda-analytics/src/main/scala/hmda/analytics/query/LarComponent.scala
```diff
@@ -12,15 +12,15 @@ trait LarComponent {
...
+        sqlu"DELETE FROM #${tableName} WHERE UPPER(lei) = ${lei.toUpperCase}"
       }
 
     def insert(le: LarEntity): Future[Int] =
       config.db.run {
         sqlu"""INSERT INTO #${tableName}
         VALUES (
           ${le.id},
-          ${le.lei},
+          ${le.lei.toUpperCase},
```
kgudel: I don't think that we should modify the institution's LEI in our db. The other line change should be sufficient to make sure we purge our db of old records regardless of LEI casing.

omnipresent07: Not modifying the LEI to upper case could cause mismatches down the stream while doing data analysis. The data analysts will have to write `UPPER` in each query.

PR #2598, opened 2019-02-26
Prevent keycloak and hmda-platform from using the same nodes 

kgudel:


kubernetes/keycloak/Jenkinsfile
```diff
@@ -28,7 +28,7 @@ podTemplate(label: 'buildHelm', containers: [
                                 --set keycloak.username=${env.KEYCLOAK_USER} \
                                 --set keycloak.password=${env.KEYCLOAK_PASSWORD} \
                                 --set image.pullPolicy=Always \
-                                keycloak kubernetes/keycloak
+                                keycloak stable/keycloak
```
omnipresent07: Do we need to pull from `stable` instead of our own chart? Seems weird so far we've been pulling from `kubernetes/` and now changing to `stable/`

kgudel: This line actually would not work before. We do not have a keycloak chart of our own, only a values file. We use the stable keycloak chart.

omnipresent07: 👍

PR #2606, opened 2019-02-28
remove 'test' from the create an account form heading

### LINKED ISSUE #2605: Update create account heading
The heading on the create an account page still has beta content.

It reads "Create a test account". We need to remove `test`. The heading should say "Create an account".

awolfe76:
Closes #2605 

Draft because we can wait until after the season to apply this **but if we NEED to do a Keycloak deployment at some point** we can add this in.


PatrickHSI:
@awolfe76  is this still a WIP?

PR #2612, opened 2019-03-05
Reporting microservice

omnipresent07:
1. Add filing year to modified lar table
2. Create reporting microservice


common/src/main/scala/hmda/query/repository/ModifiedLarRepository.scala
```diff
@@ -18,8 +18,21 @@ class ModifiedLarRepository(tableName: String,
...
+    }
+
+  /**
+    * Deletes entries in the Modified LAR table by their LEI
+    * @param lei
+    * @return the number of rows removed
+    */
+  def deleteByLei(lei: String, filingYear: Int): Future[Int] =
+    db.run(
+      sqlu"DELETE FROM #${tableName} WHERE lei = $lei and filing_year = $filingYear")
```
PatrickHSI: shouldnt the LEI references  here be wrapped in UPPER? @omnipresent07

omnipresent07: Fixed.

PR #2618, opened 2019-03-07
Added a more robust approach to newline removal in LAR data

PatrickHSI:


Asatap:
@kgudel I resolved that locally. I need to update this branch

PR #2643, opened 2019-03-25
Fix exhaustive match

### LINKED ISSUE #2175: Match not exhaustive on InstitutionState
The compiler complains about a non exhaustive match on `InstitutionState`

omnipresent07:
fix #2175


hmda/src/main/scala/hmda/persistence/institution/InstitutionState.scala
```diff
@@ -19,6 +19,7 @@ case class InstitutionState(institution: Option[Institution],
         } else {
           InstitutionState(this.institution, filing :: filings)
         }
+      case _ => this
```
PatrickHSI: Should we be swallowing all other match cases or logging that they occurred with an event?

PR #2663, opened 2019-04-01
Fixes V626

### LINKED ISSUE #2662: V626 Implementation Fix
There are a couple of significant issues w/ V626:

1) It appears that the edit is not checking that the 5 digit County value is a valid combination of FIPS state + FIPS county codes using the census microservice. I've tested this by entering random 5 digit numbers entered in the County field, and V626 is never flagged. This should be updated ASAP. 

2) It allows values of na and N/AN/ in the County field, when it should allow only NA or a 5 digit FIPS code.

omnipresent07:
Closes #2662


kgudel:
Looks like there are currently some test failures.

omnipresent07:
@kgudel Fixed

PR #2706, opened 2019-04-24
Fix median age

omnipresent07:


hmda-spark-reporting/src/main/scala/com/hmda/reports/processing/IncomeRaceEthnicityProcessing.scala
```diff
@@ -9,10 +9,10 @@ object IncomeRaceEthnicityProcessing {
     "Asian",
     "Native Hawaiian or Other Pacific Islander",
     "Free Form Text Only",
-    "Race Not Available",
+    "Race not available",
```
kgudel: It looks like for the other characteristics, like sex, we capitalize every word so `Sex Not Available`. Is there a reason that race is different?

omnipresent07: Updated

PR #2731, opened 2019-05-07
Data Browser Microservice

### LINKED ISSUE #2721: Create hmda-data-browser microservice
hmda-data-browser microservice will serve as the backend for https://github.com/cfpb/hmda-data-browser

The microservice shall:

- [X] Create API endpoints documents with resources + data expected 
- [x] Fetch data from the postgresql db if and only if it doesn't exist in the redis caching layer
- [x] Accept parameters that allow the following queries:
  - [x] selecting an MSA/Nationwide or State
  - [x] selecting action type - multiple values between `1 - 8`
  - [x] selecting race - multiple values between `Free Form Text Only`, `Race Not Available`, `American Indian or Alaska Native`, `Asian `, `Black or African American`, `Native Hawaiian or Pacific Islander`,  `Two or more minorities`, `White`, `Joint`

omnipresent07:
Closes #2721


data-browser/src/main/scala/hmda/data/browser/models/ModifiedLarEntity.scala
```diff
@@ -0,0 +1,237 @@
+package hmda.data.browser.models
+
+import java.sql.Timestamp
+
+import slick.jdbc.GetResult
+
+case class ModifiedLarEntity(
```
kgudel: Do we need to copy the modified lar entity and repository in the data-browser or can we just pull those from common?

omnipresent07: https://github.com/cfpb/hmda-platform/issues/2739

PR #2793, opened 2019-05-30
Make Kafka Topics Configurable

### LINKED ISSUE #2782: Make kafka topic names configurable
Since we'll have a persistent beta version going forward, the kafka topic names in https://github.com/cfpb/hmda-platform/blob/master/common/src/main/scala/hmda/messages/pubsub/HmdaTopics.scala need to be made configurable so that they can be changed based on the environment that platform is released in

kgudel:
Closes #2782


omnipresent07:
Conflicts need to be resolved.

omnipresent07:
Deployment yaml's should be changed as well so that the env variables in application.conf can be configured based on deployment.

common/src/main/scala/hmda/publication/KafkaUtils.scala
```diff
@@ -111,7 +112,7 @@ object KafkaUtils {
                        new StringDeserializer,
                        new InstitutionKafkaEventsDeserializer)
         .withBootstrapServers(kafkaHosts)
-        .withGroupId("institution")
+        .withGroupId(HmdaTopics.institutionTopic)
```
omnipresent07: I believe this should be `HmdaGroups.analyticsGroup`

PR #2806, opened 2019-06-04
Upgrade Alpine Base Image for Scala Microservices

kgudel:


PatrickHSI:
@kgudel I think updates  to `openjdk:jre-alpine`  also need to be made in `commercial-build.txt`

PR #2836, opened 2019-06-16
Setup hmda-analytics for multiple years

### LINKED ISSUE #2815: Change hmda-analytics to support 2019
hmda-analytics shall be changed to support 2018 and 2019. Based on the year in the kafka event, hmda-analytics will update the relevant PG tables.

omnipresent07:
Closes #2815


omnipresent07:
Yeah, they are different because of the way TS and lar components are structured. Since Lar has 100+ fields, it isn't supported using slick repositories but rather we use sql.

PR #2877, opened 2019-06-27
increase the timeout value

zencircle:
Ambassador needs to wait for check-digit to complete processing the request


omnipresent07:
LGTM 👍

PR #2887, opened 2019-06-28
Adding Public LAR codesheets

### LINKED ISSUE #2861: Creating Public LAR Code Sheet
The issue can be closed with the following resources placed in the docs/v2/spec folder. 
- [x] 2018 LAR PDF Code Sheet

MichaelFurey17:
Close #2861


Kibrael:
@MichaelFurey17 can you adjust the code sheet for 2018 so that it matches the style [here](https://www.ffiec.gov/hmda/pdf/code.pdf).

This will likely take a front and back so the documents should be set up to support that.

PR #2930, opened 2019-07-15
Serve nationwide from S3

### LINKED ISSUE #2890: serve nationwide csv file from S3
Although the backend API can successfully stream the entire nationwide CSV, since this data will not change we should serve an S3 file rather than streaming the data.

### LINKED ISSUE #2933: Normalize the order of the redis cache keys in
the redis cache keys aren't stored in any order at the moment. This makes warming up the cache rather difficult.

omnipresent07:
Closes #2890 
Closes #2933


kubernetes/beta/hmda-data-browser-api/values.yaml
```diff
@@ -69,5 +69,10 @@ service:
 
 databrowser:
   name: databrowser
+  loglevel: INFO
+  nationwideRouteCSV: https://s3.amazonaws.com/cfpb-hmda-public/prod/data-browser/nationwide_all.csv
```
kgudel: For extensibility it would probably be useful to have the `s3` route broken into pieces that we can reuse if we decide to serve more from there. As well, we already have `cfpb-hmda-public` in our `s3-configmap` if you just want to take it from there.

omnipresent07: @kgudel Changes made

PR #2970, opened 2019-07-31
Updates Lar Data Publishing

### LINKED ISSUE #2947: Update hmda-regulator to generate public data
Public Data  will come from modifiedlar2018 and modifiedlar2018_static

- [x] determine which fields would be stripped out [(#1066)](https://github.cfpb.gov/HMDA-Operations/hmda-devops/issues/1066)

- [x] HMDA Panel dynamic (CSV & PSV)
- [x] HMDA TS dynamic (CSV & PSV)
- [x]  HMDA ModifiedLAR dynamic will have 110 fields (CSV & PSV)

PatrickHSI:
closes #2947


hmda-regulator/src/main/scala/hmda/regulator/query/component/RegulatorComponent2018.scala
```diff
@@ -313,7 +315,7 @@ trait RegulatorComponent2018 {
     def propertyValue = column[String]("property_value")
     def homeSecurityPolicy = column[Int]("home_security_policy")
     def landPropertyInterest = column[Int]("lan_property_interest")
-    def totalUnits = column[Int]("total_uits")
+    def totalUnits = column[Int]("total_units")
```
omnipresent07: This is actually typo'd in the `loanapplicationregister2018` table. If we change it here but don't change it in the table, it won't work. I suggest keeping it as `total_uits` for now and ensuring the change is done for 2019 by making sure it is correct in the 2019 table and schema.

omnipresent07:
LGTM 👍

PR #2977, opened 2019-08-01
Add root liveness endpoint

### LINKED ISSUE #2929: Health Check End Point for Data Browser
Data Browser API should have an end point that returns 200 if it is running correctly

kgudel:
Also, refactors hmda-data-browser-api to better match design of other microservices

Closes #2929


omnipresent07:
@kgudel Merge conflicts.

PR #3073, opened 2019-09-06
Remove Census Microservice

### LINKED ISSUE #3021: Remove census-api microservice
This micro-service is currently only being used to stream census data to the `modified-lar` microservice on startup of `modified-lar`. It would be simpler and more reliable to include the census data in the packaging of `modified-lar` which should also allow us to remove the `census-api` microservice entirely.

kgudel:
Closes #3021


omnipresent07:
LGTM 👍

PR #3108, opened 2019-09-24
Beta changes for Institution API and Hmda Analytics

omnipresent07:
- Configure institution API to point to Beta PG table when `isBeta` flag is turned on.
- Configure `hmda-Analytics` to read table names from application.conf. The application.conf reads them as environment variables. 
- Create new beta helm chart for `hmda-analytics` and a config map for PG tables specific to beta. 
- Create new beta helm chart for `institutions-api`.


hmda-analytics/src/main/scala/hmda/analytics/query/SubmissionHistoryComponent.scala
```diff
@@ -11,12 +11,12 @@ trait SubmissionHistoryComponent {
 
   import dbConfig.profile.api._
 
-  class SubmissionHistoryRepository(config: DatabaseConfig[JdbcProfile]) {
+  class SubmissionHistoryRepository(config: DatabaseConfig[JdbcProfile], tableName: String) {
     def fetchYearTable(year: Int): String = {
```
kgudel: This function appears to pick the same table no matter what year is provided.

institutions-api/src/main/resources/application.conf
```diff
@@ -8,6 +8,8 @@ hmda {
   institution {
     createSchema = false
     createSchema = ${?INSTITUTION_PG_CREATE_SCHEMA}
+    isBeta = false
```
kgudel: From a code perspective it seems like it would it would be easier to remove the `isBeta` flag and instead have different config maps in K8 for the default and beta tables and use the the helm charts to decide what tables get passed into the environment.

My thought is that the `hmda-analytics` service only ever has to know about one set of tables. So, we could move this into configuration and out of the code, specifically removing the additional repository classes that are included in this PR.

PR #3149, opened 2019-10-04
return additional messaging information on parsing error

### LINKED ISSUE #3146: Formatting Error Language Change: "is not numeric"
Details: 

[HMDA-PLATFORM-UI-REF ](https://github.com/cfpb/hmda-platform-ui/issues/1280)
Updates will have to be made to the  ParserErrorModel.scala class to expand the JSON returned by this check to include ULI , row number , field name , value supplied , and the data type + list of acceptable values/range.

The following call is by the hmda-homepage-ui made when the filler logs into hmda-platform-ui and checks to see if they have and current/past filings for the selected year:
* API Call Example (GET):
locahost/institutions/B90YWS6AFX2LGWOXJ1LD/filings/2018

During the upload process the front also gets a submission response after file upload if applicable by getting the latest submission for the bank/year:\

API Call Example (GET):
locahost/filing/institutions/B90YWS6AFX2LGWOXJ1LD/filings/2018/submissions/latest

Now this response will change as the platform checks syntax/validatity/ maroc/ quality etc.

If the latest submission on this Upload page contains an error code it'll call an endpoint to get details of the error and uses that response JSON to display the parsing errors found
API Call Example (GET):	localhost/filing/institutions/B90YWS6AFX2LGWOXJ1LD/filings/2018/submissions/13/parseErrors

**We will need to make this backwards compatible with the current returned JSON**

kgudel:
Closes #3146


PatrickHSI:
@kgudel 

I'm assuming the change of this parse error return :

```
   "transmittalSheetErrors":[  
      "total lines is not numeric"
   ],
```

to this 

```
"fieldName": "Total Number of Entries Contained in Submission",
                    "inputValue": "",
                    "validValues": "Integer"
```

and we expect the front end to take care of the other half of the message in the platform-ui issue?

kgudel:
The specification for these changes primarily comes from [this ticket](https://github.com/cfpb/hmda-platform-ui/issues/1280) on the platform-ui. If we want to change what the exact messaging that is returned is that now pulls from a text file and so can be easily updated.

hmda/src/test/scala/hmda/validation/rules/lar/validity/common/V714Spec.scala
```diff
@@ -12,10 +12,9 @@ class V714Spec extends LarEditCheckSpec {
   property(
     "If Application Channel exemption election is taken, all fields must be exempt") {
     forAll(larGen) { lar =>
-      lar.mustPass
 
-      lar.copy(applicationSubmission = ApplicationSubmissionExempt).mustFail
-      lar.copy(payableToInstitution = PayableToInstitutionExempt).mustFail
+      lar.copy(applicationSubmission = ApplicationSubmissionExempt, payableToInstitution = PayableToInstitutionNotApplicable).mustFail
```
PatrickHSI: @KennethGudel should these test parameters be swapped? Application vs Payable?

hmda/src/test/scala/hmda/validation/rules/lar/validity/common/V710Spec.scala
```diff
@@ -11,8 +11,6 @@ class V710Spec extends LarEditCheckSpec {
 
   property("If credit is reported exempt, all fields must be reported exempt") {
     forAll(larGen) { lar =>
-      lar.mustPass
```
PatrickHSI: why was this removed from the test?

kgudel: That line tests that any randomly generated LAR should pass that edit. Which should never be the case when an edit is run. Only a subset of LARs should always pass or always fail this edit, which is tested below in that file. I am not sure why that line was added.

hmda/src/test/scala/hmda/validation/rules/lar/validity/common/V714Spec.scala
kgudel: Yes, application `applicationSubmission` should go with `ApplicationSubmissionExempt` and `payableToInstitution` should go with `PayableToInstitutionExempt`. I'm not sure how these got swapped.

PatrickHSI: @kgudel indeed they should

PatrickHSI: indeed they should @kgudel

PatrickHSI:
LGTM.

PR #3161, opened 2019-10-11
TS Regulator Date Column

### LINKED ISSUE #3110: Add a timestamp for TS Regulator file
We need to add a timestamp for regulator use, it was provided last year and needs to be added again.

The timestamp should be an additional column at the end of the TS file and should reflect the date and time that we received the submission from the filer.

PatrickHSI:
closes #3110


omnipresent07:
LGTM 👍

PR #3169, opened 2019-10-16
Fix for sign epoch intermittent difference 

### LINKED ISSUE #3068: Intermittent difference in the epoch date
There are intermittent difference in last two digits of the receipt number from the time the user signs and what is actually stored in the cassandra db

BarakStout:
Closes #3068

### DIFF
--- a/hmda/src/main/scala/hmda/persistence/submission/HmdaProcessingUtils.scala
+++ b/hmda/src/main/scala/hmda/persistence/submission/HmdaProcessingUtils.scala
@@ -53,26 +53,31 @@ object HmdaProcessingUtils {
     }
   }
 
-  def updateSubmissionReceipt(sharding: ClusterSharding, submissionId: SubmissionId, timestamp: Long, receipt: String, log: Logger)(
+  def updateSubmissionReceipt(sharding: ClusterSharding, submissionId: SubmissionId, timestamp: Long, receipt: String, modified: SubmissionStatus, log: Logger)(
     implicit ec: ExecutionContext,
     timeout: Timeout
   ): Unit = {
     val submissionPersistence: EntityRef[SubmissionCommand] =
       sharding.entityRefFor(SubmissionPersistence.typeKey, s"${SubmissionPersistence.name}-$submissionId")
 
+    val submissionManager =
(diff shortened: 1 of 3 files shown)


hmda/src/main/scala/hmda/persistence/submission/HmdaProcessingUtils.scala
```diff
@@ -53,26 +53,31 @@ object HmdaProcessingUtils {
...
 
+    val submissionManager =
+      sharding.entityRefFor(SubmissionManager.typeKey, s"${SubmissionManager.name}-$submissionId")
+
     val fSubmission: Future[Option[Submission]] = submissionPersistence ? (ref => GetSubmission(ref))
 
     for {
-      s <- fSubmission
-      m = s
+      m <- fSubmission
```
PatrickHSI: Since we're making changes here can we update to better variable names?

BarakStout: Good point

hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -283,9 +283,8 @@ object HmdaValidationError
                 .noEditsFound()) {
             Effect.persist(signed).thenRun { _ =>
               log.info(s"Submission $submissionId signed at ${Instant.ofEpochMilli(timestamp)}")
-              updateSubmissionStatus(sharding, submissionId, Signed, log)
-              updateSubmissionReceipt(sharding, submissionId, signed.timestamp, s"${signed.submissionId}-${signed.timestamp}", log)
-              publishSignEvent(submissionId).map(signed => log.info(s"Published signed event for $submissionId"))
+              updateSubmissionStatusAndReceipt(sharding, submissionId, timestamp, s"${signed.submissionId}-${signed.timestamp}", Signed, log)
+              publishSignEvent(submissionId).map(signed => log.info(s"Published signed event for $timestamp"))
```
PatrickHSI: Why was the log for the signed event publishing changed from including the  signed.submissionID and signed.timestamp to just the timestamp?

BarakStout: unintended consequence of debugging

PR #3184, opened 2019-10-21
Add LEI filter to data browser LEI

### LINKED ISSUE #3183: Add LEI filter to data browser
Add a filter for LEI to data-browser. This would be similar to state, msa, or nationwide and county.

The filter would apply to aggregates (redis) and CSV downloads (S3). As part of this ticket, indexes for LEI should also be created on the modifiedlar2018 table to ensure queries are fast.

PatrickHSI:
Dependent on PR #3182 

closes #3183


omnipresent07:
LGTM 👍

PR #3199, opened 2019-10-25
Upgrade Keycloak

kgudel:


kubernetes/keycloak/values-dev.yaml
```diff
@@ -61,10 +61,10 @@ keycloak:
   extraContainers: |
 
   ## Custom script that is run before Keycloak is started.
-  preStartScript: |
```
omnipresent07: Is there a reason `values-dev.yaml` doesn't have `script.sh: |` but `values.yaml` does?

kubernetes/keycloak/values.yaml
```diff
@@ -118,7 +119,7 @@ keycloak:
     - name: JDBC_PARAMS
       value: "ssl=true&sslfactory=org.postgresql.ssl.NonValidatingFactory"
     - name: FILING_APP_URL
-      value: "/filing/2018/"
+      value: "/filing/2018"
```
omnipresent07: This change doesn't seem to be there in `values-dev.yaml`.

zencircle:
name: keycloak_https_mapping
      ambassador_id: ambassador-default-1
      name: keycloak_https_mapping_
      ambassador_id: ambassador-default-2
Curious why we are using default-2, prod does not have default-2 ambassador id ?

PR #3248, opened 2019-11-06
added logging to VerifyQuality and VerifyMacro

### LINKED ISSUE #3196: Add more logging to platform
Add more logging to the platform so that it is easier to track status of a filer while parsing logs.

BarakStout:
Closes #3196

### DIFF
--- a/hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
+++ b/hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
@@ -259,6 +259,7 @@ object HmdaValidationError
             .thenRun { validationState =>
               val updatedStatus =
                 SubmissionStatus.valueOf(validationState.statusCode)
+              log.info(s"Verify Quality for $submissionId")
               updateSubmissionStatus(sharding, submissionId, updatedStatus, log)
               replyTo ! QualityVerified(submissionId, verified, updatedStatus)
             }
@@ -268,6 +269,7 @@ object HmdaValidationError
         }
 
       case VerifyMacro(submissionId, verified, replyTo) =>
+
         if (List(Macro.code, MacroErrors.code)
               .contains(state.statusCode) || !verified) {
           Effect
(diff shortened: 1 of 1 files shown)


hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -259,6 +259,7 @@ object HmdaValidationError
             .thenRun { validationState =>
               val updatedStatus =
                 SubmissionStatus.valueOf(validationState.statusCode)
+              log.info(s"Verify Quality for $submissionId")
```
omnipresent07: This should come after the line with `replyTo` since after that line the update would have happened and the text should be changed to `Verified Quality for <submissionId>`

hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
```diff
@@ -268,13 +269,15 @@ object HmdaValidationError
...
       case VerifyMacro(submissionId, verified, replyTo) =>
+
         if (List(Macro.code, MacroErrors.code)
               .contains(state.statusCode) || !verified) {
           Effect
             .persist(MacroVerified(submissionId, verified, SubmissionStatus.valueOf(state.statusCode)))
             .thenRun { validationState =>
               val updatedStatus =
                 SubmissionStatus.valueOf(validationState.statusCode)
+              log.info(s"Verify Macro for $submissionId")
```
omnipresent07: Same as above

hmda/src/main/scala/hmda/persistence/submission/HmdaValidationError.scala
BarakStout: In that case, should `SignSubmission` be updated as well?

omnipresent07: I believe that one is fine since `publishSignEvent` returns a `Future[Done]`.

Issue #3225, opened 2019-10-30
Update Data Publisher For Quarterly regulator file generation. 


PatrickHSI:
The data publisher needs to be updated to generate regulator ts, lar, and panel regulator files for quarterly filings.

PR #3275, opened 2019-11-24
make 2020 tables for quarterly filings

omnipresent07:
Helps with #3225


hmda-sql-doc/transmittalsheet2010.sql
```diff
@@ -0,0 +1,66 @@
+--
```
PatrickHSI: rename to "tansmittalsheet2020"

omnipresent07: @PatrickHSI addressed

PR #3276, opened 2019-11-24
Add email content for email

### LINKED ISSUE #3241: Come up with text to be included in emails received after signing
On signing the submission for an institution, the logged in user will receive an email for the institution. We should decided on the content of the email that will be sent out.

omnipresent07:
Closes #3241 

Adds html email capability. The text used is the same text from the signing page.

Screen Shot 2019-11-24 at 6 34 11 PM

### DIFF
--- a/email-service/src/main/resources/application.conf
+++ b/email-service/src/main/resources/application.conf
@@ -32,9 +32,9 @@ aws {
 hmda {
   lar {
     email {
-      subject = "You Filed your HMDA data!"
+      subject = "HMDA filing accepted!"
       subject = ${?EMAIL_SUBJECT}
-      content = "Your submission has been processed"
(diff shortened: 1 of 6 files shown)


email-service/src/main/resources/application.conf
```diff
@@ -32,9 +32,9 @@ aws {
 hmda {
   lar {
     email {
-      subject = "You Filed your HMDA data!"
+      subject = "HMDA filing accepted!"
       subject = ${?EMAIL_SUBJECT}
-      content = "Your submission has been processed"
+      content = "<b>HMDA filing accepted!</b><br>Congratulations, you have successfully completed your HMDA filing for <period>!<br>Your data and signature were received and recorded on <datetime>. Your receipt number for this submission is <uniq_id> "
```
kgudel: Should we change this in the `values.yaml` file/configmap as well or do we just want to remove that piece and not override what's here?

omnipresent07: @kgudel addressed

kgudel: 👍

PatrickHSI:
LGTM. waiting on input from @chynnakeys  and @MattPerrine

PR #3278, opened 2019-11-25
Sanitize Institution Inputs

### LINKED ISSUE #2669: Sanitize Inputs in Institutions Api
- Check that no matching tax-id exists when making or updating an institution through the institutions-api.
- Check that no matching LEIs exist
- Clean up any leading and trailing 0s, check length, and check position of `-` in tax ids

kgudel:
Closes #2669


hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -156,6 +166,29 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
     }
   }
 
+  private def validTaxIdFormat(taxIdOption: Option[String]): Boolean = {
```
PatrickHSI: You should be able to simplify this to a regex 

Example:

val taxIDPattern ="[0-9]{2}\\-[0-9]{7}$".r

taxID ="12-1234567"

taxIDPattern.matches(taxID)  // returns true

hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -156,6 +166,29 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
...
+    val taxId = taxIdOption.getOrElse("")
+    if (taxId.contains("-")) {
+      val id = taxId.split("-")
+      (id.length == 2 && id.head.length == 2 && id.last.length == 7)
+    } else false
+  }
+
+  val alphaNumeric = (('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')).toSet
```
PatrickHSI: this might also be a good spot for regex 

val leiPattern = "[A-Z0-9]{20}$"

hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -116,6 +79,53 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
       }
     }
 
+  private def postInstitution(institution: Institution, uri: Uri): Route = {
+    val institutionPersistence = InstitutionPersistence.selectInstitution(sharding, institution.LEI, institution.activityYear)
```
PatrickHSI: For this portion of the issue :
```
Check that no matching tax-id exists when making or updating an institution through the institutions-api.
```
Should we elevate tax-id to be a unique constraint as we do LEI?

PR #3287, opened 2019-11-26
Data Publisher 2020 quarterly Regulator Data

### LINKED ISSUE #3225: Update Data Publisher For Quarterly regulator file generation.

PatrickHSI:
closes #3225

### DIFF
--- a/hmda-data-publisher/src/main/scala/hmda/publisher/query/component/PublisherComponent2020.scala
+++ b/hmda-data-publisher/src/main/scala/hmda/publisher/query/component/PublisherComponent2020.scala
@@ -508,7 +508,8 @@ trait PublisherComponent2020 {
       if(includeQuarterly){
         db.stream(
           table
-            .filterNot(lar =>(lar.lei.toUpperCase inSet bankIgnoreList)&& !lar.isQuarterly  )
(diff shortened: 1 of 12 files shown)


hmda-data-publisher/src/main/scala/hmda/publisher/query/component/PublisherComponent2020.scala
```diff
@@ -0,0 +1,535 @@
...
+    def deleteByLei(lei: String): Future[Int] = {
+      db.run(table.filter(_.lei === lei).delete)
+    }
+    def count(): Future[Int] = {
+      db.run(table.size.result)
+    }
+    def getAllLARs(
+        bankIgnoreList: Array[String],includeQuarterly: Boolean): DatabasePublisher[LarEntityImpl2020] = {
+
+      if(includeQuarterly){
```
PatrickHSI: I'm not happy with this implementation however I'm concerned with the performance hit we run into nesting filter options to consolidate this into a single flow. Once we have larger amounts of quarterly data to test with I'd like to test this out more.

PR #3319, opened 2019-12-05
add message-frame-size for actors

### LINKED ISSUE #3313: Discarding oversized payload sent to Actor
Need to increase the message-frame-size because of the following error:

akka.remote.OversizedPayloadException: Discarding oversized payload sent to Actor[akka.tcp://hmda2@192.168.10.70:2552/temp/$Ecb]: max allowed size 128000 bytes, actual size of encoded class scala.Some was 178598 bytes.

omnipresent07:
Closes #3313 

1. Adds settings from application.conf to application-kubernetes.conf
2. Adds extra logging to track when the filing passes `validateTsLar` step

### DIFF
--- a/hmda/src/main/resources/application.conf
+++ b/hmda/src/main/resources/application.conf
@@ -35,6 +35,7 @@ akka {
 
   actor {
     provider = "cluster"
+    timeout = 8
   }
 
   cluster {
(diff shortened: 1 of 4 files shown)


hmda/src/main/resources/application.conf
```diff
@@ -35,6 +35,7 @@ akka {
 
   actor {
     provider = "cluster"
+    timeout = 8
```
helena: Did you mean `8s`? This seems short.

Issue #3346, opened 2019-12-11
Unimplemented deserialization for FilingState


omnipresent07:
```
[hmda-platform-f49f6d66f-x5qt9] [ERROR] [12/11/2019 07:34:26.243] [hmda2-akka.actor.default-dispatcher-2] [akka://hmda2/system/sharding/Filing/531/Filing-B90YWS6AFX2LGWOXJ1LD-2018][persistenceId:Filing-B90YWS6AFX2LGWOXJ1LD-2018][phase:rep
lay-snap] Persistence failure when replaying snapshot. Unimplemented deserialization of message with manifest [hmda.persistence.filing.FilingState] in [hmda.serialization.filing.FilingDetailsSerializer]
[hmda-platform-f49f6d66f-x5qt9] java.io.NotSerializableException: Unimplemented deserialization of message with manifest [hmda.persistence.filing.FilingState] in [hmda.serialization.filing.FilingDetailsSerializer]
[hmda-platform-f49f6d66f-x5qt9]         at hmda.serialization.filing.FilingDetailsSerializer.fromBinary(FilingDetailsSerializer.scala:30)
[hmda-platform-f49f6d66f-x5qt9]         at akka.serialization.Serialization.$anonfun$deserializeByteArray$1(Serialization.scala:233)
[hmda-platform-f49f6d66f-x5qt9]         at akka.serialization.Serialization.withTransportInformation(Serialization.scala:173)
[hmda-platform-f49f6d66f-x5qt9]         at akka.serialization.Serialization.deserializeByteArray(Serialization.scala:231)
[hmda-platform-f49f6d66f-x5qt9]         at akka.serialization.Serialization.$anonfun$deserialize$5(Serialization.scala:221)
[hmda-platform-f49f6d66f-x5qt9]         at scala.util.Try$.apply(Try.scala:209)
[hmda-platform-f49f6d66f-x5qt9]         at akka.serialization.Serialization.deserialize(Serialization.scala:213)
```

PR #3355, opened 2019-12-12
allow converting of filing state proto in filing details

kgudel:


omnipresent07:
closes https://github.com/cfpb/hmda-platform/issues/3346

PR #3387, opened 2019-12-21
Improvements to authentication

omnipresent07:
This PR does three things

1. Shows a log message when the file is fully received by the backend. Shows an error message "Entity stream truncation" along with submission-id when the file upload fails because of network error or user related error

2. Removed `reject` when authorization fails but instead sends a `403` complete. A way to test this would be to intentionally pass an incorrect authentication token and notice that the response back is a `403` as opposed to a `405`. 

3. Removes redundant `authorizeTokenWithLei` from SignHttpApi. This _could_ have been the reason for https://github.com/cfpb/hmda-frontend/issues/83


PatrickHSI:
Restarting travis.

omnipresent07:
@PatrickHSI this is addressed in the recent commit. GET on admin institution API does not need auth. Only POST, PUT, and DELETE do.

PR #3390, opened 2019-12-26
Search each institution table for email domain when no year is present

### LINKED ISSUE #3376: Update Institutions-API Email Domain Endpoint
Currently the endpoint to look up institutions by email domain only validates against institutions in the current filing season, this will need to be updated to check against all filing years or only validate against the email domain table. 

 - This bug prevents being able to create new filing accounts for previous filing seasons.

PatrickHSI:
closes #3376

### DIFF
--- a/institutions-api/src/main/scala/hmda/institution/api/http/InstitutionQueryHttpApi.scala
+++ b/institutions-api/src/main/scala/hmda/institution/api/http/InstitutionQueryHttpApi.scala
@@ -93,7 +93,7 @@ trait InstitutionQueryHttpApi extends HmdaTimeDirectives with InstitutionEmailCo
     path("institutions") {
       timedGet { uri =>
         parameter('domain.as[String]) { domain =>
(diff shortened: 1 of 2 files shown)


institutions-api/src/main/scala/hmda/institution/api/http/InstitutionQueryHttpApi.scala
```diff
@@ -93,7 +93,7 @@ trait InstitutionQueryHttpApi extends HmdaTimeDirectives with InstitutionEmailCo
     path("institutions") {
       timedGet { uri =>
         parameter('domain.as[String]) { domain =>
-          val f = findByEmail(domain, currentYear)
+          val f = findByEmailAnyYear(domain)
```
omnipresent07: If `currentYear` isn't being used, lets remove this variable along with removing it from K8 files so it is clearly visible that institutions-api no longer makes use of the `currentYear` (which comes from timed guards config map)

PR #3408, opened 2020-01-07
Throw parsing error for LEI and/or ULI with space in it

### LINKED ISSUE #3407: Errors when LEI and/or ULI have spaces in them
When LEI has spaces, a SV error is thrown. Rather parsing error should be thrown

When ULI has spaces in it, the validation halts with an error "ULI is not alphanumeric". This error comes while triggering V609. Rather a paring error should be thrown when there is a space in ULI

omnipresent07:
Closes #3407 

This is how the parserError json will respond after this PR

```
    "larErrors": [
        {
            "rowNumber": 2,
            "estimatedULI": "2 Y9CEMKWIJCYXQ2T4ZRWGY",
            "errorMessages": [
                {
                    "fieldName": "LEI",
                    "inputValue": "BANK1LEIFORTES T12345",
                    "validValues": "String with no space"
                },
                {
                    "fieldName": "ULI",
                    "inputValue": "2 Y9CEMKWIJCYXQ2T4ZRWGY",
                    "validValues": "String with no space"
                },
                {
                    "fieldName": "Loan Type",
                    "inputValue": "S",
                    "validValues": "Integer: 1, 2, 3, 4"
                }
            ]
        }
    ],
```

### DIFF
--- a/common/src/main/resources/parserFieldValidValues.txt
+++ b/common/src/main/resources/parserFieldValidValues.txt
@@ -1,3 +1,5 @@
+LEI|String with no space
+ULI|String with no space
 LAR Record Identifier|Integer: 2
 Application Date|Integer or NA
 Loan Type|Integer: 1, 2, 3, 4
--- a/common/src/main/resources/parserFieldValidValues.txt
+++ b/common/src/main/resources/parserFieldValidValues.txt
@@ -1,5 +1,5 @@
-LEI|String with no space
-ULI|String with no space
+LEI|String with only letters and numbers
+ULI|String with only letters and numbers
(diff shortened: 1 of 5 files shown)


common/src/main/resources/parserFieldValidValues.txt
```diff
@@ -1,3 +1,5 @@
+LEI|String with no space
```
kgudel: Since we're checking for more than just a space let's make this a bit more general. How about "String with only letters and numbers". Something like that?

omnipresent07: addressed

PR #3454, opened 2020-01-29
Updating Institution api call PUT to create if not found

### LINKED ISSUE #3361: admin institutions api endpoint put should do a create or update
When institution loader needs to be run it would be ideal to run a put such that if the institution doesn't exist it automatically gets created

PatrickGoRaft:
closes #3361

### DIFF
--- a/hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
+++ b/hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
@@ -53,7 +53,7 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
             sanatizeInstitutionIdentifiers(institution, true, uri, postInstitution)
           } ~
             timedPut { uri =>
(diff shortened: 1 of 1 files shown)


hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -53,7 +53,7 @@ trait InstitutionAdminHttpApi extends HmdaTimeDirectives {
             sanatizeInstitutionIdentifiers(institution, true, uri, postInstitution)
           } ~
             timedPut { uri =>
-              sanatizeInstitutionIdentifiers(institution, false, uri, putInstitution)
+              sanatizeInstitutionIdentifiers(institution, true, uri, putInstitution)
```
PatrickGoRaft: @kgudel / @Kibrael  was the logic behind this due to the case in which there might be an institution with an LEI of an invalid length already in the ecosystem? If so I can adjust this so on `PUT` we can check the LEI only when not found and creating a new institution.

PR #3480, opened 2020-02-07
IRS Publisher: Remove Census API Dependency

### LINKED ISSUE #3479: Remove dependency of irs service from census
The census micro-service is deprecated and the logic for it now lives in `common/` . Irs microservice should communicate with the census service via `common/`

PatrickGoRaft:
closes #3479


omnipresent07:
This LGTM 👍

PR #3482, opened 2020-02-10
change pod-reader name

### LINKED ISSUE #3483: Fix RBAC Issue on platform
When platform is deployed to beta and then deployed to default, the former latter takes precedence over the cluster binding service account. This stops us from being able to scale the platform 0 to 3 or perform a rolling upgrade. We're forced to helm delete and install again.

omnipresent07:
closes #3483 

After this PR, each time the platform is deployed to default or beta the following two rules will always respond in `yes`

```
$ kubectl auth can-i list pods --as=system:serviceaccount:beta:hmda-service-account-beta -n beta
$ yes
$ kubectl auth can-i list pods --as=system:serviceaccount:default:hmda-service-account-default -n default
$ yes
```


PatrickGoRaft:
LGTM.

PR #3536, opened 2020-03-13
CSV File Generation for Institutions Table by Year

### LINKED ISSUE #3465: Create endpoint to provide file for institution loader
The institutions api shall have an endpoint that can be used to fetch a CSV file that is used as an input to the institution loader. At the moment this file is generated manually by the developers. However, it should be generated via an API and placed on an internal S3 bucket. This endpoint would only be accessible internally.

PatrickGoRaft:
closes #3465


institutions-api/src/main/scala/hmda/institution/api/http/InstitutionQueryHttpApi.scala
```diff
@@ -120,12 +139,50 @@ trait InstitutionQueryHttpApi extends HmdaTimeDirectives with InstitutionEmailCo
...
+   val i = Future.sequence(possibleInstitutions)
+
+    onComplete(i) {
+      case Success(institutions) =>
+
+        if (institutions.isEmpty || institutions == None) {
+          returnNotFoundError(uri)
+        } else {
+          val institutionSource = Source.fromIterator(() =>institutions.map(institution => institution + "\n").toIterator)
+          val headerSource = Source.fromIterator(() =>
```
PatrickGoRaft: I should extract this to a static reference

PR #3558, opened 2020-03-25
Added a value to not trigger q649

### LINKED ISSUE #3535: Q649_2 not handling 9999
Hello, the FIG for Q649 says `2) If Credit Score of Co-Applicant or Co-Borrower does
not equal 7777, 8888, 9999, or 1111, Credit Score
should generally be between 300 and 900.`, but the code is not checking for 9999.  So submitting with CoApp Score = 9999 and App Score  = 783 is giving an incorrect result.

BarakStout:
Closes #3535


PatrickGoRaft:
LGTM.

PR #3585, opened 2020-04-01
remove filing status updated spec

### LINKED ISSUE #3303: FilingStatus does not change
Filing status can take different options : in-progress, completed, and cancelled. However, even when a submission is signed, the status does not change to completed. Need to check with front end whether this is used at all.

omnipresent07:
closes #3303


PatrickGoRaft:
LGTM.

PR #3612, opened 2020-04-23
Do not evaluate NA or Exempt for Q653

### LINKED ISSUE #3584: Q653 not handling Exempt
The platform is flagging Q653 when the CLTV has been set to `Exempt`.  Is this the expected behavior?

### LINKED ISSUE #3609: Update Q653 to handle NA/Exempt
Currently, based on the way the edit is written, Q653 does not handle NA or Exempt cases: 1) If Action Taken equals 1, 2, or 8, the CLTV should generally be between 0 and 250.
2) If Action Taken equals 3, 4, 5, 6, or 7, the CLTV should generally be between 0 and 1,000.

This edit logic should be updated to handle these two cases -
1) If Action Taken equals 1, 2, or 8, and CLTV is not reported NA or Exempt, the CLTV should generally be between 0 and 250.
2) If Action Taken equals 3, 4, 5, 6, or 7, and CLTV is not reported NA or Exempt, the CLTV should generally be between 0 and 250.

kgudel:
Closes #3609 
Closes #3584


omnipresent07:
👍 LGTM

PR #3635, opened 2020-05-05
docker-compose for hmda services 

### LINKED ISSUE #2695: Create docker compose file
Create a docker compose file to bring up all containers with one command.

BarakStout:
Closes #2695


hmda/src/main/scala/hmda/HmdaPlatform.scala
```diff
@@ -64,11 +66,16 @@ object HmdaPlatform extends App {
   implicit val mat     = Materializer(system)
   implicit val cluster = Cluster(system)
 
-  if (runtimeMode == "dcos" || runtimeMode == "kubernetes") {
+  if (runtimeMode == "dcos" || runtimeMode == "kubernetes" || runtimeMode == "docker-compose") {
     ClusterBootstrap(system).start()
     AkkaManagement(system).start()
   }
 
+  // if (runtimeMode == "docker-compose") {
```
omnipresent07, line 74: Could we remove these comments?

BarakStout: Good catch. Done.

PR #3727, opened 2020-06-19
updates to change institution assets field from int to long

### LINKED ISSUE #3687: Change Institutions Field `Assets` from Int to Long
Scala currently limits type Integer values to the maximum value of `2147483647` 

https://www.scala-lang.org/api/2.9.1/scala/Int$.html

Current institution data is forcing this value beyond this limit.  To compensate, this field needs to be converted to type Long.

PatrickGoRaft:
closes #3687


omnipresent07:
👍 LGTM

PR #3771, opened 2020-07-10
Ordered submission history request

### LINKED ISSUE #3368: Change filings endpoint to return submissions in an order
the frontend assumes that the filings endpoint returns submissions in order of sequenceNumber. However, there is no guarantee of that on the backend. We should update this endpoint such that the sequenceNumber ordering is guaranteed and the four flags `qualityVerified` etc are only available in the first array of the submissions array.

PatrickGoRaft:
closes #3368


omnipresent07:
LGTM 👍

PR #3784, opened 2020-08-03
make joining between lar and mlar easier

### LINKED ISSUE #3766: Create Primary Key mapping between loan application register and modifiedlar PG table
modify the `modified-lar` microservice logic that populates the internal `modifiedlar` PG table such that the table can be joined with `loanapplicationregister` based on primary key.

omnipresent07:
closes #3766

### DIFF
--- a/common/src/main/scala/hmda/query/repository/ModifiedLarRepository.scala
+++ b/common/src/main/scala/hmda/query/repository/ModifiedLarRepository.scala
@@ -154,13 +154,15 @@ class ModifiedLarRepository(databaseConfig: DatabaseConfig[JdbcProfile]) {
             sex_categorization,
             percent_median_msa_income,
             dwelling_category,
-            loan_product_type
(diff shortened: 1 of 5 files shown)


common/src/main/scala/hmda/query/repository/ModifiedLarRepository.scala
```diff
@@ -263,7 +265,8 @@ class ModifiedLarRepository(databaseConfig: DatabaseConfig[JdbcProfile]) {
             ${input.mlar.sexCategorization},
             ${incomeCategorization(input.mlar.income, input.census.medianIncome)},
             ${input.mlar.dwellingCategorization},
-            ${input.mlar.loanProductTypeCategorization}
+            ${input.mlar.loanProductTypeCategorization},
+            ${input.mlar.uli}
```
kgudel, line 269: Should uli be included twice in this object?

PR #3815, opened 2020-08-31
Fix validation tracker package name and add reference for 2020

### LINKED ISSUE #3814: File upload broken in validation tracker
The validation tracker package name seems to be named incorrectly which is breaking the file upload mid way through.

omnipresent07:
closes #3814


kgudel:
LGTM

PR #3827, opened 2020-09-04
Added action_taken_date to mlar

### LINKED ISSUE #3824: Further expand the primary key for modifiedlar microservice
In order to be able to get unique records from the modifiedlar PG tables, we need to further expand the primary key such that no duplicates are represented in the data set.

BarakStout:
Closes #3824


hmda-analytics/src/test/scala/hmda/analytics/query/LarComponentSpec.scala
```diff
@@ -1,35 +1,33 @@
...
-      val sampleLar = larGen.map(LarConverter(_, year = 2019)).sample.get.copy(isQuarterly = true)
-
-      repo
-        .insert(sampleLar)
-        .map(_ mustBe 1)
-        .flatMap(_ => repo.deletebyLeiAndQuarter(sampleLar.lei))
-        .map(_ mustBe 1)
-    }
-  }
+//  "LarComponent" must {
```
omnipresent07, line 8: @BarakStout Can we remove this commented out logic if it isn't being used?

omnipresent07 (changes requested):

PR #3833, opened 2020-09-10
insert data for quarters into separate tables

### LINKED ISSUE #3826: Modify hmda-analytics microservice to add data to PG based on submission period
Currently hmda-analytics microservice replaces the data based on the submission period only for years. We need a way to do the same for the quarters such that the data is stored in different PG tables per quarters.

omnipresent07:
closes #3826


PatrickGoRaft (approved):
LGTM.

PR #3920, opened 2020-11-20
skip kafka records that don't conform to a standard

### LINKED ISSUE #3898: Add fault tolerance to microservices
Each downstream microservice that reads records off of kafka must ignore values when they aren't in the format of `<key>:<value>` i.e. `B90:B90-2018-33`.

omnipresent07:
closes #3898


BarakStout (approved):
LGTM 👍

PR #3943, opened 2020-12-10
extract bank filter list to common

### LINKED ISSUE #3832: Extract bank-filter-list config into common
bank-filter-list configurations are separately stored in each microservice which is hard to manage. We should store these into `common/reference.conf` so each microservice has access to the same config for bank filters.

omnipresent07:
closes #3832


PatrickGoRaft (approved):
LGTM

PR #3992, opened 2021-01-07
Prevent Institutions from Being Create or Modified to Include a public Email Domain

kgudel:
Adds:
1. A list of public email domains in common
2. Checks the above list when an institution is created or modified to prevent inclusion of public domains in institution records
3. Prevents any results being returned when a public email domain is search in the institutions-api

### DIFF
--- a/common/src/main/resources/reference.conf
+++ b/common/src/main/resources/reference.conf
@@ -16,6 +16,11 @@ hmda {
     timeout = 38
   }
 
+  email {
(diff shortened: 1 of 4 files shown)


common/src/main/resources/reference.conf
```diff
@@ -16,6 +16,11 @@ hmda {
...
+    publicDomains = ${?PUBLIC_EMAIL_DOMAINS}
```
PatrickGoRaft, line 21: @kgudel  we will need to define a configmap for this environment variable

common/src/main/scala/hmda/api/http/EmailUtils.scala
```diff
@@ -0,0 +1,16 @@
+package hmda.api.http
+
+import com.typesafe.config.ConfigFactory
+
+// $COVERAGE-OFF$
+object EmailUtils {
+  val config = ConfigFactory.load("reference.conf")
+  val publicDomains = config.getString("hmda.email.publicDomains").split(",")
+
```
PatrickGoRaft, line 9: @kgudel  the checks here will need to be case insensitive.

institutions-api/src/main/scala/hmda/institution/api/http/InstitutionQueryHttpApi.scala
```diff
@@ -128,8 +129,12 @@ private class InstitutionQueryHttpApi(config: Config)(implicit ec: ExecutionCont
   path("institutions") {
     (extractUri & get) { uri =>
       parameter('domain.as[String]) { domain =>
+        if (checkIfPublicDomain(domain)) {
```
PatrickGoRaft, line 132: we should protect all endpoints looking things up by domain

PatrickGoRaft (changes requested):

hmda/src/main/scala/hmda/api/http/admin/InstitutionAdminHttpApi.scala
```diff
@@ -233,5 +236,7 @@ private class InstitutionAdminHttpApi(config: Config, sharding: ClusterSharding)
       complete((StatusCodes.BadRequest, "Incorrect lei format"))
     } else if (checkAgencyCode && !validAgencyCodeFormat(institution.agency.code)) {
       complete((StatusCodes.BadRequest, "Incorrect agency code format"))
+    } else if (checkListIfPublicDomain(institution.emailDomains)){
```
PatrickGoRaft, line 239: should remove whitespace and and make this case insensitive

common/src/main/resources/reference.conf
kgudel: I don't think it's necessary to make this a config map at the moment as this list should not change.

PatrickGoRaft (approved):
LGTM.

PR #3995, opened 2021-01-19
Update mview exemptions

BarakStout:


meissadia:
Results in the Dashboard look good to me 👍🏽

PR #4003, opened 2021-02-02
dashboard endpoints updates

BarakStout:
- Fixes weekly cutoffs to be Sunday <-> Saturday 
- Adds City, State and sign_date to top filers


meissadia (approved):
Looks great!

PR #4023, opened 2021-02-22
add estimated tag to dashboard results from mviews

BarakStout:
Dashboard results using mviews will have `estimated results` in the result


PatrickGoRaft (approved):
LGTM.

PR #4072, opened 2021-04-05
only remove lei when dealing with PG and uppercase the submission id

### LINKED ISSUE #4058: Make modified-lar regeneration more effecient
Usually when modified-lar generates records based on kafka records, it must delete the lei from PG table prior to adding the records. This delete process takes time because the PG table is quite big. When we re-generate the data there is no need to perform a delete since there will not be any data in the PG table to begin with and the lei and submission ids will be unique.

### LINKED ISSUE #4059: Process kafka records in mixed camel case
If a kafka ID is in mixed camel case, it won't pick up any data from Cassandra since the LEI's there are all capitalized. Prior to sending the id, we should capitalize the case for the kafka id

omnipresent07:
closes #4058 
closes #4059


BarakStout (approved):
LGTM 👍

PR #4107, opened 2021-05-07
Update Email Language in Keycloak

### LINKED ISSUE #4090: Improve "Forgot Password" Language
Displaying an improved message when a user requests a password for an account that does not exist. Users who don’t have an account may not realize it since the message does not indicate this.

Suggested language: **_If you have an account on the HMDA platform_**, you should receive an email . . . further instructions. _**You will not receive an email if there is no account associated with the email address you have entered.**_

image

kgudel:
Closes #4090


PatrickGoRaft (approved):
lgtm

PR #4165, opened 2021-07-20
replace default with beta

### LINKED ISSUE #4155: Remove hard-coded "default" namespace
The hmda validation error has a hard coded namespace "default" . We should remove that so that the namespace can be dynamic

omnipresent07:
closes #4155


PatrickGoRaft (approved):
LGTM

PR #4171, opened 2021-08-04
Allow running sbt when there are no git tags in the checkout

### LINKED ISSUE #4163: Allow running sbt when there are no git tags in the checkout
When the hmda-platform is checked out into an environment disconnected from referencing the original hmda-platform repo, their is a bug that prevents `sbt` from executing without tagged github versions.

omnipresent07:
closes #4163


BarakStout (approved):
LGTM 👍

PR #4207, opened 2021-09-28
configmap fails to apply

zencircle:
ConfigMap in version "v1" cannot be handled as a ConfigMap: v1.ConfigMap.Data: ReadString: expects


PatrickGoRaft (approved):
LGTM

PR #4211, opened 2021-09-30
Updates to check digit to better communicate errors in batch uploads

### LINKED ISSUE #4193: Check Digit Batch Generation/Validation update
- If a file contains a ULI that has less than 23 characters, or one that is alphanumeric, the return validation removes this value as opposed to returning an "invalid" or False value. This should be updated to show an invalid value message. - - 
 - Additionally, we can add a message to the frontend instructions that ULIs should be alphanumeric and more than 23 characters to be batch validated. In order to batch generate check digits, LEIs should be alphanumeric and more than 20 characters.

PatrickGoRaft:
closes #4193


lchen-2101 (approved):

lchen-2101:
LGTM

PR #4262, opened 2021-11-12
Update ambassador-rbac-eks.yaml

zencircle:
- remove CRDs from the rbac file, add appropriate affinity rules and bump version 
- remove unwanted files


RoachMJ (approved):
lgtm

PR #4263, opened 2021-11-15
update data-browser-api deployment

zencircle:
Add tls and auth


lchen-2101:
I'm guessing the values file is copied over from the redis sample?

zencircle:
> m guessing the values file is copied over from the redis sample?

Yes, copied from the upstream and   added `appendonly no` https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml

lchen-2101 (approved):
LGTM

PR #4270, opened 2021-11-17
Generalize Auth Rules and make Beta Users

kgudel:


common/src/main/scala/hmda/auth/OAuth2Authorization.scala
```diff
@@ -15,98 +15,69 @@ import org.keycloak.adapters.KeycloakDeploymentBuilder
...
-        }
-    }
+  protected def authorizeTokenWithRuleReject(passing: Boolean, token: VerifiedToken): Directive1[VerifiedToken] = {
+        if (passing) provide(token)
+        else reject(AuthorizationFailedRejection).toDirective[Tuple1[VerifiedToken]]
+  }
 
-  protected def authRejectionHandler: RejectionHandler =
+  protected def authRejectionHandler(rejectionMessage: String): RejectionHandler = {
+    println("authRejectioHandler")
```
lchen-2101, line 62: still need to digest, but println's should be replaced

PR #4279, opened 2021-11-19
2022 Quarterly edit V665_1 and V665_2

### LINKED ISSUE #4232: 2022 Edits: Update V665-2
Edit #: V665 (Existing edit with changes to edit logic)
Edit Type: Validity
Category: LAR
Data Fields: Credit Score

Instructional Text for Users: An invalid Credit Score data field was reported. Please review the information below and update your file accordingly.

Edit Logic: Change to edit logic #2 to add new credit scoring model value (new changes in bold).

1) Credit Score of Co-Applicant or Co-Borrower must be a number, and cannot be left blank.
2) Co-Applicant or Co-Borrower, Name and Version of Credit Scoring Model must equal 1111, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, or **11** and cannot be left blank.

PatrickGoRaft:
closes #4232


hmda/src/main/scala/hmda/validation/rules/lar/validity/2022/V665_2.scala
```diff
@@ -0,0 +1,17 @@
...
+import hmda.validation.dsl.ValidationResult
+import hmda.validation.rules.EditCheck
+
+object V665_2 extends EditCheck[LoanApplicationRegister] {
+  override def name: String = "V665-2"
+
+  override def parent: String = "V665"
+
+  override def apply(lar: LoanApplicationRegister): ValidationResult =
+    lar.coApplicant.creditScoreType not equalTo(new InvalidCreditScoreCode)
```
kgudel, line 16: I think it would be best if we checked against a full list of available options incase new options are added in the future.

For example, if an option 12 was added next year this edit would allow that option for 2022 which would be incorrect.

kgudel (changes requested):

hmda/src/main/scala/hmda/validation/rules/lar/validity/2022/V665_2.scala
PatrickGoRaft: @kgudel  Pushed up a change for this

kgudel (changes requested):
The 2022 version is looking good!

But could we include the change to innumerate the valid options in common version of V665 in this PR as well?

PatrickGoRaft:
> The 2022 version is looking good!
> 
> But could we include the change to innumerate the valid options in common version of V665 in this PR as well?

@kgudel  this has been added to the commit as well now

kgudel (approved):

PR #4303, opened 2021-12-04
annuals should not have quarter

omnipresent07:
Annual filings should not have quarter flag . Closes HMDA-Operations/hmda-devops/issues/2883


PatrickGoRaft (approved):
LGTM.

PR #4324, opened 2021-12-15
turn gap free read to off mitigating inconsistent sequence numbers

lchen-2101:


hmda/src/main/resources/persistence.conf
```diff
@@ -37,6 +37,10 @@ cassandra-journal {
   ssl.truststore.password = ${?TRUSTSTORE_PASSWORD}
 }
 
+cassandra-query-journal {
+  gap-free-sequence-numbers = off
```
PatrickGoRaft, line 41: Lets include the details of what this setting does 

```
  # Sequence numbers for a persistenceId is assumed to be monotonically increasing
  # without gaps. That is used for detecting missing events.
  # In early versions of the journal that might not be true and therefore
  # this can be relaxed by setting this property to off.
```

PatrickGoRaft (changes requested):

hmda/src/main/resources/persistence.conf
lchen-2101: yeap, added reference link and description

RoachMJ (approved):
Looks good.

PR #4342, opened 2022-01-04
log the exception when uploading fails

### LINKED ISSUE #4341: Log the error when uploading failed appears
When "uploading failed for (submission id)" message appears in the logs, the exception is not logged; update so it is logged.

lchen-2101:
closes #4341


PatrickGoRaft (approved):
LGTM.

PR #4349, opened 2022-01-06
log response when request time out occurred, if process finishes.

### LINKED ISSUE #4335: add logging when api process responds after akka http default timeout of 20 seconds
akka http has a default timeout of 20 seconds for api calls, and will respond with 598 after that timeout period, we can look into increasing that limit, as well as add logging for processes that responds after that limit (failures in particular)

lchen-2101:
closes #4335


PatrickGoRaft (approved):
LGTM

PR #4365, opened 2022-01-20
fix keycloak config variable substitutions

### LINKED ISSUE #4364: Fix build with invalid configuration substitutions.
Locally running the platform fails with undefined environment variables that are supposed to be optional

lchen-2101:
closes #4364


PatrickGoRaft (commented):
LGTM

PR #4369, opened 2022-01-24
Add More Robust Parser Check for Date

### LINKED ISSUE #4367: Make invalid field date parsing error visible
On a few occasions, the submission contains invalid date fields causing parsing exception to occur, e.g. `202103015`; we'd like to make this more visible to the customers / support. Could expand on the current file format verification tool?

kgudel:
Closes #4367

### DIFF
--- a/common/src/main/resources/parserFieldValidValues.txt
+++ b/common/src/main/resources/parserFieldValidValues.txt
@@ -1,7 +1,7 @@
 LEI|String with only letters and numbers
 ULI|String with only letters and numbers
 LAR Record Identifier|Integer: 2
-Application Date|Valid Date in format yyyyMMdd or NA
+Application Date|Valid Date in format YYYYMMDD or NA
 Loan Type|Integer: 1, 2, 3, 4
 Loan Purpose|Integer: 1, 2, 31, 32, 4, 5
 Preapproval|Integer: 1, 2
--- a/common/src/main/resources/parserFieldValidValues.txt
+++ b/common/src/main/resources/parserFieldValidValues.txt
@@ -1,14 +1,14 @@
 LEI|String with only letters and numbers
 ULI|String with only letters and numbers
 LAR Record Identifier|Integer: 2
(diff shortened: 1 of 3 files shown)


common/src/main/resources/parserFieldValidValues.txt
```diff
@@ -1,14 +1,14 @@
 LEI|String with only letters and numbers
 ULI|String with only letters and numbers
 LAR Record Identifier|Integer: 2
-Application Date|Integer or NA
+Application Date|Valid Date in format yyyyMMdd or NA
```
PatrickGoRaft, line 4: @KennethGudel  are we sure these are the correct date formate regex i.e. yyyy vs YYYY

common/src/main/resources/parserFieldValidValues.txt
```diff
@@ -1,14 +1,14 @@
...
 Loan Type|Integer: 1, 2, 3, 4
 Loan Purpose|Integer: 1, 2, 31, 32, 4, 5
 Preapproval|Integer: 1, 2
 Construction Method|Integer: 1, 2
 Occupancy Type|Integer: 1, 2, 3
 Action Taken|Integer: 1, 2, 3, 4, 5, 6, 7, 8
-Action Taken Date|Integer
+Action Taken Date|Valid Date in format YYYYMMDD
```
PatrickGoRaft, line 11: @KennethGudel  are we sure these are the correct date formate regex i.e. yyyy vs YYYY

PatrickGoRaft (changes requested):

common/src/main/resources/parserFieldValidValues.txt
kgudel: This text is only used for the error message that is displayed to users. @chynnakeys suggested that using all caps would be less confusing to users.

PatrickGoRaft (approved):
LGTM.

PR #4392, opened 2022-02-25
trim each element of the lei list coming from keycloak

### LINKED ISSUE #4382: Session Expired error when there is a space in LEI
When there is a space between LEIs in Keycloak, the platform forces a reload and does not allow you to login. 
https://ffiec.cfpb.gov/filing/2021/?session=expired

image

Related: GHE/DevOps #2414

omnipresent07:
If there are spaces in the LEI list in the keycloak attribute, trim all the leading and ending spaces. closes #4382


PatrickGoRaft (approved):
LGTM

Issue #4400, opened 2022-03-09
Testing: test / load test post annual filing season branch


lchen-2101:
Testing: test / load test on updated akka cassandra plugin

PR #4408, opened 2022-03-18
minor version scala upgrade instead of major, change log level for ak…

lchen-2101:
relates #4400

### DIFF
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -5,7 +5,7 @@ import sbtassembly.AssemblyPlugin.autoImport._
 object BuildSettings {
   val buildOrganization = "cfpb"
   val buildVersion      = "2.0.0"
-  val buildScalaVersion = "2.13.8"
+  val buildScalaVersion = "2.12.15"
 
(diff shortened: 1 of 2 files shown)


project/BuildSettings.scala
```diff
@@ -5,7 +5,7 @@ import sbtassembly.AssemblyPlugin.autoImport._
 object BuildSettings {
   val buildOrganization = "cfpb"
   val buildVersion      = "2.0.0"
-  val buildScalaVersion = "2.13.8"
+  val buildScalaVersion = "2.12.15"
```
PatrickGoRaft, line 8: @lchen-2101  is this version rollback intentional?

lchen-2101: yes sir, 2.12.x to 2.13.x is a major version update, there are other dependencies that no longer works and needs to be updated alongside the 2.13.x update; there are also other code breaking changes by going to 2.13.x; I couldn't build successfully with the change of just the scala version update.

PatrickGoRaft (approved):
LGTM.

PR #4428, opened 2022-04-22
Fall Back to Using County for Census

### LINKED ISSUE #4427: Determine Median Family Income Percentage At County Level
Median Family Income is determined on the county, rather the tract level. However, there are currently many fields where the `median_family_income` is set to `0` even though the record relates to a field with a county reported.

We should update the code to ensure that any records with a reported county have a non-zero `median_family_income`.

kgudel:
Closes #4427


PatrickGoRaft (approved):

PatrickGoRaft:
LGTM.

PR #4489, opened 2022-07-05
Update to remove wrapped quotation marks from post processing

### LINKED ISSUE #4482: Data Publisher CSV  Double Quotes Check Before Export
There's a condition in which  strings already escaped to double quotes could be wrapped into another set of double quotes when exporting to CSV format.

PatrickGoRaft:
closes #4482


lchen-2101 (approved):
LGTM

PR #4490, opened 2022-07-06
added institution lar count endpoint; refactored future completion

### LINKED ISSUE #4481: Institution Endpoint for LAR Totals
This would be a new endpoint or addition to the existing institution endpoint that allows one to query by LEI and get the number of LAR in their latest signed submission by year.

This data should query the TS table and get the LAR count field.

lchen-2101:
closes #4481


PatrickGoRaft:
LGTM.

PR #4497, opened 2022-07-14
S3proxy

kgudel:
Addresses #4516


meissadia:
## Modified LAR
 
We also need to support MLAR files w/ header row
|Proxy|S3|
|---|----|
|/file/modifiedLar/year/<year\>/institution/<lei\>/txt|https://s3.amazonaws.com/cfpb-hmda-public/prod/modified-lar/<year\>/<lei\>.txt|
|/file/modifiedLar/year/<year\>/institution/<lei\>/htxt|https://s3.amazonaws.com/cfpb-hmda-public/prod/modified-lar/<year\>/<lei\>_header.txt|

PatrickGoRaft (changes requested):

file-proxy/src/main/scala/hmda/uli/FileProxy.scala
```diff
@@ -0,0 +1,33 @@
+package hmda.fileProxy
```
PatrickGoRaft, line 1: The package names don't correspond to the directory structure.

file-proxy/src/main/resources/application.conf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+akka.http.parsing {
+  max-to-strict-bytes = 20m
+}
+akka.http.server.parsing {
+  max-content-length = 20m
+}
+
+akka.http.server.request-timeout = 4888888 seconds
```
PatrickGoRaft, line 13: Let's try setting this to `infinite`  to see if we can get around just setting some arbitrarily large value  ( I know we're doing this in other places in the code it'll be good to test it out here).

file-proxy/src/main/scala/hmda/uli/api/http/ProxyHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private def retrieveData(path: String): Future[Option[Source[ByteString, NotUsed]]] = {
+    val timeout: Timeout = Timeout(config.getInt("hmda.http.timeout").seconds)
+    S3.download("cfpb-hmda-public", path).withAttributes(S3Attributes.settings(s3Settings)).runWith(Sink.head)
+      .map(opt => opt.map { case (source, _) => source })
+  }
+
+  private def streamingS3Route(s3Key: String): Route = {
+    val fStream: Future[Source[ByteString, NotUsed]] = retrieveData(s3Key).flatMap {
```
PatrickGoRaft, line 124: A map without later creating a Future should suffice here. Something we can note in a future refactor

PatrickGoRaft (changes requested):

PR #4580, opened 2022-12-07
reduce kubernetes java client logging

### LINKED ISSUE #4578: Decrease logging by Kubernetes API Java Client library
Current kubernetes api java client logs whenever resource watch expires; and need to be re-watched, which is not important, so remove these from logging.

lchen-2101:
closes #4578


PatrickGoRaft (approved):
LGTM

PR #4608, opened 2023-01-11
[quarterly graphs] Update graph descriptions

### LINKED ISSUE #4582: Quarterly Graphs Verbiage update
update quarterly graphs verbiage for subtitles

meissadia:
Closes #4582 

Updates language for Graphs UI with the text provided in https://github.com/cfpb/hmda-platform/issues/4582#issuecomment-1378833342

### DIFF
--- a/hmda-quarterly-data-service/src/main/resources/verbiage.conf
+++ b/hmda-quarterly-data-service/src/main/resources/verbiage.conf
@@ -160,7 +160,7 @@ graph {
       label: "Median DTI"
       label: ${?DTI_LABEL}
       by_type {
-        title: "How has median DTI changed?"
+        title: "How have median DTIs changed?"
         title: ${?DTI_TITLE}
         subtitle: "Median DTIs have generally risen for all loan types since 2021. FHA loans have the highest median DTI."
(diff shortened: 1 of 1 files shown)


kgudel (changes requested):
This all looks good except I'm not finding the text for `How have median DTI changed?`

meissadia:
@kgudel 
> This all looks good except I'm not finding the text for `How have median DTI changed?`

https://github.com/cfpb/hmda-platform/pull/4608/files#diff-e1e4b0a465fc75a0afebbf4105e856f5d04fb70e1f594cffdca4ecc20dde016cR163-R165
 
I missed the change in the questions language, but the updated version doesn't sound right to me.
If DTI = "Debt-to-Income Ratio" then it seems the question should either read
- How `has` median `DTI` changed? (if we're referring to a single, combined measure of D-to-I)
- How `have` median `DTIs` changed (if we're referencing the multiple DTIs displayed in the graph)

Since the new chart description starts by referencing all DTIs ("Median DTIs have generally...") it feels best to go with the latter version, so I've updated to the PR to reflect that.

kgudel (approved):
Looks good!

PR #4615, opened 2023-01-20
Update clusterrole and clusterrolebinding

zencircle:


zencircle:
Resolves the error
```
Error: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "pod-reader" namespace: "default" from "": no matches for kind "ClusterRole" in version "rbac.authorization.k8s.io/v1beta1"
ensure CRDs are installed first, resource mapping not found for name: "pod-reader-binding" namespace: "default" from "": no matches for kind "ClusterRoleBinding" in version "rbac.authorization.k8s.io/v1beta1"
ensure CRDs are installed first]
```

PR #4618, opened 2023-01-24
Redis Granular

kgudel:


lchen-2101 (approved):
LGTM, it is understood this change will render previous version of cache invalid with key structure being different.

PR #4620, opened 2023-01-24
[Keycloak] Redirect Institutions API on ProdBeta to Prod

### LINKED ISSUE #4619: Keycloak Should Always looks at default Institutions API for Account Creation
When a user enters their email address the keycloak theme code sends a request to the institutions-api to get the list of institutions that are associated with that email domain.

That request is made [here](https://github.com/cfpb/hmda-platform/blob/master/kubernetes/keycloak/theme-provider/themes/hmda/login/resources/js/register.js#L108) in the theme code.

This code is using the web domain that keycloak is currently on + `/v2/public/institutions`. This means that when you sign in for an account on `ffiec.cfpb.gov` it sends the request to `ffiec.cfpb.gov/v2/public/institutions` and if they are signing up on `ffiec.beta.cfpb.gov` it sends the request to `ffiec.beta.cfpb.gov/v2/public/institutions`.

However, the code should always send the request to the default (non-beta) site.

meissadia:
Closes #4619 

## Changes
When running on ProdBeta the Keycloak registration page will rely on the Institutions API data from ProdDefault.

All other environments will continue to use the Institutions API of the host their running on.

### DIFF
--- a/kubernetes/keycloak/theme-provider/themes/hmda/login/resources/js/register.js
+++ b/kubernetes/keycloak/theme-provider/themes/hmda/login/resources/js/register.js
@@ -103,10 +103,19 @@
 
   }
 
+  // ProdBeta will use Prod data, 
+  // other systems will use data from their host.
+  function deriveInstitutionsApiHost() {
+    if (window.location.host == 'ffiec.beta.cfpb.gov')
+      return 'https://ffiec.cfpb.gov'
+    
+    return ''
+  }
+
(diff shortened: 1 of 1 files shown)


meissadia:
Moving this back to `Draft` state.  
 
Abhi brought up the scenario of Beta-only users.  This approach, of always using the PROD Institutions API, does not account for those users so we will need to rethink how to resolve this.

meissadia:
This will require a dev with more in depth knowledge of the intricacies of auth management.

PR #4657, opened 2023-03-29
adding 2023 support to hmda reporting

PatrickGoRaft:


common/src/main/resources/reference.conf
```diff
@@ -188,7 +188,7 @@ hmda {
         filename = ${?2022_CENSUS_FILENAME}
       }
       2023 {
-        filename = "ffiec_census_2023.txt"
+        filename = "ffiec_census_2022.txt"
```
lchen-2101, line 191: is this right? just double checking

PatrickGoRaft: @lchen-2101  2023 census isn't out until the fall so we're forced to use this for quarterly

lchen-2101: Gotcha, fun times :joy:. LGTM, possible future improvement, maybe have a configmap that houses temporary env vars for overriding, and still have the default be what it's supposed to be, so u don't have to rebuild later.

lchen-2101 (approved):
LGTM

PR #4730, opened 2023-12-15
Update README file

jaredb96:


README.md
```diff
@@ -101,12 +101,31 @@ The image below shows the cloud vendor agnostic technical architecture for the H
...
+
+```bash
+sdk install sbt
+```
+Clone the repo and go into the repo directory:
+```bash
+git clone https://github.com/cfpb/hmda-platform.git
+cd hmda-platform
+```
+Before running with sbt, make sure your sbt version matches with the project version. Exit the project directory and run sbt -version to check the version. Go back to the project directory, go to project directorym then edit the build.properties to match your actual sbt version. NOTE: Please do NOT commit any changes to this file to the remote repo, revert it before pushing
```
rkovalik-raft, line 122: I don't think updating this is necessary, as long as your script version when you run `sbt -version` is higher than the version listed in `project/build.properties`

jaredb96: That makes sense. Should we note that or remove this paragraph altogether?

rkovalik-raft: I think it would be good to note that we should check what version of sbt is in `project/build.properties` and make sure to install sbt at that version or higher.

rkovalik-raft (approved):

PR #4820, opened 2024-06-04
Update logic in MLAR income categorization method to match the resulting percentage range

rkovalik-raft:
In the helper method used to create income categorization data for static MLAR datasets, some of the income comparison logic doesn't match with the returned percentage range string. This PR fixes this and adds unit tests for ModifiedLarRepository.


PatrickGoRaft (approved):
LGMT

PR #4824, opened 2024-06-12
updates to de-duplicate email domain lists

### LINKED ISSUE #4817: Institutions Email Domain Consolidation
There's a logic flaw occurring when an  institution is added / updated to our data store that causes multiple versions of the email domain to be stored and referenced due to each filing season being treated as its own entity but the domains as a single source. 

While we investigate the underlying cause of the issue we can deduplicate repeated domains in the interim.

PatrickGoRaft:
closes #4817

### DIFF
--- a/hmda-data-publisher/src/main/scala/hmda/publisher/scheduler/PanelScheduler.scala
+++ b/hmda-data-publisher/src/main/scala/hmda/publisher/scheduler/PanelScheduler.scala
@@ -114,39 +114,11 @@ class PanelScheduler(publishingReporter: ActorRef[PublishingReporter.Command], s
     }
   }
 
-  def appendEmailDomains2018(institution: InstitutionEntity): Future[InstitutionAltEntity] = {
-
-    val emails: Future[Seq[InstitutionEmailEntity]] =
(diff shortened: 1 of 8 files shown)


hmda-data-publisher/src/main/scala/hmda/publisher/scheduler/PanelScheduler.scala
```diff
@@ -114,11 +115,10 @@ class PanelScheduler(publishingReporter: ActorRef[PublishingReporter.Command], s
     }
   }
 
-  def appendEmailDomains2018(institution: InstitutionEntity): Future[InstitutionAltEntity] = {
-
+  def appendEmailDomains(institution: InstitutionEntity): Future[InstitutionAltEntity] = {
     val emails: Future[Seq[InstitutionEmailEntity]] =
       emailRepository.findByLei(institution.lei)
-
+    var seqTest:Seq[String] = Seq("1","2,3","4","5","6")
```
jaredb96, line 121: Shouldn't this be "2", "3"?

jaredb96 (approved):

PR #4826, opened 2024-06-12
add validation to LEI in data browser

### LINKED ISSUE #4825: Validate LEI Format in Data-Browser
LEI is currently being passed into the API as a raw string, we should confirm its in the right format before passing it on

kgudel:
Closes #4825


PatrickGoRaft (changes requested):
We could move this logic to our HMDAMessageFilter object 
 
 https://github.com/cfpb/hmda-platform/blob/master/common/src/main/scala/hmda/messages/HmdaMessageFilter.scala
 
 and we can probably simplify this with the LEI regex already defined in this chunk of code. 
 
 https://github.com/cfpb/hmda-platform/blob/master/common/src/main/scala/hmda/messages/HmdaMessageFilter.scala#L19-L22

PatrickGoRaft:
LGTM

PR #4829, opened 2024-06-20
Data Publisher Tests

### LINKED ISSUE #4813: Add tests to Data Publisher

jaredb96:
This PR is to add some tests to the data publisher microservice. 

closes #4813


hmda-data-publisher/build.sbt
jaredb96, line 1: This package isn't used in our code so we exclude it from coverage

hmda-data-publisher/build.sbt
```diff
@@ -1,3 +1,4 @@
 packageName in Docker := "hmda-data-publisher"
 
 version in Docker := "latest"
+coverageExcludedPackages := "hmda.publisher.qa*"
```
jaredb96, line 4: Excluding this package, it isn't used in the code

hmda-data-publisher/src/test/scala/hmda/publisher/query/component/PublisherComponent2021Spec.scala
jaredb96, line 1: These Component tests pretty much work the same way across years, save for some minor changes to the structure of the data

hmda-data-publisher/src/test/resources/loanapplicationregister2022.sql
jaredb96, line 1: These are just new mock tables for our tests

hmda-data-publisher/src/test/scala/hmda/publisher/util/ScheduleCoordinatorSpec.scala
jaredb96, line 1: This class should be tested more but for now I want to start here

PatrickGoRaft:
LGTM.

hmda-data-publisher/src/test/scala/hmda/publisher/query/component/PublisherComponent2021Spec.scala
```diff
@@ -0,0 +1,131 @@
...
+        )
+      ),
+      30.seconds
+    )
+  }
+
+  "InstitutionRepository2021 runthrough" in {
+    import institutionRepo._
+    val data = InstitutionEntity("EXAMPLE-LEI", activityYear = 2020, institutionType = 1, taxId = "ABC", hmdaFiler = true)
+    data.toPSV
```
tptignor, line 50: What is the function of the toPSV and all the toFooPSV invocations? I understand these methods return strings and leave data as is. Is something catching and testing the returned strings? It looks like the tests are based on the original "data" vals.

hmda-data-publisher/src/test/scala/hmda/publisher/query/component/PublisherComponent2021Spec.scala
```diff
@@ -0,0 +1,131 @@
+package hmda.publisher.query.component
+
```
tptignor, line 2: One main general comment: while I shudder to think about adding more abstraction to anything in scala, has anyone thought about generalizing these tests (PublisherComponent, LarEntityImpl, possibly others?) As noted, the tests "pretty much work the same way." If it could be done, we wouldn't have to create and maintain separate tests for every year in perpetuity.

hmda-data-publisher/src/test/scala/hmda/publisher/query/component/PublisherComponent2021Spec.scala
jaredb96: I thought that PSV was a data format for mlar and it's invocation here in the tests is just to make sure it runs but I could be wrong, this was based on the previous tests that were in the codebase.

hmda-data-publisher/src/test/scala/hmda/publisher/query/component/PublisherComponent2021Spec.scala
jaredb96: I agree, there's a ton of refactoring that we can do on the current tests and beyond. A lot of data in the codebase in structured in a way that we have to "create and maintain separate X for every year" so that seems like a clean-up project during down-season.

PR #4837, opened 2024-06-26
Fix snake CVE

rkovalik-raft:
Updates Kubernetes api client-java version to [21.0.0-legacy](https://mvnrepository.com/artifact/io.kubernetes/client-java/21.0.0-legacy), which fixes CVE issue related to snake.yaml. I chose the `21.0.0-legacy` version because we would need to make significant code changes if we used the `21.0.0` version, since that version makes a method we rely on private instead of public.

Compiles and passes unit tests for hmda-platform and check-digit services, working on testing other services.


rkovalik-raft:
Note that the higher s3mock version requires Java 17, and tests/builds still fail due to cassandra issues

rkovalik-raft:
Note: to run this with sbt, need to use this command:
`env JDK_JAVA_OPTIONS="--add-opens java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.net=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED --add-opens=java.base/java.util.concurrent=ALL-UNNAMED --add-opens=java.base/java.util.concurrent.atomic=ALL-UNNAMED" sbt`

There are still some errors in `hmda-platform` when running locally, but it passes tests and the docker image builds fine.

Also `ratespread-calculator` still has cassandra related errors that cause a test to fail.

project/Version.scala
```diff
@@ -37,6 +40,6 @@ object Version {
   val scalacheckShapeless    = "1.2.5"
   val diffx                  = "0.4.0"
   val log4j                  = "2.15.0"
-  val kubernetesApi          = "15.0.1"
+  val kubernetesApi          = "21.0.0-legacy"
```
rkovalik-raft, line 43: Upgrading this past version 17 will fix the snake.yaml cve issue. I'm using `21.0.0-legacy` version here because we would need to make significant code changes if we used the 21.0.0 version, since that version makes a method we rely on private instead of public.

project/scoverage.sbt
```diff
@@ -1 +1,2 @@
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.10")
\ No newline at end of file
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.10")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
rkovalik-raft, line 2: Allows us to generate dependency graphs for our services

ratespread-calculator/src/test/scala/hmda/calculator/scheduler/APORSchedulerSpec.scala
```diff
@@ -1,122 +1,122 @@
...
-  }
-
-  override def afterAll(): Unit = {
-    Try(s3mock.stop())
-    super.afterAll()
-  }
-
-}
\ No newline at end of file
+//package hmda.calculator.scheduler
```
rkovalik-raft, line 1: Another test that is commented out because it relies on S3Mock

PatrickGoRaft (approved):
These changes LGTM.

Pending verification in Dev environment smoke testing

rkovalik-raft:
Confirmed with @Michaeldremy that the Cypress DEV tests looked good

PR #4878, opened 2024-08-30
Separate tests into dedicated jobs

jaredb96:


tptignor:
This all looks like what we planned with the one outstanding consideration for the failing data browser test. @jaredb96 , please post or send new info on that when you can.

jaredb96:
> This all looks like what we planned with the one outstanding consideration for the failing data browser test. @jaredb96 , please post or send new info on that when you can.

I'm going to make a ticket to address fixing this failing test. So for now I think this ticket is good to go if you don't see any other problems.

tptignor (approved):
LGTM with an upcoming data_browser_tests ticket

PR #4879, opened 2024-08-30
Update quartz scheduler version to resolve cve

jaredb96:


tptignor:
Hi @jaredb96 ,

Hoping you can provide a little context. Is the goal simply to resolve a CVE? Have not looked into Akka event scheduling until now, but it seems like the added resource is intended to replace the current "enragedginger" scheduler (https://github.com/enragedginger/akka-quartz-scheduler). Is that right? If so, shouldn't we be replacing akkaQuartzScheduler where it appears?

jaredb96:
> Hi @jaredb96 ,
> 
> Hoping you can provide a little context. Is the goal simply to resolve a CVE? Have not looked into Akka event scheduling until now, but it seems like the added resource is intended to replace the current "enragedginger" scheduler (https://github.com/enragedginger/akka-quartz-scheduler). Is that right? If so, shouldn't we be replacing akkaQuartzScheduler where it appears?

Returning this PR to drafts but yes, it is to resolve the CVE caused by the quartz scheduler. See here: https://scout.docker.com/vulnerabilities/id/CVE-2023-39017?s=gitlab&n=quartz&ns=org.quartz-scheduler&t=maven&vr=%3C%3D2.3.2

PR #4880, opened 2024-08-30
Update zookeeper version to resolve cve

jaredb96:


jaredb96:
This CVE cannot currently be resolved since it requires updating Cormorant, and the most updated version does not update zookeeper: https://mvnrepository.com/artifact/io.chrisdavenport/cormorant-core_2.13/0.5.0-M1

PR #4920, opened 2024-10-28
Analytics duplication refactor

jaredb96:
I implemented repository wrapper classes to decrease the code duplication in the hmda analytics app


hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -66,118 +66,105 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
+    def createSchema(): Unit = transmittalSheet.createSchema()
+    def dropSchema(): Unit = transmittalSheet.dropSchema()
+
+    def insert(ts: TransmittalSheetEntity): Future[Int] = transmittalSheet.insert(ts)
+
+    def findByLei(lei: String): Future[Seq[TransmittalSheetEntity]] = transmittalSheet.findByLei(lei)
+
+    def deleteByLei(lei: String): Future[Int] = transmittalSheet.deleteByLei(lei)
+
+    def deleteByLeiAndQuarter(lei: String): Future[Int] = transmittalSheet.deleteByLeiAndQuarter(lei)
```
tptignor, line 85: We don't need quarterly functions in yearly wrappers, do we? If not, let's remove.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -66,118 +66,105 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
+
+    def updateByLei(ts: TransmittalSheetEntity): Future[Int] = transmittalSheet.updateByLei(ts)
+
+    def count(): Future[Int] = transmittalSheet.count()
+
+    private def getTableNameByYear(year: String): String = {
```
tptignor, line 120: Looks like the getTableNameByYear functions in the quarterly wrappers take only the year as a function arg but get the quarter through the constructor, which is the class body. This is confusing. Recommend to standardize the getTableNameByYear functions or just remove them.

jaredb96: I removed the paramters to just make them simple private helpers

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
jaredb96: Removed

tptignor: Reloaded a couple times and still seeing deleteByLeiAndQuarter right above this comment... Does it look different on your end?

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -66,118 +66,105 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
+    def createSchema(): Unit = transmittalSheet.createSchema()
+    def dropSchema(): Unit = transmittalSheet.dropSchema()
+
+    def insert(ts: TransmittalSheetEntity): Future[Int] = transmittalSheet.insert(ts)
+
+    def findByLei(lei: String): Future[Seq[TransmittalSheetEntity]] = transmittalSheet.findByLei(lei)
+
+    def deleteByLei(lei: String): Future[Int] = transmittalSheet.deleteByLei(lei)
+
+    def deleteByLeiAndQuarter(lei: String): Future[Int] = transmittalSheet.deleteByLeiAndQuarter(lei)
```
tptignor, line 114: So how do transmittalSheet.deleteByLeiAndQuarter (and larRepo.deletebyLeiAndQuarter below) actually work if we don't call them with a quarter?

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -227,27 +214,27 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
+              case Period(2022, Some("Q1")) => QuarterlyTransmittalSheetRepositoryWrapper("2022", "Q1").deleteByLei(ts.lei)
+              case Period(2022, Some("Q2")) => QuarterlyTransmittalSheetRepositoryWrapper("2022", "Q2").deleteByLei(ts.lei)
+              case Period(2022, Some("Q3")) => QuarterlyTransmittalSheetRepositoryWrapper("2022", "Q3").deleteByLei(ts.lei)
+              case Period(2023, Some("Q1")) => QuarterlyTransmittalSheetRepositoryWrapper("2023", "Q1").deleteByLei(ts.lei)
+              case Period(2023, Some("Q2")) => QuarterlyTransmittalSheetRepositoryWrapper("2023", "Q2").deleteByLei(ts.lei)
+              case Period(2023, Some("Q3")) => QuarterlyTransmittalSheetRepositoryWrapper("2023", "Q3").deleteByLei(ts.lei)
+              case Period(2024, Some("Q1")) => QuarterlyTransmittalSheetRepositoryWrapper("2024", "Q1").deleteByLei(ts.lei)
+              case Period(2024, Some("Q2")) => QuarterlyTransmittalSheetRepositoryWrapper("2024", "Q2").deleteByLei(ts.lei)
+              case Period(2024, Some("Q3")) => QuarterlyTransmittalSheetRepositoryWrapper("2024", "Q3").deleteByLei(ts.lei)
```
tptignor, line 237: Don't we want to replace the deleteByLeiAndQuarter invocations with invocations of new deleteByLeiAndQuarter functions?

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -392,85 +379,85 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
         .mapAsync(1) { lar =>
           for {
             insertorupdate <- submissionId.period match {
```
rkovalik-raft, line 381: Could we replace this very long match statement with a map? And maybe do the same for the other match statements here? It would cut down on a ton of lines of code

tptignor:
@PatrickGoRaft , what are your thoughts on this PR generally? I'm sure it can be made to work if there are any current functional problems, and I'm always for reusing code, but how many lines are we cutting out of the total code size? I wonder if we could do even better with.a 2D (year, quarter) array, if it's worth it.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
jaredb96: Both were probably needed at some point, the second one just further filters by quarterly records. This function is no longer needed especially after the refactor, will delete.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
jaredb96: We don't have to, and given the above comment, we shouldn't! Will fix

jaredb96:
> @PatrickGoRaft , what are your thoughts on this PR generally? I'm sure it can be made to work if there are any current functional problems, and I'm always for reusing code, but how many lines are we cutting out of the total code size? I wonder if we could do even better with.a 2D (year, quarter) array, if it's worth it.

@tptignor I don't want to mix up work to much. Let's keep this PR with its current class structure that I wrote. A further refactor, with a complete redesign, ought to be another ticket.

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
jaredb96: Hmm could you clarify a bit what you were thinking?

hmda-analytics/src/main/scala/hmda/analytics/query/TransmittalSheetComponent.scala
```diff
@@ -87,9 +87,6 @@ trait TransmittalSheetComponent {
     def deleteByLei(lei: String): Future[Int] =
       db.run(table.filter(_.lei.toUpperCase === lei.toUpperCase).delete)
 
-    def deleteByLeiAndQuarter(lei: String): Future[Int] =
```
jaredb96, line 90: This function is redundant right now so I deleted it - not used anywhere that deleteByLei isn't

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -347,27 +328,27 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
-              case Period(2022, Some("Q1")) => larRepository2022Q1.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2022, Some("Q2")) => larRepository2022Q2.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2022, Some("Q3")) => larRepository2022Q3.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2023, Some("Q1")) => larRepository2023Q1.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2023, Some("Q2")) => larRepository2023Q2.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2023, Some("Q3")) => larRepository2023Q3.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2024, Some("Q1")) => larRepository2024Q1.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2024, Some("Q2")) => larRepository2024Q2.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
-              case Period(2024, Some("Q3")) => larRepository2024Q3.deletebyLeiAndQuarter(lar.larIdentifier.LEI)
+              case Period(2018, None) => YearlyLarRepositoryWrapper("2018").getLarRepository.deleteByLei(lar.larIdentifier.LEI)
```
tptignor, line 331: Does this work? I just noticed above that YearlyLarRepositoryWrapper doesn't have a deleteByLei method. Seems to be unique in that way.

jaredb96: You're correct it doesn't have one - it grabs a repo object and calls that ones deleteByLei. It does work since its doing that. I chose not to reimplement the lar repo functions in the wrapper only because the lar repo is such a small class. No other reason, good catch!

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
rkovalik-raft: Never mind, map isn't the right thing here. Not sure what I was thinking yesterday. 

What I meant is we can simplify this to 2-3 cases: `case Period(year, None) =>` and `case Period(year, Some(quarter)) =>` (and keep the final `case _ =>` case). Using a variable for the year and quarter will get rid of a ton of these match cases.

jaredb96: Oh yeah let's do that, will take some time to update

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -317,85 +317,12 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
           for {
             insertorupdate <- submissionId.period match {
               case Period(2018, None) => YearlyLarRepositoryWrapper("2018").getLarRepository.insert(LarConverter2018(lar))
```
jaredb96, line 319: 2018 has a special lar converter so we're keeping this here

hmda-analytics/src/main/scala/hmda/analytics/HmdaAnalyticsApp.scala
```diff
@@ -392,86 +316,13 @@ object HmdaAnalyticsApp extends App with TransmittalSheetComponent with LarCompo
...
-                )
-              case Period(2024, Some("Q2")) =>
-                larRepository2024Q2.insert(
-                  LarConverter(lar = lar, 2024, isQuarterly = true)
-                )
-              case Period(2024, Some("Q3")) =>
-                larRepository2024Q3.insert(
-                  LarConverter(lar = lar, 2024, isQuarterly = true)
+              case Period(2018, None) => YearlyLarRepositoryWrapper("2018").getLarRepository.insert(LarConverter2018(lar))
+              case Period(submissionId.period.year, None) => YearlyLarRepositoryWrapper(submissionId.period.year.toString).getLarRepository.insert(
```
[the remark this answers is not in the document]
jaredb96: Yeah this works, on another unit test ticket I'm going to refactor this a bit to demonstrate that. I think we should stick to making this depend on the submissionId for now since this line is a part of reading from that.

PR #4939, opened 2024-12-04
resolve missing region in AWS Email Client

### LINKED ISSUE #4938: Resolve Missing Region Bug With Email Service
The HMDA Email service has encountered a bug in which the default AWS region is no longer being set when creating the `AmazonSimpleEmailServiceClientBuilder` object.

PatrickGoRaft:
closes #4938


tptignor (approved):
Noting the region hasn't changed in 6 years. LGTM

PR #4941, opened 2024-12-05
Remove V602 from quarterly engine

kgudel:


PatrickGoRaft (changes requested):
can you update these lines to be 

`  override def validityChecks(ctx: ValidationContext) = TsEngine2024.validityChecks(ctx).appendedAll(Vector(V718.withContext(ctx)))filter(_ != V602)}
`

PatrickGoRaft:
Changes copied into https://github.com/cfpb/hmda-platform/pull/4944

PR #4949, opened 2024-12-09
resolving cert resolution issue

PatrickGoRaft:


kgudel (commented):
Let's combine the two docker files into a single init that everything can use

PR #4953, opened 2024-12-17
Revert to keycloak 25

kgudel:


kubernetes/keycloak/keycloak_values.yaml
```diff
@@ -1,7 +1,7 @@
 image:
   registry: docker.io
   repository: bitnami/keycloak
-  tag: 26.0.5
+  tag: 25.0.6
```
PatrickGoRaft, line 4: Can we make this 25.0.3 @KennethGudel

PatrickGoRaft (changes requested):

PatrickGoRaft (changes requested):
Lets switch this to  v25.0.3. Versions beyond this (even minor) result in unstable deployments

PatrickGoRaft (approved):

PR #4971, opened 2025-01-28
Remove CI writing to PRs to accommodate forks

DVentresco:
- Writing to PRs from forks in CI isn't allowed. Allowing it causes a security risk, so removing it altogether 
    - https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
- Broken submodule removed since it throws a git exit 128 in CI cleanup


PatrickGoRaft (approved):

PatrickGoRaft:
LGTM.

PR #4974, opened 2025-01-30
Cleanup for Snyk Testing

DVentresco:
- Remove outdated code from repo
- Change .gitignore to account for local sbt builds


kgudel (approved):
-7k, that's pretty impressive. Looks good to me.

PR #5005, opened 2025-04-10
chore: update base image and dependencies

### LINKED ISSUE #5007: HMDA Platform Keycloak Adapter Deprecation Research
From @lchen-2101 : 

> Upon further investigation, the latest adapter is no longer compatible with the latest core library. Keycloak adapter has been in deprecation state since version 22; seems they finally stopped support in version 25; so we need to overhaul our auth code to remediate the keycloak related CVEs (possibly removing keycloak libs altogether, and switch to the standard OIDC approach).

lchen-2101:
closes #5007


project/Version.scala
```diff
@@ -18,7 +18,11 @@ object Version {
   val postgres               = "42.6.0"
   val h2                     = "2.1.214"
   val apacheCommons          = "2.6"
-  val keycloak               = "25.0.3"
+  val keycloakAdapter        = "25.0.3"
+  val keycloakCore           = "25.0.6"
+//  val keycloakCore           = "26.1.4"
```
lchen-2101, line 23: seems the deployed instance needs to be updated accordingly to have the latest dep work.

PatrickGoRaft (approved):
LGTM

PR #5027, opened 2025-06-08
Resolve formatting issue with hmda analytics and mlar

PatrickGoRaft:


common/src/main/scala/hmda/parser/filing/lar/LarParser.scala
```diff
@@ -10,16 +10,21 @@ import scala.util.{Failure, Success, Try}
...
-      case Failure(e) =>
-        parserValidationError.invalidNel
+      case Failure(e) => parserValidationError.invalidNel
     }
 
-  def validateDoubleField(value: String, parserValidationError: ParserValidationError): LarParserValidationResult[Double] =
+  def validateDoubleFieldReturnString(value: String, parserValidationError: ParserValidationError): LarParserValidationResult[String] =
     Try(value.toDouble) match {
-      case Success(i) => i.validNel
+      case Success(i) => value.validNel
```
lchen-2101, line 27: if user actually put in scientific notation, do we want to prevent it? I'm wondering if we might want to go the regex route?

PR #5041, opened 2025-07-21
Akka resolver test

lchen-2101:


lchen-2101:
Found out secrets are not passed to the runner when PR is triggered from a forked repo: https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/using-secrets-in-github-actions#using-secrets-in-a-workflow

So had to create branch and PR from within the base repo; there might be a way to deal with this using `GITHUB_TOKEN` permissions, but we don't have access to see / modify the permissions.

PR #5048, opened 2025-08-11
Correcting formatting on large values

PatrickGoRaft:


common/src/main/scala/hmda/parser/filing/lar/LarParser.scala
```diff
@@ -23,11 +23,10 @@ trait LarParser {
     }
 
   def validateDoubleFieldReturnString(value: String, parserValidationError: ParserValidationError): LarParserValidationResult[String] =
-    Try(value.toDouble) match {
-      case Success(_) => if ("[-\\d.]+".r.matches(value)) value.validNel else parserValidationError.invalidNel
+    Try(BigDecimal(value)) match {
+      case Success(i) => i.bigDecimal.toPlainString.validNel
```
lchen-2101, line 27: if we still want to prevent sci-note input from users, we need to have the regex portion

PR #5052, opened 2025-08-20
chore: sqlite cve patch

lchen-2101:
deleting the helm charts here as well as we don't use the custom charts here, and having it stay causes helm installs to create resources not managed by us, and not being used.


PatrickGoRaft (approved):
LGTM

PR #5065, opened 2025-09-24
Change Q659 logic to use double instead of string regex

kgudel:


PatrickGoRaft:
@lchen-2101  I'll let you be the tie breaker on which solution we should use.

PR #5111, opened 2026-03-25
Delete mlar by submission id rather than lei 

kgudel:
This is to account for institutions filing with an lei other than expected


common/src/main/scala/hmda/query/repository/ModifiedLarRepository.scala
```diff
@@ -51,6 +51,21 @@ class ModifiedLarRepository(databaseConfig: DatabaseConfig[JdbcProfile]) {
...
+    * This is to account for the change from S303 to Q303
+    * @param submissionId
+    * @return the number of rows removed
+    */
+  def deleteBySubmissionID(submissionId: SubmissionId): Future[Int] = {
+    val period = submissionId.period.year
+    val lei = submissionId.lei
+    val likeStatement = s"${lei}-${period}-%"
+    db.run(
+      sqlu"""DELETE FROM #${fetchYearTable(period.toInt)} WHERE submission_id LIKE $likeStatement"""
```
lchen-2101: we should be able to just delete by the lei only? since the mlar tables themselves are already period specific?

PatrickGoRaft: @KennethGudel

PatrickGoRaft (approved):
lgtm

PatrickGoRaft (approved):

PR #5113, opened 2026-03-31
updates in support of mlar release

PatrickGoRaft:


file-proxy/src/main/resources/application.conf
```diff
@@ -15,11 +15,11 @@ akka.http.server.request-timeout = 4888888 seconds
 hmda {
   publication {
     years {
-      dynamic = "2017,2018,2019,2020,2021,2022,2023,2024"
+      dynamic = "2017,2018,2019,2020,2021,2022,2023,2024,2025"
```
kgudel: Let's remove the no longer relevant years.

PatrickGoRaft: fair enough,  I'll remove 2021 and prior

kgudel (approved):

PR #5116, opened 2026-04-29
fix(github-actions): pin versions to sha

billhimmelsbach:
Due to some recent supply chain attacks, Lamin would prefer if we locked the github actions to SHAs instead of just the version number. I used the same script that cf.gov [used over here](https://github.cfpb.gov/Design-Development/cfgov/issues/4673#issuecomment-375885), thanks @chosak!

## Changes

- Instead of using a version number, locks github actions to SHAs

## Testing

1. Do the github actions still work? Yes!
2. Do the tests still pass? Yes!

### DIFF
--- a/.github/workflows/cve-scan-pr.yml
+++ b/.github/workflows/cve-scan-pr.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
       - name: Check out repo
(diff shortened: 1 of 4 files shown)


PatrickGoRaft (approved):
LGTM.

PR #5117, opened 2026-04-29
update base alpine jdk image

zencircle:


lchen-2101 (approved):

lchen-2101:
welp, the previous tag actually meant the latest version of alpine, so the next build's would've been alpine 3.23 anyway

zencircle:
> welp, the previous tag actually meant the latest version of alpine, so the next build's would've been alpine 3.23 anyway
Its actually alpine `3.23.4` but jdk images do not have that tag

lchen-2101:
yeap, forgot to paste the image here, the digest are the same between the 2 tags:
image
and going into the details, both are 3.23.4 on the base image

PR #5122, opened 2026-05-05
fix check-digit and data-browser

zencircle:


lchen-2101:
why is this needed? and what determines whether the module should be prefixed with `hmda` or not? feels like this would cause a lot of unforeseen downstream issues. the databrowser test failure in github action is one apparent issue from this.

PatrickGoRaft:
> why is this needed? and what determines whether the module should be prefixed with `hmda` or not? feels like this would cause a lot of unforeseen downstream issues. the databrowser test failure in github action is one apparent issue from this.

just consolidating our naming across the services

zencircle:
I could keep project name `data-browser`and update kubernets helm chart currently`hmda-data-browser-api`

zencircle:
Closing this PR will fix helm chart to databrowser

PR #5161, opened 2026-09-17
Send an institution modified event when the filing flag is set

kgudel:


lchen-2101:
I think this can probably be simplified, since the `ModifyInstitution` event is already fired in the `setHmdaFilerFlag` method, which persists into the actor system, as well as publishing a kafka event for institutions api to pick up; we'd be publishing multiple events from here. I think if we revert what we did in `SignHttpApi`, and just move the name updation into the `setHmdaFilerFlag` method, it should accomplish the previously required update to set the institution name

lchen-2101:
alternate PR merged to resolve the same ticket: https://github.com/cfpb/hmda-platform/pull/5162

